Issue #1151, opened 2023-08-19
Fail to get service endpoint if helm resource does not contains namespace


gitlawr:
**Steps to reproduce**
1. Create template from `https://github.com/walrus-tutorials/helm-chart?ref=v0.0.1`
2. Use the template to create a service, use chart from `https://github.com/walrus-tutorials/assets/releases/download/microservices-demo/sock-shop-0.3.0.tgz`

**Result**
service endpoint does not show up.

**Environment**
- Walrus version: v0.3.0-rc2

**Additional context**
Logs:
```
2023-08-19T08:26:00.160Z	D	task.resource-status-sync	processed in 137.699215ms
2023-08-19T08:26:00.160Z	E	task	error executing task: resource-status-sync: error getting kubernetes service endpoints /carts-db: an empty namespace may not be set when a resource name is provided
github.com/seal-io/walrus/utils/cron.timeoutTask.Process
	github.com/seal-io/walrus/utils@v0.0.0-00010101000000-000000000000/cron/cron.go:174
reflect.Value.call
	reflect/value.go:584
reflect.Value.Call
	reflect/value.go:368
github.com/go-co-op/gocron.callJobFuncWithParams
	github.com/go-co-op/gocron@v1.29.0/gocron.go:116
github.com/go-co-op/gocron.runJob
	github.com/go-co-op/gocron@v1.29.0/executor.go:76
github.com/go-co-op/gocron.(*executor).runJob
	github.com/go-co-op/gocron@v1.29.0/executor.go:173
github.com/go-co-op/gocron.(*executor).run.func1
	github.com/go-co-op/gocron@v1.29.0/executor.go:239
```

Finenyaco:
**Environment**
- Walrus version: main b51994e3fdba0aaeaba9439f6c609c105e5cbb92
- UI version: dev-3b160ec

**Test Result:** Failed

return an empty record

<img width="1647" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/9523c5aa-347f-4a5d-8620-e66931b554c6">

Finenyaco:
**Environment**
- Walrus version: main 5cf1d9ef1dcb6bd27a686a24454d760fdc19ec22
- UI version: dev-5cf1d9e

**Test Result:** pass
<img width="2475" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/386c6a5b-fa11-4ddb-9858-b7261a3cd5a1">

PR #1164, opened 2023-08-21
fix: helm manifest may not include namespace

aiwantaozi:

**Problem:**
Manifest from the helm chart may be without a namespace.

**Solution:**
Get the namespace from the release.

**Related Issue:**
#1151


thxCode (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1166, opened 2023-08-21
Import template with ref may have template versions more than one


alexcodelf:
**Describe the bug**
Manually added templates may contain multiple versions

**Steps to reproduce**
1.  import template with `https://github.com/walrus-catalog-demo/demo?ref=main`
2. edit template source `https://github.com/walrus-catalog-demo/demo?ref=dev`
3. check template versions.

**Result**
Template with ref have two versions, but it declares specific version.
<img width="1680" alt="image" src="https://github.com/seal-io/walrus/assets/85595852/46b2ead7-603c-48fa-b009-3e959c02a30a">

**Expected behavior**
Only one template version same as ref `dev` is expected.

**Environment**
- Walrus version: cecb2dd40defcea9c17cd084581406fd258f6f99 dev-HEAD

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main b51994e3fdba0aaeaba9439f6c609c105e5cbb92
- UI version: dev-3b160ec

**Test Result:** pass

- Check UI: cannot modify the template source now.

<img width="696" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/d954556e-ed59-479e-a2aa-d9f4236311e5">

- Check api. Return 202 after changing the source but it did not work.

<img width="1434" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/3ce38069-681a-47a5-9c2d-5ba8c1c439f0">

PR #1167, opened 2023-08-21
chore: make template and catalog source immutable.

alexcodelf:

**Problem:**
As template and template version used by service can not deleted, when the source of template changed, it may cause template version changed after template sync is done, we do not want user change this field which may cause unexpected result of service.

**Solution:**
Make source field immutable, UI need adjustment. @hibig 

**Related Issue:**
#1166


thxCode (approved):
LGTM

aiwantaozi (approved):

PR #1175, opened 2023-08-21
Fix setting update API must add name field

aiwantaozi:
**IMPORTANT: Please do not create a Pull Request without creating an issue first.**

**Problem:**
setting API could not update by below API request, must add field: name to the request body

```sh
curl --location --request PUT 'https://localhost:443/v1/settings/:id' \
--header 'Content-Type: application/json' \
--header 'Cookie: 
--data '{
  "value": "0 */6 * ? * *"
}'
```
and
```sh
curl --location --request PUT 'https://localhost:443/v1/settings/:name' \
--header 'Content-Type: application/json' \
--header 'Cookie: 
--data '{
  "value": "0 */6 * ? * *"
}'
```

**Solution:**
Update setting update input to allow update setting value with id and name

**Related Issue:**
#1172

Issue #1172, opened 2023-08-21
[v0.3.0][CLI]Optimizing input when using the CLI to update settings


linyinli:
**Is your enhancement related to a problem? Please describe.**
When using the CLI to update settings, now we should input `<setting>` and `--name <setting>`, for example:
```
walrus setting update OpenAiApiToken --name OpenAiApiToken --value ${OPENAI_TOKEN}
```
It doesn't make sense, I think we should remove one.

**Describe the solution you'd like**
Keep `<setting>` or `--name <setting>` only.

Finenyaco:
**Environment**
- Walrus version: main 2314b9d4571c3b6b341dee7ed630a5ffc5793dfe
- UI version: dev-0835269

**Test Result:** pass

`--name <setting>`  has been removed.

```
 ~/.walrus/ wa setting update -h
Usage:
  walrus setting update <setting> [flags]

Flags:
      --value string
```

```
 ~/.walrus/ wa  setting update OpenAiApiToken --value "xxx"

```

PR #1176, opened 2023-08-22
fix: lose index fields after update input modeling

thxCode:

**Problem:**

Failed to gain the index fields from the *UpdateInput's `Model` function.

**Solution:**

Adjust entx generation template to value the index fields during *UpdateInput's `Model` function.

**Related Issue:**
PR https://github.com/seal-io/walrus/pull/1175 addresses #1172, but it exposes the problem that this PR described. I prefer to make this PR a further fix for #1172 .


aiwantaozi (approved):

gitlawr (approved):
LGTM

Issue #1174, opened 2023-08-21
Failed to generate certificate by acme after changing the DNS


Finenyaco:
**Describe the bug**

Failed to generate certificate by acme after changing the DNS

**Steps to reproduce**
1. Install walrus on an existing Kubernetes cluster by using acme to generate trusted certificate.
2. Change the value of `SERVER_TLS_AUTO_CERT_DOMAINS` and reapply.
3. Check the logs of walrus pod.

**Result**

1. Cannot use the new DNS to generate certificate.  It has to manually delete the old token's secret.
```
2023-08-21T09:42:21.843Z	W	api.https	http: TLS handshake error from 192.168.0.173:60010: acme/autocert: no token cert for "xxx"
```

**Expected behavior**

No need to manually delete the secret.

**Environment**
- Walrus version: v0.3.0

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 3486239019a0f63bb024a12309027f88b8f4bad9

**Test Result:** pass

Issue #1173, opened 2023-08-21
Failed to run walrus if probe is used when generating certificates with acme


Finenyaco:
**Describe the bug**

When using ACME to generate certificate, using probes will make pods fail to run.

**Steps to reproduce**
1. Install walrus by [using ACME to generate trusted certificate](https://seal-io.github.io/docs/deploy/replication#use-acme-to-generate-trusted-certificate). 
2. Add probes to pods.
https://github.com/seal-io/walrus/blob/main/deploy/manifests/kubernetes.yaml#L624-L643
3. Check the installation results.

**Result**

1. Failed to run the pod.

```
 ~/seal/manifests/ kubectl describe pod walrus-6cd767b69f-6sn46
Name:             walrus-6cd767b69f-6sn46
Namespace:        walrus-system
Priority:         0
Service Account:  walrus
Node:             ip-192-168-0-173.ap-east-1.compute.internal/192.168.0.173
Start Time:       Mon, 21 Aug 2023 14:22:49 +0800
Labels:           app.kubernetes.io/component=walrus
                  app.kubernetes.io/name=walrus-server
                  app.kubernetes.io/part-of=walrus
                  app.kubernetes.io/test=w1
                  pod-template-hash=6cd767b69f
Annotations:      <none>
Status:           Running
IP:               192.168.4.103
IPs:
  IP:           192.168.4.103
Controlled By:  ReplicaSet/walrus-6cd767b69f
Containers:
  walrus-server:
    Container ID:   containerd://2005cceb228b0a6b1c7a12d96b135d2e3a402d104cad4d834db5ce54c35cb0cf
    Image:          sealio/walrus:v0.3.0-rc3
    Image ID:       docker.io/sealio/walrus@sha256:04fad1cc95a60ede46e9b345f53635969d27c81867540225713921fcba5f0e56
    Ports:          80/TCP, 443/TCP
    Host Ports:     0/TCP, 0/TCP
    State:          Waiting
      Reason:       CrashLoopBackOff
    Last State:     Terminated
      Reason:       Error
      Exit Code:    2
      Started:      Mon, 21 Aug 2023 14:29:14 +0800
      Finished:     Mon, 21 Aug 2023 14:30:05 +0800
    Ready:          False
    Restart Count:  6
    Limits:
      cpu:     4
      memory:  8Gi
    Requests:
      cpu:      500m
      memory:   512Mi
[...]

Issue #1168, opened 2023-08-21
Remove unnecessary logs


Finenyaco:
**Environment**
- Walrus version: main

Too many TLS handshake error logs were prints. Please remove it if it is not necessary.

```
2023-08-21T08:10:00.195Z	W	api.https	http: TLS handshake error from 192.168.54.152:4578: EOF

2023-08-21T08:10:02.501Z	W	api.https	http: TLS handshake error from 192.168.0.173:15798: EOF

2023-08-21T08:10:09.168Z	W	api.https	http: TLS handshake error from 192.168.0.173:41700: EOF

2023-08-21T08:10:10.195Z	W	api.https	http: TLS handshake error from 192.168.54.152:26190: EOF

2023-08-21T08:10:13.528Z	W	api.https	http: TLS handshake error from 192.168.54.152:3414: EOF

2023-08-21T08:10:16.861Z	W	api.https	http: TLS handshake error from 192.168.54.152:52123: EOF

2023-08-21T08:10:19.168Z	W	api.https	http: TLS handshake error from 192.168.0.173:4808: EOF

2023-08-21T08:10:29.168Z	W	api.https	http: TLS handshake error from 192.168.0.173:9090: EOF

2023-08-21T08:10:32.501Z	W	api.https	http: TLS handshake error from 192.168.0.173:45415: EOF

2023-08-21T08:10:33.528Z	W	api.https	http: TLS handshake error from 192.168.54.152:22088: EOF

2023-08-21T08:10:35.835Z	W	api.https	http: TLS handshake error from 192.168.0.173:40000: EOF

2023-08-21T08:10:46.861Z	W	api.https	http: TLS handshake error from 192.168.54.152:56002: EOF

2023-08-21T08:10:55.835Z	W	api.https	http: TLS handshake error from 192.168.0.173:47680: EOF

2023-08-21T08:10:56.861Z	W	api.https	http: TLS handshake error from 192.168.54.152:55844: EOF

2023-08-21T08:10:59.168Z	W	api.https	http: TLS handshake error from 192.168.0.173:39493: EOF

2023-08-21T08:11:02.501Z	W	api.https	http: TLS handshake error from 192.168.0.173:36213: EOF

2023-08-21T08:11:03.528Z	W	api.https	http: TLS handshake error from 192.168.54.152:22523: EOF

2023-08-21T08:11:05.835Z	W	api.https	http: TLS handshake error from 192.168.0.173:42534: EOF

[...]

PR #1179, opened 2023-08-22
fix: cert generated

thxCode:
**Problem:**

1. Log too many useless EOF when deployed behind AWS ELB.
2. Cannot finish the ACME after changing domain name.
3. Cannot finish the ACME with the default Kubernetes deployment.

**Solution:**

1. Discard the HTTP error log with the EOF suffix.
2. Introduce a Group concept to KMS service, different groups can own the key with same name.
3. Donot redirect the Kubelet probe request from 80 to 443.

**Related Issue:**

#1168 #1174  #1173

### DIFF
--- a/deploy/manifests/kubernetes.yaml
+++ b/deploy/manifests/kubernetes.yaml
@@ -639,6 +639,10 @@ spec:
             timeoutSeconds: 5
             periodSeconds: 10
             httpGet:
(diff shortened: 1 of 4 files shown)


alexcodelf (approved):
LGTM

aiwantaozi (approved):

Issue #1180, opened 2023-08-22
refactor: backgroud jobs


thxCode:
**Is your enhancement related to a problem? Please describe.**

1. The registration of background jobs is odd. https://github.com/seal-io/walrus/blob/34837e37921bbcf7fdd76c92765ff5aa9eee9543/pkg/server/init_background_jobs.go#L21-L43
2. We should remove the progressing mutex.
3. We already have a Cost status syncing loop per 1 hour, but the Connector status syncer also syncs FinOps status per 5 mins.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main b0c3bd4666d7e891f3713760d3223998d4388a0f
- UI version: dev-c12f198

**Test Result:** pass, it won't synchronize cost data when executing ConnectorStatusSync task now.

PR #1181, opened 2023-08-22
Fix jobs

thxCode:
**Problem:**

1. The registration of background jobs is odd.
2. We should remove the progressing mutex.
3. We already have a Cost status syncing loop per 1 hour, but the Connector status syncer also syncs FinOps status per 5 mins.

**Solution:**

1. adjust the registration of background jobs with builder pattern. 
2. remove the progressing mutex and print a warning log for the long-term execution job within the distribution locker locking.
3. do not sync the cost status during connector status checking.

**Related Issue:**
#1180

### DIFF
--- a/pkg/server/init_background_jobs.go
+++ b/pkg/server/init_background_jobs.go
@@ -3,140 +3,169 @@ package server
 import (
 	"context"
 
-	"k8s.io/client-go/rest"
-
 	"github.com/seal-io/walrus/pkg/cron"
-	"github.com/seal-io/walrus/pkg/dao/model"
 	catalogskd "github.com/seal-io/walrus/pkg/scheduler/catalog"
 	connskd "github.com/seal-io/walrus/pkg/scheduler/connector"
-	serviceskd "github.com/seal-io/walrus/pkg/scheduler/service"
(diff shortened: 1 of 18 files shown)


pkg/server/init_background_jobs.go
```diff
@@ -19,124 +16,175 @@ import (
...
 
-func buildConnectorCostCollectJobCreator(mc model.ClientSet) cron.JobCreator {
-	return func(ctx context.Context, name, expr string) (cron.Expr, cron.Task, error) {
-		task, err := connskd.NewCollectTask(mc)
+// jobCreatorBuilder is the stereotype of a function that creates a cron.JobCreator,
+// the given context.Context is a long-term processing context until the server is stopped.
+type jobCreatorBuilder func(context.Context, initOptions) (expr settings.Value, job cron.JobCreator)
+
+func buildCatalogTemplateSyncJobCreator(
+	_ context.Context,
```
aiwantaozi, line 46: Why shadow this context?

thxCode: lint

aiwantaozi (approved):

alexcodelf (approved):
LGTM

aiwantaozi (approved):

PR #1182, opened 2023-08-23
fix: missing project in env connectors

gitlawr:
**Problem:**
When Get/List environments, the connector project is missing. as a result, it fails when we PUT using the entity from GET response.

**Solution:**
Add missing fields.


thxCode (approved):
LGTM

Issue #1192, opened 2023-08-24
Icon for standalone template does not show


gitlawr:
**Describe the bug**
When a template is imported from a catalog, the icon is shown. 
When it's added as a standalone template, the icon is missing.

**Environment**
- Walrus version: v0.3.0

Finenyaco:
**Environment**
- Walrus version: main b51994e3fdba0aaeaba9439f6c609c105e5cbb92
- UI version: dev-3b160ec

**Test Result:** pass

<img width="624" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/30d23dd0-982a-48cd-9e81-100f112f384b">

PR #1194, opened 2023-08-24
fix: icon for standalone template does not show

alexcodelf:
**Problem:**
Icon for standalone template does not show, manually sync will not update icon. The reason is that using the nickname instead of repository name to fetch icon.

**Solution:**
When sync standalone template or catalog template, support fetch icon and description.

**Related Issue:**

#1192


thxCode (approved):
LGTM

aiwantaozi (approved):

Issue #1189, opened 2023-08-24
refactor: speed up the response of the service resources listing


thxCode:
**Is your enhancement related to a problem? Please describe.**

The response of the service resources listing is slow with `withKeys` query parameter. 

Assuming the cluster is around, a service that owns 10 resources can cost almost 8s.

**Describe the solution you'd like**

Speed up the response.

**Describe alternatives you've considered**

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main b0c3bd4666d7e891f3713760d3223998d4388a0f
- UI version: dev-c12f198

**Test Result:** pass

<img width="2560" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/e85cda92-a64a-439a-8242-848b7f2aebe3">

If the cluster is disconnected, it takes about 5s to get the resources.
<img width="2554" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/8911867c-fa24-4c95-9593-8f69ab8ad867">

PR #1197, opened 2023-08-24
refactor: speed up getting resource keys

thxCode:
**Problem:**

The response of the service resources listing is slow with `withKeys` query parameter. If the connectors are disconnected, the situation will intensify.

**Solution:**

1. For an operator, initialize the required clients first, and change the content type of the k8s rest config.
2. Get the keys of resources in parallel.

**Related Issue:**
#1189

### DIFF
--- a/pkg/operator/k8s/operator.go
+++ b/pkg/operator/k8s/operator.go
@@ -14,6 +14,7 @@ import (
 	dynamicclient "k8s.io/client-go/dynamic"
 	batchclient "k8s.io/client-go/kubernetes/typed/batch/v1"
 	coreclient "k8s.io/client-go/kubernetes/typed/core/v1"
+	networkingclient "k8s.io/client-go/kubernetes/typed/networking/v1"
 	"k8s.io/client-go/rest"
 
 	"github.com/seal-io/walrus/pkg/dao/types"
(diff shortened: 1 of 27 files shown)


pkg/operator/k8s/operator.go
```diff
@@ -90,10 +90,29 @@ func (Operator) Type() optypes.Type {
 
 // IsConnected implements operator.Operator.
 func (op Operator) IsConnected(ctx context.Context) error {
-	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	ctx, cancel := context.WithTimeout(ctx, time.Second)
```
gitlawr, line 93: keep the origin

pkg/serviceresources/graph.go
```diff
@@ -87,7 +82,7 @@ func GetVerticesAndEdges(
 			dfs(entity.Edges.Instances[j])
 		}
 
-		// Hide instance resources's dependencies.
+		// Hide instance resource's dependencies.
```
gitlawr, line 85: s/instance/service/

pkg/serviceresources/key.go
```diff
@@ -0,0 +1,183 @@
...
+// SetKeys sets the keys of the resources for operations like log and exec.
+//
+// The given model.ServiceResource item must specify the following fields:
+// Shape, Mode, ID, DeployerType, Type, Name and ConnectorID.
+func SetKeys(
+	ctx context.Context,
+	logger log.Logger,
+	modelClient model.ClientSet,
+	candidates []*model.ServiceResource,
+	operators map[object.ID]optypes.Operator,
```
gitlawr, line 28: Do we need this param? Seems like it is only used internally

thxCode: the graph of the environment.  https://github.com/seal-io/walrus/pull/1197/files#diff-1051d160d8e9e62077a18876598246711d6d5d7c236ac5e0d2d99f4ae94a91afR88-R96

aiwantaozi (approved):

pkg/serviceresources/graph.go
thxCode: updated

gitlawr (approved):
LGTM

Issue #998, opened 2023-07-26
Upgrade deploy-source-code service fails after modification: Provider produced inconsistent result after apply


Finenyaco:
**Version:** main	
**UI:** dev-91cbed0
**Commit ID:** e8abf74ef7061965340ac1985a2d6d8a08747bdd
**Env:** AWS EC2 Ubuntu 22.04 amd64

**Test Steps:** 

1. Create a service using `deploy-source-code` template.
2. Edit the service, modify the tag for image and save.
3. Check the upgrade result.

**Test Result:** 

1. Fail to upgrade service.
<img width="770" alt="image" src="https://github.com/seal-io/seal/assets/26084756/a2cca491-1c9e-4e13-a716-146ead0324b8">

```
Error: Provider produced inconsistent result after apply

When applying changes to module.my-flask.kaniko_image.image, provider
"provider[\"registry.terraform.io/seal-io/kaniko\"]" produced an unexpected
new value: .destination: was cty.StringVal("moerer/my-flask:07261636"), but
now cty.StringVal("moerer/my-flask:07261618").

This is a bug in the provider, which should be reported in the provider's own
issue tracker.
```

Finenyaco:
**Environment**
- Walrus version: main 5cf1d9ef1dcb6bd27a686a24454d760fdc19ec22
- UI version: dev-5cf1d9e

**Test Result:**  Failed

Template： deploy-source-code  Version：v0.0.3

Another error is returned when upgrading.

```
If you also control the child module, add a required_providers entry named
"kubernetes" with the source address "hashicorp/kubernetes".

Error: kaniko build failed

  with module.test-source-upgrade.kaniko_image.image,
  on .terraform/modules/test-source-upgrade/main.tf line 14, in resource "kaniko_image" "image":
  14: resource "kaniko_image" "image" {

Secret "" is invalid: metadata.name: Required value: name or generateName is
required
```

Finenyaco:
**Environment**
- Walrus version: main 7dddda0f4b8d278b4731528af87dbf93f8a17d97
**Test Result:** pass

**Test Steps:**
1. Add a template `https://github.com/walrus-catalog/deploy-source-code?ref=main`.
2. Use the template to create a service.
3. Change the tag and upgrade.

PR #1200, opened 2023-08-25
feat: optimize the repetitive execution of seal-io/kaniko tasks

cdryzun:
**Problem:**

When using the deploy-source-code model to run CI/CD, if the action being executed is an Upgrade, value conflicts may occur.

**Solution:**

1. The reason is that the seal-io/kaniko in provider cannot effectively manage one-time tasks when integrating with Terraform, which may not necessarily be a problem with the provider.
2. Before executing apply, delete the state of seal-io/kaniko in the state to make it run.

**Related Issue:**
#998

---
image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,11 +6,12 @@ permissions:
   actions: read
 
 env:
-  REPO: "sealio"
+  REPO: "${{ vars.REPO || 'sealio' }}"
   VERSION: "${{ github.ref_name }}"
   BUILD_PLATFORMS: "linux/amd64,linux/arm64"
   PARALLELIZE: "false"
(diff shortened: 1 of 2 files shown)


alexcodelf:
This issue looks like a provider issue, can you fix this in [terraform-provider-kaniko](https://github.com/seal-io/terraform-provider-kaniko)?

The '.destination' value is defined in here: https://github.com/seal-io/terraform-provider-kaniko/blob/main/kaniko/resource_image.go#L32

cdryzun:
I think I can give it a try, but I'm not sure if I will succeed.

gitlawr:
Resolved in https://github.com/walrus-catalog/deploy-source-code/releases/tag/v0.0.4

Issue #1202, opened 2023-08-25
support listening ipv6 address


thxCode:
**Is your enhancement related to a problem? Please describe.**

Currently, we only support listening ipv4 address in dual stack, failed to start with a ipv6 address.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

thxCode:
1. verify `--bind-address="ipv6"`, print a log: `serving https on "ipv6" by "tcp"`.
2. verify `--bind-address="ipv4"`, print a log: `serving https on "ipv4" by "tcp"`.
3. verify `--bind-address="ipv6"` and `--bind-with-dual-stack=false`, print a log: `serving https on "ipv6" by "tcp6"`.
4. verify `--bind-address="ipv4"` and `--bind-with-dual-stack=false`, print a log: `serving https on "ipv4" by "tcp4"`.

Finenyaco:
**Environment**
- Walrus version: main 7dddda0f4b8d278b4731528af87dbf93f8a17d97

**Test Result:** pass


1. `--bind-address="ipv6"`:
```
2023-09-07T02:33:12.103Z	I	api	serving https on "[2001:db8:2::123]:443" by "tcp"
2023-09-07T02:33:12.103Z	I	api	serving http on "[2001:db8:2::123]:80" by "tcp"
```

```
root@9aec7df7a6a1:/# curl -k https://[2001:db8:2::123]:443
[...]

PR #1204, opened 2023-08-25
fix: support binding ipv6 address

thxCode:
**Problem:**

Unable to start the server with --bind-address="ipv6 address", as the string in the form of "ipv6:port" is invalid for net listening.

**Solution:**

Parse the given bind address, if the given address is a strict IPv6, we use "[ipv6]:port" format.

Allow user to pick dual-stack mode or single-stack mode now.

**Related Issue:**
#1202

### DIFF
--- a/pkg/apis/server.go
+++ b/pkg/apis/server.go
@@ -38,6 +38,7 @@ type ServeOptions struct {
 	SetupOptions
 
 	BindAddress        string
+	BindWithDualStack  bool
 	TlsMode            TlsMode
 	TlsCertFile        string
 	TlsPrivateKeyFile  string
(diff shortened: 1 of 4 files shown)


rainfd (commented):
LGTM

rainfd (dismissed):
LGTM

pkg/apis/server.go
```diff
@@ -222,15 +235,39 @@ func serve(ctx context.Context, handler http.Handler, errorLog *stdlog.Logger, l
 	return nil
 }
 
-func newTcpListener(ctx context.Context, ip string, port int) (net.Listener, error) {
-	address := fmt.Sprintf("%s:%d", ip, port)
+func parseBindAddress(ip string, dual bool, port int) (network, address string, err error) {
```
gitlawr, line 238: nit: The order `ip, port, dual` looks more comfortable to me.

Can we add some unit tests?

thxCode: updated

gitlawr (approved):
LGTM

Issue #1201, opened 2023-08-25
leak goroutine when cron locker switches


thxCode:
**Describe the bug**

The renewal approach of the cron locker will leak goroutine 15 minutes when the cron locker switches.

https://github.com/seal-io/walrus/blob/2d3fa5bb8d7206219bc311b26c071bef6ccfb629/pkg/cron/locker.go#L173-L177.

**Steps to reproduce**

**Result**
Leaking goroutine keeps in 15 minutes.

**Expected behavior**
Revert the leaking goroutine.

**Environment**
- Walrus version: main

**Additional context**
The renewal approach cannot guarantee only one running task at a moment.

thxCode:
closed as https://github.com/seal-io/walrus/pull/1205 merged.

PR #1205, opened 2023-08-25
refactor: cron locker

thxCode:
**Problem:**

The original renew loop may not quit after the holder changes.

**Solution:**

- If the holder is not me, quit the loop.
- Adjust the log to print the same style text of cron job.

**Related Issue:**
#1201

### DIFF
--- a/pkg/cron/locker.go
+++ b/pkg/cron/locker.go
@@ -2,41 +2,42 @@ package cron
 
 import (
 	"context"
+	"database/sql"
+	"errors"
 	"fmt"
-	"strings"
 	"time"
 
 	"entgo.io/ent/dialect/sql/sqlgraph"
+	"k8s.io/apimachinery/pkg/util/wait"
 
 	"github.com/seal-io/walrus/pkg/dao/model"
 	"github.com/seal-io/walrus/pkg/dao/model/distributelock"
 	"github.com/seal-io/walrus/utils/cron"
 	"github.com/seal-io/walrus/utils/gopool"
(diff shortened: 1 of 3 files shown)


pkg/cron/locker.go
```diff
@@ -119,136 +123,117 @@ func (l *Locker) createLock(ctx context.Context, key string) error {
...
-		case <-ctx.Done():
-			return
-		}
+	if err != nil {
+		return nil, err
 	}
+
+	return &Lock{
+		logger: l.logger.WithName("locker").
+			WithValues("key", key, "id", strs.Hex(6)),
```
aiwantaozi, line 155: What is the id used for? 
Can you unify the logger for Locker and Lock?

pkg/cron/locker.go
```diff
@@ -119,136 +123,117 @@ func (l *Locker) createLock(ctx context.Context, key string) error {
...
 	})
+	if err != nil && ctx.Err() == nil {
+		if model.IsNotFound(err) {
+			// Not found, should not renew again.
+			return true, nil
+		}
+
+		lk.logger.Warnf("error renewing, try at next loop: %v", err)
+	}
+
```
aiwantaozi, line 213: while the job is finished, and the key is unlocked, seems like the renew goroutine is still here, while the current instance still successfully gets lock next time, the renew goroutine can't exit

aiwantaozi (approved):

gitlawr (approved):
LGTM

Issue #980, opened 2023-07-20
A large number of services (5000+) can cause the server not responding


Finenyaco:
**Version:** main	
**UI:** dev-afae2b2
**Commit ID:** 68e121d64cab4d59254b9f8b69d3c5e4aafe4321
**Env:** AWS EC2 Ubuntu 22.04 amd64

**prerequisites:**
1.  Set postgresql max connections to 403.
```
postgres:
    image: "postgres:14.8"
    # set the max connections to 403
    command: postgres -c max_connections=403
   ...
```
2. The number CPUs of seal server is 4. Modify seal's  `data-source-conn-max-open` and `data-source-conn-max-idle` as follows:
```
seal:
  command:
      - "--data-source-conn-max-open=400"
      - "--data-source-conn-max-idle=400"
     ...
```

**Test Steps:**

1. Add a template called `variables` with source `github.com/Finenyaco/modules_test//variables`.
2. Create 5000+ service by `variables` template v0.0.1 without resources.
<img width="1251" alt="image" src="https://github.com/seal-io/seal/assets/26084756/a5873cae-760a-4e94-a1e8-25bbc4849cc3">
3. Check the `goroutine pool` on grafana dashboard.

**Test Result:** 

1. 100% goroutine utilization, the server is not responding.
<img width="751" alt="image" src="https://github.com/seal-io/seal/assets/26084756/0f08dda6-81b9-40d2-8fc2-ddcd8ec4fb4a">
<img width="560" alt="image" src="https://github.com/seal-io/seal/assets/26084756/99e04b40-bc28-48b5-a6ba-3ae9eaee6020">
<img width="1194" alt="image" src="https://github.com/seal-io/seal/assets/26084756/6230415c-c04f-4047-b0b5-82250c63406d">

2. After restarting the seal server,  it still does not recover.

3. Modify the `gopool-worker-factor` of seal to 200. The seal server is back to normal.

<img width="766" alt="image" src="https://github.com/seal-io/seal/assets/26084756/7c0c0592-4c6c-4c1e-a6b9-90c43414acd1">
<img width="565" alt="image" src="https://github.com/seal-io/seal/assets/26084756/947f9252-4c69-4d6b-8d4d-a92ebfe81c7b">

PR #1208, opened 2023-08-27
fix: no response with a large number of services

thxCode:
**Problem:**

The execution unit of the current background task does not take the following factors into consideration:
1. Most clients have a request rate limitation(i.e. Kubernetes client-go), much concurrency will result in an excessively long execution cycle.
2. Much concurrency will cause the request rate limitation of server(i.e. Kubernetes api-server), even inducing instability of the server, finally affecting the subsequent scheduling cycles.
3. A long-running task usually consists of multiple subtasks. When the task triggers timeout management and is canceled, subsequent subtasks should be canceled directly.

**Solution:**

1. Submit parallel tasks according to the concurrency that the client can provide.
2. Connectors with the same trait usually have the same burst. For example, both connector A and connector B are connected to the same cluster. Connectors with the same trait need to be managed uniformly through the burst.

**Related Issue:**
#980


pkg/scheduler/serviceresource/helper.go
```diff
@@ -22,4 +60,82 @@ func listCandidateConnectors(ctx context.Context, modelClient model.ClientSet) (
...
+	}
+
+	return indexer, limiter, nil
+}
+
+// getBucket returns the bucket size and count with the given total and burst,
+// the multiple of the returning results must be greater than or equal to the given total,
+// especially, the returning count must be less than or equal to the given burst.
+func getBucket(total, burst, minSize int) (size, count int) {
+	if minSize <= 0 {
```
rainfd, line 114: The count parameter is confusing. Because the burst parameter generally means the maximum number of tokens allowed in a bucket, but here brust is applied to the count parameter.

pkg/scheduler/serviceresource/helper_test.go
```diff
@@ -0,0 +1,81 @@
...
+				burst:   64,
+				minSize: 100,
+			},
+			expected: output{
+				size:  100,
+				count: 50,
+			},
+		},
+		{
+			name: "total/burst is less than minSize, and total is not a multiple of burst",
```
alexcodelf, line 60: give cases of negative and zero

pkg/scheduler/serviceresource/helper.go
thxCode: updated, modify to `getBatches` and enrich the comment message.

pkg/scheduler/serviceresource/helper_test.go
thxCode: updated

pkg/operator/types/types.go
```diff
@@ -24,6 +24,12 @@ type Operator interface {
 	// Burst returns the maximum number of operations that can be called at once.
 	Burst() int
 
+	// Trait returns the trait of this operator,
+	// which is used for grouping or other else.
+	//
+	// Trait returns a blank string if no trait provides.
+	Trait() string
```
gitlawr, line 31: ID() should be more clear

pkg/apis/servicerevision/extension.go
```diff
@@ -206,88 +246,97 @@ func (h Handler) manageResources(ctx context.Context, entity *model.ServiceRevis
...
 		}
 	})
 
 	return nil
 }
 
-func (h Handler) syncServiceStatusAndResourceLabel(
+// completeResources completes the resources,
+// including states resources/labels resources/composes resources.
+func completeResources(
```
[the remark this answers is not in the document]
thxCode: updated

pkg/operator/types/types.go
thxCode: updated

pkg/apis/servicerevision/extension.go
```diff
@@ -206,88 +246,97 @@ func (h Handler) manageResources(ctx context.Context, entity *model.ServiceRevis
...
-		op, err := operator.Get(ctx, optypes.CreateOptions{
-			Connector: *c,
-		})
+		// Discover resources.
+		ncrs, err := serviceresources.Discover(ctx, op, modelClient, crs)
 		if err != nil {
-			logger.Errorf("error getting operator of connector %s: %v",
-				c.ID, err)
-			continue
+			logger.Errorf("error discovering component resources: %v", err)
```
aiwantaozi, line 316: why remove the continue

gitlawr (approved):
LGTM

pkg/apis/servicerevision/extension.go
thxCode: do not block the following operations.

aiwantaozi (approved):

Issue #1190, opened 2023-08-24
Update the cron expression in system settings not working as expected


hibig:
**Describe the bug**
1. In the `system settings`  tab `Task Settings` edit the  'Check service resource status' or 'Check service dependency status'
2. Edit any one, but the other  is Updated.

**Steps to reproduce**
1. Go to 'system settings' page. click tab `Task Settings`, click Basic Task edit icon.
2. Edit  'Check service resource status' 
3. Click save icon

**Result**
1. `Check service resource status` not updated, but `Check service dependency status` updated.

**Expected behavior**
`Check service resource status` updated

**Environment**
- Walrus version: main or v0.3.0

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 5cf1d9ef1dcb6bd27a686a24454d760fdc19ec22
- UI version: dev-2c89dfb

**Test Result:** pass

PR #1218, opened 2023-08-28
fix: serval issues of settings, especially job expressions

thxCode:
**Problem:**

1. Since the SQL SELECT results do not match the WHERE clause predicates, a collection update validation will refill the info to the wrong position.
https://github.com/seal-io/walrus/blob/b51994e3fdba0aaeaba9439f6c609c105e5cbb92/pkg/dao/model/setting_view.go#L582-L585
2. After introducing data listening https://github.com/seal-io/walrus/pull/1125, the `Settings` update/collection update no need to notify data changes anymore.
3. Data listening does not perform cache inactivation management on the `Settings` value of the non-job expression.
4. The name of cron.Syncer seems odd, actually, it acts as a cron expression correcter.

**Solution:**

1. Introduce an indexer to record the requested items' order, and make the refilling work as expected.
2. Remove the bus notification of the handler, and refactor the Sync function of `pkg/cron` package.
3. Remove the cache implementation of `Settings` now. 
    + Although the inactivation management of the settings cache can be performed through data listening, it is too fragmented. We can implement a cache manager on the DAO layer later.
4. Rename the cron.Syncer to cron.Correcter, and private it.

**Related Issue:**
#1190

### DIFF
--- a/pkg/apis/setting/basic.go
+++ b/pkg/apis/setting/basic.go
@@ -31,9 +31,7 @@ func (h Handler) Update(req UpdateRequest) error {
 			return errorx.HttpErrorf(http.StatusNotFound, "setting %s not found", req.Name)
 		}
 
-		_, err := s.Set(req.Context, tx, req.Value)
-
-		return err
+		return s.Set(req.Context, tx, req.Value)
 	})
 }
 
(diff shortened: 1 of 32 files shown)


gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #1212, opened 2023-08-28
Remove the logs about cache


Finenyaco:
**Describe the solution you'd like**

Remove the logs about cache. Too many logs are printed.

```
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       46 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.310Z	D	cache.memory	expired:       47 B | auths/auth...
2023-08-28T09:28:13.872Z	D	task	executed task	{"createBy": "ResourceComponentsDiscoverCronExpr"}
2023-08-28T09:28:21.651Z	D	task	executed task	{"createBy": "ResourceLabelApplyCronExpr"}
[...]

PR #1220, opened 2023-08-29
fix: verbosity of memory cache deletion not found

thxCode:
**Problem:**

Print too many `Entry not found` logs during cache deleting.

**Solution:**

Mute the error if the key is not found under the memory cache.

**Related Issue:**
#1212


rainfd (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1224, opened 2023-08-29
Fail to synchronize catalog templates: Bad credentials


Finenyaco:
**Steps to reproduce**
1. Add a catalog `https://github.com/terraform-alicloud-modules`.
2. Check the catalog status.

**Result**

1. Fail to synchronize templates.
<img width="2468" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/09efb530-c0e9-4e50-a02e-27ea2db7cf6e">

```
2023-08-29T07:25:38.484Z	E	catalog	failed to sync catalog terraform-alicloud-modules templates: Bad credentials
github.com/seal-io/walrus/pkg/catalog.catalogSyncer.Do.func1
	github.com/seal-io/walrus/pkg/catalog/catalog.go:178
github.com/alitto/pond.(*WorkerPool).executeTask
	github.com/alitto/pond@v1.8.3/pond.go:454
github.com/alitto/pond.worker
	github.com/alitto/pond@v1.8.3/worker.go:32
```

**Expected behavior**

**Environment**
- Walrus version: b51994e3fdba0aaeaba9439f6c609c105e5cbb92

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 7c753a44166934745fe27656a6df6744edfa3473
- UI version: dev-d735146

**Test Result:** pass
<img width="2428" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/dabbca32-d742-438c-819a-93d5cc85a042">

PR #1226, opened 2023-08-29
fix: vsc client bad credential

alexcodelf:
**Problem:**
Failed to sync github catalog, the reason is vcs client add empty bearer token when token equals to `""`

**Solution:**
return client wirth user agent when token is empty

**Related Issue:**

#1224


thxCode (approved):
LGTM

gitlawr (approved):
LGTM

PR #1187, opened 2023-08-24
fix: create serivce with nonexistent variables

rainfd:
**IMPORTANT: Please do not create a Pull Request without creating an issue first.**

**Problem:**
Create a service that references a variable that does not exist, request return failed, but the service is created.

**Solution:**
Validate variables before handling request.

**Related  #Issue:**
https://github.com/seal-io/walrus/issues/1003

Issue #1003, opened 2023-07-28
Issue for creating a service that references a variable that does not exist


Finenyaco:
**Version:** main	
**UI:** dev-f2d563b
**Commit ID:** e8abf74ef7061965340ac1985a2d6d8a08747bdd
**Env:** AWS EC2 Ubuntu 22.04 amd64

**Test Steps:**
1. Create a service that references a variable that does not exist. It shows `failed to create service`

<img width="1340" alt="image" src="https://github.com/seal-io/seal/assets/26084756/e1ac1f7a-3cf5-4bb8-aca1-1a7efdadb4f4">

```
2023-07-28T01:29:24.362Z	E	api	error requesting /v1/services?projectID=471301531866497031: missing variables: [host-ip]
github.com/seal-io/seal/pkg/apis/runtime.Erroring.func1
	github.com/seal-io/seal/pkg/apis/runtime/middleware_error.go:40
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.0/context.go:174
github.com/seal-io/seal/pkg/apis/runtime.Recovering.func1
	github.com/seal-io/seal/pkg/apis/runtime/middleware_recovery.go:44
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.0/context.go:174
github.com/seal-io/seal/pkg/apis/runtime.Observing.func1
	github.com/seal-io/seal/pkg/apis/runtime/middleware_observe.go:78
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.0/context.go:174
github.com/gin-gonic/gin.(*Engine).handleHTTPRequest
	github.com/gin-gonic/gin@v1.9.0/gin.go:620
github.com/gin-gonic/gin.(*Engine).ServeHTTP
	github.com/gin-gonic/gin@v1.9.0/gin.go:576
net/http.serverHandler.ServeHTTP
	net/http/server.go:2947
net/http.initALPNRequest.ServeHTTP
	net/http/server.go:3556
net/http.(*http2serverConn).runHandler
	net/http/h2_bundle.go:5910
2023-07-28T01:29:24.362Z	D	api	500 | HTTP/2.0 |      385 B |       88 B |   18.284019ms |   183.14.28.169 | POST    /v1/services?projectID=471301531866497031
```

2. Remove the variable and save again. It shows `dupilicated`

[...]

PR #1228, opened 2023-08-29
fix: creating a service that references a service output that does not exist

rainfd:
**Problem:**
PR https://github.com/seal-io/walrus/pull/1187  only add variable validtion before reqeust handing.
Now creating service with a invalid service output still fail.

**Solution:**
Run all deployer task background. 
User could get response immediately after service creating. The failure with service outputs will be recorded into service revision.

**Related Issue:**
https://github.com/seal-io/walrus/issues/1003

### DIFF
--- a/pkg/service/service.go
+++ b/pkg/service/service.go
@@ -96,26 +96,8 @@ func Apply(
 ) (err error) {
 	logger := log.WithName("service")
 
-	defer func() {
-		if err == nil {
-			return
-		}
-		// Update a failure status.
-		status.ServiceStatusDeployed.False(entity, err.Error())
-
-		uerr := UpdateStatus(ctx, mc, entity)
-		if uerr != nil {
-			logger.Errorf("error updating status of service %s: %v",
-				entity.ID, uerr)
-		}
-	}()
-
 	if !status.ServiceStatusDeployed.IsUnknown(entity) {
 		return errorx.Errorf("service status is not deploying, service: %s", entity.ID)
-	}
-
-	if dp == nil {
-		return errorx.New("deployer is not set")
 	}
 
 	applyOpts := deptypes.ApplyOptions{
@@ -125,7 +107,16 @@ func Apply(
 
 	err = dp.Apply(ctx, entity, applyOpts)
 	if err != nil {
-		return errorx.Errorf("failed to apply service: %v", err)
+		err = fmt.Errorf("failed to apply service: %w", err)
+		logger.Error(err)
+
+		// Update a failure status.
+		status.ServiceStatusDeployed.False(entity, err.Error())
+
+		err = UpdateStatus(ctx, mc, entity)
+		if err != nil {
+			logger.Errorf("error updating status of service %s: %v", entity.ID, err)
+		}
 	}
 
 	return nil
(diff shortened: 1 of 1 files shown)


pkg/service/service.go
```diff
@@ -152,7 +159,8 @@ func Destroy(
 			logger.Errorf("error updating status of service %s: %v",
 				entity.ID, uerr)
 		}
-	}()
+	}
+	defer recycle()
```
aiwantaozi, line 163: ditto

pkg/service/service.go
```diff
@@ -108,7 +109,8 @@ func Apply(
...
```
aiwantaozi, line 113: This defer seems useless, since recycle is already called in gopool.Go.

Could you change the name recycle, since no recycle action here?

pkg/service/service.go
```diff
@@ -118,15 +120,20 @@ func Apply(
...
-	err = dp.Apply(ctx, entity, applyOpts)
-	if err != nil {
-		return errorx.Errorf("failed to apply service: %v", err)
-	}
+		applyOpts := deptypes.ApplyOptions{
+			SkipTLSVerify: !opts.TlsCertified,
+			Tags:          opts.Tags,
+		}
+
+		err = dp.Apply(ctx, entity, applyOpts)
```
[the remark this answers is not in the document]
rainfd: updated

pkg/service/service.go
```diff
@@ -96,7 +98,8 @@ func Apply(
 ) (err error) {
 	logger := log.WithName("service")
 
-	defer func() {
+	// Variable err and ctx are captured by closure.
+	cleanUp := func() {
```
aiwantaozi, line 102: `cleanUp` is usually used for resource release and file removal, etc. You can use `setStatus`, `updateStatus` when performing this action.

pkg/service/service.go
```diff
@@ -118,15 +122,24 @@ func Apply(
...
 	}
 
-	applyOpts := deptypes.ApplyOptions{
-		SkipTLSVerify: !opts.TlsCertified,
-		Tags:          opts.Tags,
-	}
+	// Deployer run in background.
+	gopool.Go(func() {
+		var cancel context.CancelFunc
+		ctx, cancel = context.WithTimeout(context.Background(), 3*time.Minute)
```
aiwantaozi, line 128: ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)

rainfd: updated

alexcodelf (approved):
LGTM

pkg/service/service.go
```diff
@@ -96,36 +96,27 @@ func Apply(
...
-
 	applyOpts := deptypes.ApplyOptions{
 		SkipTLSVerify: !opts.TlsCertified,
 		Tags:          opts.Tags,
 	}
 
 	err = dp.Apply(ctx, entity, applyOpts)
 	if err != nil {
-		return errorx.Errorf("failed to apply service: %v", err)
+		err = fmt.Errorf("failed to apply service: %v", err)
```
aiwantaozi, line 110: use `%w` instead of `%v`, then could unwrap the original error while needed.

rainfd: Updated

aiwantaozi (approved):

Issue #1232, opened 2023-08-29
ServiceRelationshipCheck task will trigger to deploy the service that is depended on other services


Finenyaco:
**Steps to reproduce**
1. Create two services `d` and `b`. Make `b` depend on `d`.
2. Check the deployment result.

**Result**

1. ServiceRelationshipCheck task will trigger to deploy the service, causing the service deployment to fail.
<img width="2497" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/e5fc6e77-ac67-4e87-8059-b27aab3feb68">

**Expected behavior**

It should wait for the dependent services to be deployed before deploying.

**Environment**
- Walrus version: b51994e3fdba0aaeaba9439f6c609c105e5cbb92

**Additional context**

PR #1235, opened 2023-08-30
fix: check dependency failed

alexcodelf:
**Problem:**
ServiceRelationshipCheck task will trigger to deploy the service that is depended on other services

**Solution:**
Missing return false when dependency is not ready， fix check dependency status

**Related Issue:**
#1232


rainfd (approved):
LGTM

PR #1239, opened 2023-08-30
fix: wrong collection validation of entx template

thxCode:
**Problem:**

The collection update validation doesn't work as expected introduced by https://github.com/seal-io/walrus/pull/1218, as there is a wrong indexing during refilling.

**Solution:**

Do indexing according to target elements.

**Related Issue:**
#1190


alexcodelf (approved):
LGTM

Issue #1222, opened 2023-08-29
summary-queried-costs causes panic: invalid memory address or nil pointer dereference


Finenyaco:
**Steps to reproduce**
1. Create a perspective grouped by namespace.
2. View the perspective.

**Result**

1. Returned an error.
<img width="2485" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/a2ff2f9f-4298-4b71-aec3-99bab4f7d81f">
<img width="633" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/91c3d8c1-ac72-467c-838e-92a5d02f0478">

2. Check the logs.
```
2023-08-29T04:40:46.210Z	E	api	panic observing: runtime error: invalid memory address or nil pointer dereference, callstack:
runtime.panicmem
	runtime/panic.go:260 (0x47c3f)
runtime.sigpanic
	runtime/signal_unix.go:839 (0x611f7)
github.com/seal-io/walrus/pkg/apis/cost/validation.ValidateShareCostFilters
	github.com/seal-io/walrus/pkg/apis/cost/validation/query.go:114 (0x3f38f70)
github.com/seal-io/walrus/pkg/apis/cost.(*CollectionRouteGetSummaryQueriedCostsRequest).Validate
	github.com/seal-io/walrus/pkg/apis/cost/extension_view.go:196 (0x3f40687)
github.com/seal-io/walrus/pkg/apis/runtime.(*Router).Routes.func1
	github.com/seal-io/walrus/pkg/apis/runtime/router_route.go:124 (0x3e53d17)
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x3e65623)
github.com/seal-io/walrus/pkg/auths.Account.Filter
	github.com/seal-io/walrus/pkg/auths/account.go:56 (0x3e65600)
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x3e4d503)
github.com/seal-io/walrus/pkg/apis/runtime.If.func1
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_filter.go:54 (0x3e4d4c4)
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x3e4df8f)
github.com/seal-io/walrus/pkg/apis/runtime.RequestThrottling.func2
[...]

Issue #1038, opened 2023-08-11
Synchronizing cost data returns errors: cannot assign requested address


Finenyaco:
**Describe the bug**

**To Reproduce**

1. Synchronize cost data for Kubernetes connector.
<img width="1516" alt="image" src="https://github.com/seal-io/seal/assets/26084756/a1f934cf-02d8-4e7a-8de2-bd6d284cfc6f">

```
[...]

PR #1240, opened 2023-08-30
fix: add max query concurrency to open cost and fix validate bugs

aiwantaozi:
**Problem:**
1. Opencost opens a lot of connections to prometheus, after the request is finished, a bunch of connections are still in TIME_WAIT status, which makes can't open connections to prometheus.
2. Missing empty check for share cost causes validate crash

**Solution:**
1. Adding opencost query concurrency
2. Adding empty share cost query

**Related Issue:**

#1038 
#1222


thxCode (approved):
LGTM

Issue #1196, opened 2023-08-24
Make builtin catalog configurable


gitlawr:
**Is your enhancement related to a problem? Please describe.**
Case 1: users need the builtin catalog from custom URL
Case 2: users bring their own catalog and wants to disable the builtin one.

alexcodelf:
Let's give a setting for user to configure whether to enable the builtin catalog or not, when disable the builtin catalog, the builtin catalog could be removed and will not create again.

Finenyaco:
**Environment**
- Walrus version: main 7c753a44166934745fe27656a6df6744edfa3473
- UI version: dev-d735146

**Test Result:**
- If disable to use the built-in template catalog, the built-in catalog can be removed and will not generate again after restarting walrus.
<img width="825" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/60620ab4-4e56-4fc0-b77d-f64ad85f3253">

- But when `use the built-in template catalog` is enable, the built-in catalog can still be deleted. Please restrict in UI to not allow deletion.

Finenyaco:
**Environment**
- Walrus version: main 5cf1d9ef1dcb6bd27a686a24454d760fdc19ec22
- UI version: dev-2c89dfb

**Test Result:** It doesn't work in latest ui version.


<img width="569" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/2aab27f8-0b81-4350-ba8b-f0be866475c1">
<img width="2501" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/0dc44f83-c66c-4996-bbc1-25c87184b572">

hibig:
Because there is no catalog  type to distinguish between built-in and non-built-in, so the catalog name is "builtin" and the source is "https://catalog.seal.io/walrus-catalog" cannot be deleted, when system setting "use built-in template catalog" enable.

Finenyaco:
**Environment**
- Walrus version: main 5cf1d9ef1dcb6bd27a686a24454d760fdc19ec22
- UI version: dev-5cf1d9e

**Test Result:** pass

<img width="1622" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/4db656a0-3a14-427d-a395-fa64e80402b2">

PR #1241, opened 2023-08-30
feat: add enable builtin catalog setting

alexcodelf:
**Problem:**
Make builtin catalog configurable, even though we can delete builtin catalog. When walrus restart the service, it will auto create a builtin catalog again.

**Solution:**
Add setting to enable or disable builtin catalog.
**Related Issue:**

#1196


thxCode (approved):
LGTM

gitlawr (approved):
LGTM

Issue #744, opened 2023-06-06
Optimize the error message when updating user password with less than 6 characters


Finenyaco:
**Version:** main	
**UI:** dev-6168160
**Commit ID:** 8f7ced5963979c432a3953e24c49434c916b5853
**Env:** Ubuntu 22.04 amd64, aws EC2

**Test Steps:**

1. Create a user.
2. Change user password (less than 6 characters).

**Test Result:** 

1. Return an error `failed to update subject`, which is confusing.
<img width="934" alt="image" src="https://github.com/seal-io/seal/assets/26084756/eed2967a-7551-4334-b83b-66fd8724ca9f">

2. View logs. It said that new password must have at least 6 characters.
```
2023-06-06T08:18:19.715Z	E	api	error requesting /v1/subjects/462709585971511314: failed to update user password to casdoor: failed to set password: New password must have at least 6 characters
github.com/seal-io/seal/pkg/apis/runtime.Erroring.func1
	github.com/seal-io/seal/pkg/apis/runtime/middleware_error.go:40
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.0/context.go:174
github.com/seal-io/seal/pkg/apis/runtime.Recovering.func1
	github.com/seal-io/seal/pkg/apis/runtime/middleware_recovery.go:44
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.0/context.go:174
github.com/seal-io/seal/pkg/apis/runtime.Logging.func2
	github.com/seal-io/seal/pkg/apis/runtime/middleware_route.go:80
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.0/context.go:174
github.com/gin-gonic/gin.(*Engine).handleHTTPRequest
	github.com/gin-gonic/gin@v1.9.0/gin.go:620
github.com/gin-gonic/gin.(*Engine).ServeHTTP
	github.com/gin-gonic/gin@v1.9.0/gin.go:576
net/http.serverHandler.ServeHTTP
	net/http/server.go:2947
net/http.(*conn).serve
	net/http/server.go:1991
2023-06-06T08:18:19.715Z	D	api	500 | HTTP/1.1 |       88 B |   70.088291ms |      172.18.0.1 | PUT     /v1/subjects/462709585971511314
```

gitlawr:
Now it says
[...]

PR #1243, opened 2023-08-31
fix: http error message have status fields

rainfd:
**Problem:**
HTTP error message have status fields.
image
```
{
   "message": "400 Bad Request: error setting password: dial tcp4 127.0.0.1:8000: connect: connection refused",
   "status": 400,
   "statusText": "Bad Request"
}
```

**Solution:**
Remove it when formatting.

**Related Issue:**
https://github.com/seal-io/walrus/issues/744

### DIFF
--- a/staging/utils/errorx/error.go
+++ b/staging/utils/errorx/error.go
@@ -3,9 +3,6 @@ package errorx
 import (
 	"errors"
 	"fmt"
-	"net/http"
-	"reflect"
-	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


thxCode (dismissed):
LGTM

staging/utils/errorx/error.go
```diff
@@ -229,10 +223,8 @@ func (e HttpError) Error() string {
 			}
```
aiwantaozi, line 223: Before will print error with http status while ev.IsNil, after removing the status, this switch case is confused and may return an empty string, which needs to be removed or updated. Could refer to `ErrorX’s Error()`

Issue #1185, opened 2023-08-23
Unable to pull prometheus image over domestic network when installing cost analysis tools


Finenyaco:
**Environment**
- Walrus version:  main 34837e37921bbcf7fdd76c92765ff5aa9eee9543
- UI version: dev-4810199

**Steps to reproduce**
1. Create a domestic Kubernetes cluster.
2. Add the cluster as a connector to walrus and enable FinOps.
3. Check the connector status.

**Result**

1. Failed to install cost analysis tools.
<img width="2418" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/0d39f474-574e-4ad7-94ab-52b53bcda400">

2. View the logs on the cluster.

```
 ~/.kube/ k get pod
NAME                                                    READY   STATUS             RESTARTS   AGE
walrus-opencost-7c5c9f6f9f-pz6gp                        1/1     Running            0          95s
walrus-prometheus-kube-state-metrics-7cbb476bd7-grjnx   0/1     ImagePullBackOff   0          64s
walrus-prometheus-prometheus-node-exporter-8pm7r        0/1     Pending            0          64s
walrus-prometheus-prometheus-node-exporter-zqbd5        0/1     Pending            0          64s
walrus-prometheus-server-cdd94955-fj7l2                 1/2     Running            0          64s

```
```
 ~/.kube/ k describe pod walrus-prometheus-kube-state-metrics-7cbb476bd7-grjnx
...
Events:
  Type     Reason          Age                   From               Message
  ----     ------          ----                  ----               -------
  Normal   Scheduled       14m                   default-scheduler  Successfully assigned walrus-system/walrus-prometheus-kube-state-metrics-7cbb476bd7-grjnx to cn-shenzhen.172.16.24.223
  Normal   AllocIPSucceed  14m                   terway-daemon      Alloc IP 172.16.24.238/20
[...]

PR #1244, opened 2023-08-31
refactor: add default value for cost tools

aiwantaozi:
**Problem:**
Now we need prometheus charts and need to get chart dir, cost image from the environment, which makes it inconvenient in the dev environment

**Solution:**
Support download prometheus charts while it isn't existed, and add default value to make the cost tools

**Related Issue:**
#1185


thxCode (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1096, opened 2023-08-16
Server will restart if too many services are created at the same time


Finenyaco:
**Environment**
- Walrus version: main 1e00759b7dfcf71208e74364c050ef3289b8f2d5
- UI version: dev-a411e9f
 - Testing server: AWS EC2 Ubuntu 22.04 amd64 (4C8G)

**To Reproduce**

1. Deploy the Walrus server via docker-compose.
2. Create 100 services at the same time. 
3. Check the logs. Walrus server has been restarted.

```
2023-08-16T02:26:31.030Z	D	task.service-relationship-check	processed in 1.488287ms
2023-08-16T02:26:31.030Z	D	task	executed task: service-relationship-check
2023-08-16T02:26:32.025Z	D	api	200 | HTTP/2.0 |        0 B |        0 B |    5.674334ms |   18.167.117.84 | GET     /v1/projects/473873687670423554/environments/474064599268720642/services/474162935766253570/revisions/474162935766253570/terraform-states
2023-08-16T02:26:32.237Z	D	api	200 | HTTP/2.0 |        0 B |        0 B |    3.641191ms |   18.167.117.84 | GET     /v1/projects/473873687670423554/environments/474064599268720642/services/474162936353390594/revisions/474162936353390594/terraform-states
2023-08-16T02:26:34.945Z	D	api	200 | HTTP/2.0 |        0 B |        0 B |      2.7683ms |   18.167.117.84 | GET     /v1/projects/473873687670423554/environments/474064599268720642/services/474162937628459010/revisions/474162937645236226/terraform-states
2023-08-16T02:26:34.983Z	D	api	200 | HTTP/2.0 |        0 B |        0 B |    2.337841ms |   18.167.117.84 | GET     /v1/projects/473873687670423554/environments/474064599268720642/services/474162936336613378/revisions/474162936336613378/terraform-states
2023-08-16T02:26:35.275Z	D	api	200 | HTTP/2.0 |        0 B |        0 B |     3.15913ms |   18.167.117.84 | GET     /v1/projects/473873687670423554/environments/474064599268720642/services/474162934860218370/revisions/474162934860218370/terraform-states
2023-08-16T02:26:35.276Z	D	api	200 | HTTP/2.0 |        0 B |        0 B |   18.246565ms |   18.167.117.84 | GET     /v1/projects/473873687670423554/environments/474064599268720642/services/474162935766188034/revisions/474162935766188034/terraform-states
[...]

PR #1245, opened 2023-08-31
fix: shutdown when leader election lost

thxCode:
**Problem:**

Too many service creations will make the Kubernetes cluster inaccessible for a while, and then the inside controller manager will quit as failed to renew its leadership.

**Solution:**

- Support customizing the leader election lease duration and renewing timeout.
- Do not shut down immediately when lost leader election, instead, retry for some time.

**Related Issue:**
#1096

### DIFF
--- a/pkg/k8sctrls/manager.go
+++ b/pkg/k8sctrls/manager.go
@@ -3,76 +3,195 @@ package k8sctrls
 import (
 	"context"
 	"fmt"
+	"strings"
+	"sync/atomic"
+	"time"
 
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes/scheme"
 	"k8s.io/client-go/rest"
+	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	"github.com/seal-io/walrus/pkg/dao/types"
+	"github.com/seal-io/walrus/utils/gopool"
 	"github.com/seal-io/walrus/utils/log"
 )
+
+// ManagerOptions holds the options for creating a new manager.
+type ManagerOptions struct {
+	// IsReady observes whether the manager is ready,
+	// the caller can leverage this symbol to be aware the manager's progress.
+	IsReady *atomic.Bool
+	// LeaderElection indicates whether to enable leader election.
+	LeaderElection bool
+	// LeaderLease indicates the duration of the lease that keeps the leadership.
+	LeaderLease time.Duration
+	// LeaderRenewTimeout indicates the timeout of renewing the leadership.
+	LeaderRenewTimeout time.Duration
+}
 
 func NewManager(opts ManagerOptions) (*Manager, error) {
 	logger := log.WithName("k8sctrl")
 
-	mgr, err := ctrl.NewManager(opts.K8sConfig, ctrl.Options{
-		Scheme:    scheme.Scheme,
-		Logger:    log.AsLogr(logger),
-		Namespace: types.WalrusSystemNamespace,
+	// Defaults.
+	if opts.IsReady == nil {
+		opts.IsReady = &atomic.Bool{}
+	}
+
+	if opts.LeaderLease == 0 {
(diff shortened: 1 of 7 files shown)


pkg/k8sctrls/manager.go
```diff
@@ -3,76 +3,195 @@ package k8sctrls
...
-		LeaderElectionNamespace: types.WalrusSystemNamespace,
-	})
-	if err != nil {
-		return nil, fmt.Errorf("error creating kubernetes controller manager: %w", err)
+		LeaderElection:                opts.LeaderElection,
+		LeaderElectionID:              "walrus-leader-election",
+		LeaderElectionNamespace:       types.WalrusSystemNamespace,
+		LeaderElectionReleaseOnCancel: true,
+		LeaseDuration:                 pointer.Duration(opts.LeaderLease),
+		RenewDeadline:                 pointer.Duration(opts.LeaderRenewTimeout),
```
aiwantaozi, line 64: Why we don't keep the name RenewDeadline?

pkg/k8sctrls/manager.go
```diff
@@ -3,76 +3,195 @@ package k8sctrls
...
-}
+	// Watch for cache sync.
+	gopool.Go(func() {
+		err := wait.PollUntilContextCancel(ctx, time.Second, true,
+			func(ctx context.Context) (done bool, err error) {
+				return mgr.GetCache().WaitForCacheSync(ctx), ctx.Err()
+			})
+		if err != nil {
+			return
+		}
```
aiwantaozi, line 189: Do think we should add a log here for the cache sync error?

pkg/server/runner.go
```diff
@@ -81,22 +85,24 @@ type Server struct {
...
+		EnableTls:              true,
+		TlsCertDir:             apis.TlsCertDirByK8sSecrets,
+		ConnQPS:                10,
+		ConnBurst:              20,
+		KubeConnTimeout:        5 * time.Minute,
+		KubeConnQPS:            16,
+		KubeConnBurst:          64,
+		KubeLeaderElection:     true,
+		KubeLeaderLease:        10 * time.Second,
+		KubeLeaderRenewTimeout: 5 * time.Second,
```
aiwantaozi, line 99: Why we choose 10s lease duration and 5 renew timeout?

pkg/server/runner.go
```diff
@@ -515,12 +560,12 @@ func (r *Server) Run(c context.Context) error {
...
 
 	initOpts := initOptions{
-		K8sConfig:      k8sCfg,
-		K8sCacheReady:  make(chan struct{}),
-		ModelClient:    modelClient,
-		SkipTLSVerify:  len(r.TlsAutoCertDomains) != 0,
-		DatabaseDriver: databaseDrv,
-		CacheDriver:    cacheDrv,
+		K8sConfig:         k8sCfg,
+		K8sCtrlMgrIsReady: &atomic.Bool{},
```
aiwantaozi, line 564: Do you think K8sMgrIsReady is enough? since the startK8sCtrlsOptions use MgrIsReady

pkg/k8sctrls/manager.go
thxCode: Deadline seems more like an epoch not a duration.

pkg/k8sctrls/manager.go
thxCode: the cache waiter has an internal logger.

image

but the --log-verbosity has not affected this, prefer to use another issue to address.

we can trace another log to ensure the cache has be ready.

```
2023-08-31T19:34:25.734+0800	I	k8sctrl	Starting Controller	{"controller": "job", "controllerGroup": "batch", "controllerKind": "Job"}
```

pkg/server/runner.go
thxCode: the `startK8sCtrlsOptions` indicates the `MgrIsReady` is the manager of K8s controllers.

aiwantaozi (approved):

gitlawr (approved):
LGTM

Issue #1247, opened 2023-08-31
refactor: keep the klog/v2 verbosity as the same with utils/log


thxCode:
https://github.com/seal-io/walrus/pull/1245#discussion_r1311498119

thxCode:
closed as https://github.com/seal-io/walrus/pull/1249 merged.

PR #1249, opened 2023-09-01
refactor: make klog verbosity works

thxCode:
**Problem:**

1. Can not recognize the `klog.V()` caller.
2. Always print error logs of logrus in the information level.

**Solution:**

1. Configure the `-v` for the klog logger before starting. 
2. Implement a logrus formatter to process the log according to the level.

**Related Issue:**
#1247

### DIFF
--- a/pkg/server/runner.go
+++ b/pkg/server/runner.go
@@ -5,9 +5,11 @@ import (
 	"database/sql"
 	"encoding/hex"
 	"errors"
+	"flag"
 	"fmt"
 	stdlog "log"
 	"net"
+	"strconv"
 	"strings"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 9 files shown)


pkg/server/runner.go
```diff
@@ -428,13 +429,37 @@ func (r *Server) Flags(cmd *cli.Command) {
...
+	pb := cmd.Before
+	cmd.Before = func(c *cli.Context) error {
+		l := log.GetLogger()
+
+		// Sink the output of standard logger to util logger.
+		stdlog.SetOutput(l)
+
+		// Turn on the logrus logger
+		// and sink the output to util logger.
+		logrus.SetLevel(logrus.TraceLevel)
```
gitlawr, line 441: Why do we set this level?

thxCode: `logrus logger -> [sink to] -> util logger`: turn on the lowest level of the logrus logger, so we can limit the level output of "util logger".

gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #1225, opened 2023-08-29
Remove all legacy status of entity


alexcodelf:
1. replace all `mixin.LegacyStatus()` to `mixin.Status`
2. check status change of the entity, update status with definede status type.
3. Adjust UI
4. clean unused code of legacy status.

currently, service revision used legacy status.

rainfd:
The original `status` of service revision is moved to `status->summaryStatus`, and the `statusMessage` is renamed to `record`.
Update the `History Revisions` and `log detail` in the service detail page please. @hibig

rainfd:
Can't get service revision summary status from the interface v1/projects/477380128381337604/environments/477380544053641220/services/477383247014133764/revisions?watch=true

Finenyaco:
**Environment**
- Walrus version: v0.3.1-rc1

**Test Result:** pass

Finenyaco:
**Environment**
- Walrus version: main ed21adcc5770196dc4a5654502d4a98a0987a631
- UI version: dev-1270234

**Test Result:** pass
<img width="2440" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/85c705ce-3a39-41e5-b09f-1c9660ff4627">

PR #1256, opened 2023-09-06
feat: remove all legacy status of entity

rainfd:
**Problem:**

**Solution:**

1. replace all mixin.LegacyStatus with mixin.Status
2. replace service_revision status_message string field with record text field.
3. clean unused code of legacy status.

migration sql
```sql
-- Modify "service_revisions" table

-- Modify "status_message" column to "record" column
ALTER TABLE service_revisions ADD COLUMN "record" text NULL;

UPDATE service_revisions SET record = status_message WHERE status_message IS NOT NULL;

ALTER TABLE service_revisions DROP COLUMN status_message;

-- Modify "status" string column to jsonb column
ALTER TABLE service_revisions ADD COLUMN new_status jsonb;

UPDATE service_revisions
SET new_status = jsonb_build_object(
        'summaryStatus', COALESCE(status, 'Failed'),
        'error', status = 'Failed',
        'transitioning', status = 'Running'
    )::jsonb;

ALTER TABLE service_revisions DROP COLUMN status;
ALTER TABLE service_revisions RENAME COLUMN new_status TO status;
```

**Related Issue:**
https://github.com/seal-io/walrus/issues/1225


pkg/apis/dashboard/extension.go
```diff
@@ -185,26 +183,43 @@ func getServiceRevisionStatusStats(
...
-				Select(
-					sql.As(sql.Count(servicerevision.FieldStatus), "count"),
-					sql.As(groupBy, servicerevision.FieldCreateTime),
-					servicerevision.FieldStatus).
-				GroupBy(
-					groupBy,
-					servicerevision.FieldStatus)
-		}).
-		Scan(ctx, &counts)
+	// Ent doesn't support group by json field yet.
```
gitlawr, line 186: Can you try the following?
```
q.
Select(
	sql.As(sql.Count(servicerevision.FieldStatus), "count"),
	sql.As(groupBy, servicerevision.FieldCreateTime),
	sql.As("status ->> 'summaryStatus'", "summaryStatus")).
GroupBy(
	servicerevision.FieldCreateTime,
	"summaryStatus")
```
Let's avoid using raw sql if possible.

pkg/deployer/terraform/jobctrl.go
```diff
@@ -139,17 +141,17 @@ func (r JobReconciler) syncApplicationRevisionStatus(ctx context.Context, job *b
...
 	if job.Status.Failed > 0 {
 		r.Logger.Info("failed", "application-revision", appRevisionID)
-		revisionStatus = status.ServiceRevisionStatusFailed
+		status.ServiceRevisionStatusReady.False(appRevision, "")
 	}
 
 	// Report to application revision.
-	appRevision.Status = revisionStatus
-	appRevision.StatusMessage = revisionStatusMessage
+	appRevision.Record = logs
```
gitlawr, line 148: Should we name the field `logs`?

pkg/deployer/terraform/deployer.go
```diff
@@ -396,7 +401,7 @@ func (d Deployer) createRevision(
 		return nil, err
 	}
 
-	if prevEntity != nil && prevEntity.Status == status.ServiceRevisionStatusRunning {
+	if prevEntity != nil && status.ServiceRevisionStatusReady.IsUnknown(prevEntity) {
```
alexcodelf, line 404: Previously, the if condition is equals to a running status. Here, the state is ready

pkg/apis/dashboard/extension.go
rainfd: Not working. error message column "status ->> 'summaryStatus'" does not exist

pkg/dao/types/status/status_servicerevision.go
```diff
@@ -0,0 +1,40 @@
+package status
+
+const (
+	ServiceRevisionStatusReady ConditionType = "Ready"
```
alexcodelf, line 4: Only ready status here is too weak to represent different states of revision. Ready used to represent final status.

pkg/deployer/terraform/jobctrl.go
rainfd: replace with name "record"

pkg/dao/types/status/status_servicerevision.go
rainfd: Currently, service revision only contain 3 status, running, succeed and failed, correspond to the 3 types of the ServiceRevisionStatusReady condition one by one. May be add more conditions or status in the future, but not in this pull request.

pkg/deployer/terraform/deployer.go
rainfd: The unknown ServiceRevisionStatusReady condition is stand for the service revision is created but not finished, it means the "running state" as before. More conditions or status has more precise meaning. It's explained below.

gitlawr (approved):
LGTM

Issue #1048, opened 2023-08-11
Incorrect catalog status when synchronizing templates timeout


Finenyaco:
**Describe the bug**

The catalog status is incorrect when synchronizing templates timeout.

**To Reproduce**

1. Add a catalog.
2. Wait to synchronize template.
<img width="1357" alt="image" src="https://github.com/seal-io/seal/assets/26084756/4db74b0d-3751-4399-93ba-569cc9d0b0e4">

3. The timeout is too short, and it remains in Initializing state after timeout.
<img width="1397" alt="image" src="https://github.com/seal-io/seal/assets/26084756/5d139c46-01c5-46ee-94ef-83df411d6cf8">

```
2023-08-11T06:34:45.564Z	E	catalog	failed to update sync info: context deadline exceeded
github.com/seal-io/seal/pkg/catalog.Sync.func1
	github.com/seal-io/seal/pkg/catalog/catalog.go:110
github.com/alitto/pond.(*WorkerPool).executeTask
	github.com/alitto/pond@v1.8.3/pond.go:454
github.com/alitto/pond.worker
	github.com/alitto/pond@v1.8.3/worker.go:32
2023-08-11T06:34:45.574Z	E	catalog	failed to update catalog cloudposse status: context deadline exceeded
github.com/seal-io/seal/pkg/catalog.Sync.func1
	github.com/seal-io/seal/pkg/catalog/catalog.go:119
github.com/alitto/pond.(*WorkerPool).executeTask
	github.com/alitto/pond@v1.8.3/pond.go:454
github.com/alitto/pond.worker
	github.com/alitto/pond@v1.8.3/worker.go:32
```

**Expected behavior**

**Environment**
- Seal version: 0265239705f6f6ba511c842b466ae843cc94e775

**Additional context**

alexcodelf:
Currently not reproducible, need to keep watching

Finenyaco:
**Environment**
- Walrus version: main 7dddda0f4b8d278b4731528af87dbf93f8a17d97
- UI version: dev-42d12aa

Trigger catalog synchronization via CatalogTemplateSync task. Sometimes it may time out and the catalog stuck in `Initializing` state.

<img width="2476" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/fa82ae86-2aac-4b0b-8a48-2fcf30d4fcaf">


```
[...]

PR #1259, opened 2023-09-06
fix(catalog): remove timeout and add logs

alexcodelf:
**Problem:**
- Sync an template with tags more than 100 may take more than 5 min to update template version, remove timeout.
- The logs of sync an catalog are too brief, it is hard to monitor the progress of sync an catalog.

The reason is the cron task will set timeout with previous task time. While syncing catalog, the timeout limit will be longer than the calculation after setting a short crontab expression, it may lead to timeout.

**Solution:**
- Remove timeout of sync template.
- Add debug logs of sync an catalog.
- Use a longer task timeout, using a new context to update catalog status.

**Related Issue:**

#1048

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -3,6 +3,7 @@ package catalog
 import (
 	"context"
 	"fmt"
+	"sync/atomic"
 	"time"
 
 	"github.com/drone/go-scm/scm"
@@ -113,7 +114,13 @@ func SyncTemplates(ctx context.Context, mc model.ClientSet, c *model.Catalog) er
 
 	logger.Infof("found %d repositories in %s", len(repos), c.Source)
 
-	wg := gopool.Group()
+	var (
+		total     = len(repos)
+		processed = int32(0)
+		failed    = int32(0)
+
+		wg = gopool.Group()
+	)
 
 	batchSize := 10
 	for i := 0; i < batchSize; i++ {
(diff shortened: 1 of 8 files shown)


pkg/templates/git.go
```diff
@@ -431,15 +432,16 @@ func getValidVersions(
 			Mode:   git.HardReset,
 		})
 		if err != nil {
-			logger.Warnf("failed to reset to tag %s: %v", tag, err)
+			logger.Warnf("failed set to %s/%s: %v", templateName, tag, err)
```
thxCode, line 435: failed to set %s/%s

pkg/templates/git.go
```diff
@@ -286,7 +286,7 @@ func SyncTemplateFromGitRepo(
 	defer func() {
 		rerr := updateTemplateStatus(ctx, mc, entity, err)
 		if rerr != nil {
-			logger.Errorf("failed to update template status: %v", rerr)
+			logger.Errorf("failed to update template %s status: %v", entity.Name, rerr)
```
thxCode, line 289: id is better than name.

thxCode (commented):
can we unify the info in the format: `"catalog.Name/template.Name:template.Version"`?

alexcodelf:
> can we unify the info in the format: `"catalog.Name/template.Name:template.Version"`?

template may not belongs to a catalog.

thxCode:
> > can we unify the info in the format: `"catalog.Name/template.Name:template.Version"`?
> 
> template may not belong to a catalog.

use "template.Name:template.Version"

pkg/catalog/catalog.go
```diff
@@ -135,15 +142,22 @@ func SyncTemplates(ctx context.Context, mc model.ClientSet, c *model.Catalog) er
...
-						fmt.Errorf("error syncing template %s: %w",
-							repo.Name, serr))
+						fmt.Errorf("error syncing \"%s:%s\" of catalog %q: %w",
+							c.Name, repo.Name, c.ID, serr))
+
+					atomic.AddInt32(&failed, 1)
+				} else {
+					atomic.AddInt32(&processed, 1)
 				}
+
```
aiwantaozi, line 158: total: %d

pkg/templates/git.go
alexcodelf: update to %s:%s

aiwantaozi (approved):

staging/utils/cron/cron.go
```diff
@@ -216,19 +216,23 @@ func (in *scheduler) Schedule(jobName string, cron Expr, task Task, taskArgs ...
...
+		next = ceParsed.Next(now).Sub(now) * 3
 	)
 
-	if next > atLeast {
-		next >>= 1
-	}
-
-	if next < atLeast {
+	switch {
+	case next < 0:
```
thxCode, line 232: next <= 0 ?

aiwantaozi (approved):

gitlawr (approved):
LGTM

Issue #1257, opened 2023-09-06
Fail to connect to the resource terminal (status code 1011) and sometimes panic


Finenyaco:
**Environment**
- Walrus version:  main 7dddda0f4b8d278b4731528af87dbf93f8a17d97
- UI version: dev-ec890a0

**Steps to reproduce**
1. Add a template with source `https://github.com/walrus-tutorials/helm-chart?ref=v0.0.1`.
2. Use the template to create a service. Fill in the `chart_url` https://github.com/walrus-tutorials/assets/releases/download/microservices-demo/sock-shop-0.3.0.tgz.
3. Go to the service details page. Connect to the terminal of queue-master.

**Result**

1. Failed to connect to the terminal. Check the logs, it returned status code 1011 instead of 1003.

<img width="2528" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/a8b28372-e5a3-4b78-b26b-273bbdc079f1">

2. And sometimes fail to connect to the terminal will cause panic.
```
2023-09-06T09:44:35.283Z	E	api	error processing bidirectional stream request: error streaming exec: command terminated with exit code 126
github.com/seal-io/walrus/pkg/apis/runtime.doBidiStreamRequest
	github.com/seal-io/walrus/pkg/apis/runtime/router_stream.go:239
github.com/seal-io/walrus/pkg/apis/runtime.(*Router).Routes.func1
	github.com/seal-io/walrus/pkg/apis/runtime/router_route.go:139
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/auths.Account.Filter
	github.com/seal-io/walrus/pkg/auths/account.go:56
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.RequestCounting.func3
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_flowcontrol.go:58
github.com/seal-io/walrus/pkg/apis/runtime.Per.func1
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_filter.go:84
github.com/seal-io/walrus/pkg/apis/runtime.If.func1
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_filter.go:58
github.com/gin-gonic/gin.(*Context).Next
[...]

PR #1264, opened 2023-09-08
fix: failed connect resource terminal

thxCode:
**Problem:**

Close the resize channel before processing the resize read, which often, happens on `shell` negotiate phase.

**Solution:**

- Do not close the resize channel and rely on the go runtime GC.
- Always send `exit` in case of session leaking.
- Modify the return error from Terminal `Read` to reduce below tedious logs.
```
E       websocket: close 1005 (no status)
E       websocket: close 1001 (going away)
E       read [::443] -> [::xxx] use of closed network connection
```

**Related Issue:**
#1257


alexcodelf (approved):
LGTM

aiwantaozi (approved):

PR #1278, opened 2023-09-12
fix: failed exec with docker runtime

thxCode:
**Problem:**

Failed to execute into the Pod based on Docker Runtime.

**Solution:**

- Filter out a SPDY v4 exit error with 126 code, and consider it as an unresolved exec shell request.
- Make it easy to develop with a new deployer runtime.

**Related Issue:**
#1257


pkg/server/init_deployer_runtime.go
```diff
@@ -0,0 +1,205 @@
...
+			Create(ctx, &rb, meta.CreateOptions{})
+		if err != nil {
+			return err
+		}
+	case !reflect.DeepEqual(rb.RoleRef, rb_.RoleRef):
+		// Delete.
+		err = cli.RbacV1().
+			RoleBindings(rb.Namespace).
+			Delete(ctx, rb.Name, meta.DeleteOptions{
+				PropagationPolicy: point(meta.DeletePropagationForeground),
```
gitlawr, line 173: Can we get rid of the point func in this way: https://github.com/kubernetes/kubernetes/blob/35199e42a41f83520e640fbfa9f409516faa7501/cmd/kubeadm/app/phases/upgrade/health.go#L200

alexcodelf (dismissed):
LGTM

aiwantaozi (approved):

pkg/server/init_deployer_runtime.go
thxCode: updated

aiwantaozi (approved):

Issue #1276, opened 2023-09-12
Failed to get field values for perspectives: converting NULL to string is unsupported


Finenyaco:
**Steps to reproduce**
1. New a perspective.
2. Select `Namespace` in costs source filter.

**Result**

1. An error was returned when clicking the drop-down list : `failed to get field values for perspectives`. `Node`, `Controller` and `Controller Kind` also have this problem.
<img width="2521" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/5d22b192-914b-483f-9d19-c94434d5bed9">

2. View logs.
```
2023-09-12T07:23:47.038Z	E	api	error requesting GET /v1/perspectives/_/field-values?endTime=2023-09-12T14%3A49%3A26%2B00%3A00&fieldName=namespace&fieldType=filter&startTime=2023-09-06T00%3A00%3A00%2B00%3A00: 1 error occurred:
	* sql/scan: failed scanning rows: sql: Scan error on column index 0, name "namespace": converting NULL to string is unsupported


github.com/seal-io/walrus/pkg/apis/runtime.erroring
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_error.go:45
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.recovering
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_recovery.go:39
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.observing
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_observation.go:89
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/gin-gonic/gin.(*Engine).handleHTTPRequest
	github.com/gin-gonic/gin@v1.9.1/gin.go:620
github.com/gin-gonic/gin.(*Engine).ServeHTTP
	github.com/gin-gonic/gin@v1.9.1/gin.go:576
github.com/seal-io/walrus/pkg/apis/runtime.(*Router).ServeHTTP
	github.com/seal-io/walrus/pkg/apis/runtime/router.go:191
net/http.serverHandler.ServeHTTP
	net/http/server.go:2947
net/http.initALPNRequest.ServeHTTP
[...]

PR #1287, opened 2023-09-15
fix: cost field value query cause crash

aiwantaozi:
**Problem:**
cost field value query cause crash

**Solution:**

add SQL COALESCE

**Related Issue:**

#1276


pkg/apis/perspective/extension.go
```diff
@@ -152,6 +153,30 @@ func (h Handler) CollectionRouteGetFieldValues(
...
+	SELECT
+	  DISTINCT COALESCE(namespace, '')
+	FROM
+	  "cost_reports"
+	WHERE
+	  "start_time" >= '2023-09-09T00:00:00Z'
+	  AND "end_time" <= '2023-09-15T16:16:51Z'
+	*/
+
+	fieldStr = fmt.Sprintf(`COALESCE(%s, '')`, req.FieldName)
```
alexcodelf, line 166: what will happened if the req.FieldName is not the column name of table `cost_reports`

aiwantaozi: The view `Validate` already checks the fieldName should be the builtin filter fields

alexcodelf (approved):
LGTM

thxCode (approved):
LGTM

Issue #1281, opened 2023-09-13
API key names can have the same name


Finenyaco:
**Steps to reproduce**
1. Go to `User Center` and add a api key.
2. Add another api key with the same name.

**Result**

1. The names are the same, can't tell the difference.

<img width="1515" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/c45bc8f9-6725-4828-9889-a3869dbf223c">

**Expected behavior**

1. Cannot add api keys with the same name.

**Environment**
- Walrus version: main 829ac7a2a65c2ee5092f994e81f8b41214e376a4
- UI version: dev-e2db642

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main de80628d3f1e14c28c190b9afea1608170721bc8
- UI version: dev-efebf61

**Test Result:** pass

<img width="817" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/e16e41fc-b647-4848-8bf6-b82a9b120d13">

PR #1290, opened 2023-09-15
fix: can create API keys with duplicate names

shanewxy:
**Problem:**

User can add another api key with the same name as an exisiting API key.

**Solution:**

- Set the database `name` field of the `token` table to be `unique`
- Regenerate the database migration code

**Related Issue:**
#1281


alexcodelf:
Can you make two commit?  one for modify schema, another is generated files.

shanewxy:
> Can you make two commit? one for modify schema, another is generated files.

updated

aiwantaozi (approved):

alexcodelf (approved):
LGTM

Issue #1267, opened 2023-09-11
Sometimes the list of template versions is out of order


Finenyaco:
**Steps to reproduce**
1. Add a template `https://github.com/terraform-aws-modules/terraform-aws-app-runner`
2. View the template details.

**Result**

1. Sometimes the version list is out of order.
<img width="2488" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/3d3f4a08-2aaf-4763-8704-957a3d388ecb">

**Expected behavior**

Sort by version in descending order.

**Environment**
- Walrus version: main ed21adcc5770196dc4a5654502d4a98a0987a631
- UI version: dev-1270234

**Additional context**

shanewxy:
Can we add sort to the link to solve it through frontend request to avoid backend hard coding? @hibig 
e.g. `https://localhost/v1/templates/478552182208082510/versions?sort=-version`

Finenyaco:
**Environment**
- Walrus version: main 597f55b631e6dcbd9725c4b140a9dda22eaaf035
- UI version: dev-281154c

**Test Result:** pass

<img width="2545" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/cfd5f41d-f84d-447a-80ec-58ec4e37982c">


<img width="2547" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/4c097f3b-65cc-488d-834d-5991e649d915">

PR #1292, opened 2023-09-18
fix: template versions sometimes incorrect order

shanewxy:
**Problem:**

Template versions with same `create_time` is ordered randomly instead of by version number.

**Solution:**

- Add `version` number as a subsequent column in default sorting.
- Convert the `version` strings into int arrays to avoid string sorting problem like v0.0.10 sorted before v0.0.2.
- Remove the customized sorting params, since it only allowed `version` field anyway.

**Related Issue:**
#1267

### DIFF
--- a/pkg/apis/template/extension.go
+++ b/pkg/apis/template/extension.go
@@ -1,6 +1,8 @@
 package template
 
 import (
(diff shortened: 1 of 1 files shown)


thxCode (approved):
LGTM

Issue #985, opened 2023-07-21
Support for cloning environment variables when cloning an environment


Finenyaco:
**Version:** main	
**UI:** method
**Commit ID:** 68e121d64cab4d59254b9f8b69d3c5e4aafe4321
**Env:** AWS EC2 Ubuntu 22.04 amd64

Currently cloning an environment does not clone environment variables, it will cause a service deployment to fail if the service in the cloned environment references an environment variable.

<img width="917" alt="image" src="https://github.com/seal-io/seal/assets/26084756/1a16dc98-4a2e-4e09-b16d-a4a787cc2741">

Finenyaco:
**Environment**
- Walrus version: main 597f55b631e6dcbd9725c4b140a9dda22eaaf035
- UI version: dev-3b0b2d4

**Test Result:** pass

<img width="2560" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f8f162ac-a54d-4827-956d-a3fb922de0ef">

PR #1293, opened 2023-09-20
feat: Support cloning environment variables when cloning an environment

shanewxy:
**Problem:**

Support cloning environment variables when cloning an environment.

**Solution:**

- Modify API to support variables inputs when creating environments.
- Add variable validation and creation logic 

**Related Issue:**
#985


rainfd (dismissed):
LGTM

pkg/apis/environment/basic.go
```diff
@@ -46,6 +46,24 @@ func (h Handler) Create(req CreateRequest) (CreateResponse, error) {
...
+		}
+
+		variables, err := tx.Variables().CreateBulk().
+			Set(variableInputs...).
+			Save(req.Context)
+		if err != nil {
+			return err
+		}
+
+		entity.Edges.Variables = variables
```
alexcodelf, line 65: Can you use a function `EnvironmentVariablesEdgeSave` like `dao.EnvironmentConnectorsEdgeSave` and use it in SaveE?

shanewxy: Moved variables save logic to dao. Moving service creation to dao will cause import cycle, so leave it for now.

alexcodelf (approved):
LGTM

Issue #1227, opened 2023-08-29
Return more information when the variable validation failed


Finenyaco:
**Steps to reproduce**
1. Add a template with source `https://github.com/terraform-google-modules/terraform-google-kubernetes-engine`.
2. Create a service using the templated added in step1.

**Result**

1. Failed to create the service: `invalid variables: unexpected value parsed: missing expected {`
<img width="2063" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/2d080aec-9da9-4c41-908d-9f6a14f78ee3">

2. But cannot find out which variable has the problem.

**Expected behavior**

Return more information when the variable validation failed.

**Environment**
- Walrus version: b51994e3fdba0aaeaba9439f6c609c105e5cbb92

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 597f55b631e6dcbd9725c4b140a9dda22eaaf035
- UI version: dev-281154c

**Test Result:** pass

<img width="1266" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/ff62352c-44ad-4dbd-be19-059266136572">

<img width="1440" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/1e7a33f9-91bc-40e3-a62e-dc2880e7b5c4">

PR #1294, opened 2023-09-21
feat: add variable name to validation error message

shanewxy:
**Problem:**

When an error occurs during variable parsing, the error message lacks the specific failed variable which makes it hard to diagnose for templates with a large number of variables, like `https://github.com/terraform-google-modules/terraform-google-kubernetes-engine`.

**Solution:**

Added corresponding variable name to the variable validation's error message.

**Related Issue:**
#1227

### DIFF
--- a/pkg/dao/types/property/field.go
+++ b/pkg/dao/types/property/field.go
@@ -488,7 +488,7 @@ func (i Properties) Cty() (cty.Type, cty.Value, error) {
 	for x := range i {
 		t, v, err := i[x].Cty()
 		if err != nil {
-			return t, v, err
+			return t, v, fmt.Errorf("error parsing property %q: %w", x, err)
 		}
 		ot[x] = t
 		ov[x] = v


pkg/dao/types/property/field.go
```diff
@@ -488,7 +488,7 @@ func (i Properties) Cty() (cty.Type, cty.Value, error) {
 	for x := range i {
 		t, v, err := i[x].Cty()
 		if err != nil {
-			return t, v, err
+			return t, v, fmt.Errorf("error parsing variable \"%s\": %w", x, err)
```
gitlawr, line 491: `error parsing property %q`
It knows little about variable in the context.

shanewxy: updated

gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #1283, opened 2023-09-13
Sometimes unable to get aws resources


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Use `aws-rds` to create a service.
2. Check the service resources.

**Result**

1. Display only one resource `aws_db_parameter_group`.
<img width="2456" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/3d6878e8-896f-48df-8d85-6d5862c761b0">

2. Check the api response. It returned a resource as well.
<img width="915" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/47a20fb7-18c0-4767-bba3-9e07f078ae66">

3. Check terraform output file. Some resource instances are empty.
```
{
  "version": 4,
  "terraform_version": "1.4.5",
  "serial": 1,
  "lineage": "9a715b76-bf19-c9de-6346-8637bc5103e9",
  "outputs": {
    "aws-rds_db_driver": {
      "value": "mysql",
      "type": "string"
    },
    "aws-rds_db_name": {
      "value": "rdsdb",
      "type": "string"
    },
    "aws-rds_db_password": {
      "value": "xxx",
      "type": "string",
      "sensitive": true
    },
    "aws-rds_db_port": {
      "value": 3306,
      "type": "number"
    },
    "aws-rds_db_username": {
      "value": "rdsusr",
      "type": "string"
    }
  },
  "resources": [
    {
      "module": "module.aws-rds",
      "mode": "data",
      "type": "aws_availability_zones",
      "name": "rds",
      "provider": "provider[\"registry.terraform.io/hashicorp/aws\"].connector--477956833806424834",
      "instances": [
        {
          "index_key": 0,
          "schema_version": 0,
          "attributes": {
            "all_availability_zones": null,
            "exclude_names": null,
            "exclude_zone_ids": null,
            "filter": null,
            "group_names": [
              "ap-east-1"
            ],
            "id": "ap-east-1",
            "names": [
              "ap-east-1a",
              "ap-east-1b",
              "ap-east-1c"
            ],
            "state": "available",
            "timeouts": null,
            "zone_ids": [
              "ape1-az1",
              "ape1-az2",
              "ape1-az3"
            ]
          },
          "sensitive_attributes": []
        }
      ]
    },
    {
      "module": "module.aws-rds",
      "mode": "data",
      "type": "aws_subnets",
      "name": "rds",
[...]

PR #1295, opened 2023-09-22
fix: service resources not updated when redeploying

shanewxy:
**Problem:**

When deploying services fails for the first time(e.g. Fail to create VPC), the instance of some resources are empty. Upgrading the service does not cause the instance field updated.

**Solution:**

When getting a new service revision, in addition to resource creation and deletion, update the instance Edge for the existing resources as well.

**Related Issue:**
#1283

### DIFF
--- a/pkg/apis/servicerevision/extension.go
+++ b/pkg/apis/servicerevision/extension.go
@@ -155,15 +155,21 @@ func manageResources(
 		return err
 	}
 
-	// Calculate creating list and deleting list.
+	// Calculate creating list, deleting list and updating list.
 	observedRessIndex := dao.ServiceResourceToMap(observedRess)
 
 	deleteRessIDs := make([]object.ID, 0, len(recordRess))
 
+	updatedRess := make([]*model.ServiceResource, 0, len(recordRess))
+
 	for _, c := range recordRess {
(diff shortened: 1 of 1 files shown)


rainfd (commented):
Why is this operation not in the transaction?

pkg/apis/servicerevision/extension.go
rainfd, line 1: Why is this operation not in the transaction?

pkg/apis/servicerevision/extension.go
```diff
@@ -163,7 +163,15 @@ func manageResources(
 	for _, c := range recordRess {
 		k := dao.ServiceResourceGetUniqueKey(c)
 		if observedRessIndex[k] != nil {
+			c.Edges.Instances = observedRessIndex[k].Edges.Instances
+
+			err = dao.ServiceResourceInstancesEdgeSave(ctx, modelClient, c)
```
rainfd, line 168: Why is this operation not in the transaction?

shanewxy: Updated

rainfd (approved):
LGTM

alexcodelf (approved):
LGTM

PR #1300, opened 2023-09-26
feat: add clone environment api

shanewxy:
**Problem:**

The original API used for cloning environment is the same as creating environment, which cannot handle the case of cloning sensitive variables.

**Solution:**

Add a specific clone environment API to decouple with the creation API, which queries and fills in the value of sensitive variables.

**Related Issue:**
#985

### DIFF
--- a/pkg/apis/environment/extension.go
+++ b/pkg/apis/environment/extension.go
@@ -3,6 +3,10 @@ package environment
 import (
 	"entgo.io/ent/dialect/sql"
 
+	"github.com/seal-io/walrus/pkg/dao/model/variable"
+	"github.com/seal-io/walrus/pkg/dao/types/crypto"
+	"github.com/seal-io/walrus/utils/errorx"
+
 	"github.com/seal-io/walrus/pkg/dao"
 	"github.com/seal-io/walrus/pkg/dao/model"
 	"github.com/seal-io/walrus/pkg/dao/model/service"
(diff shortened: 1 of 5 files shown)


pkg/apis/environment/extension.go
```diff
@@ -150,3 +153,25 @@ func getCaps(entities model.Services) (int, int) {
...
+	entity := req.Model()
+
+	for i := range req.Variables {
+		v := entity.Edges.Variables[i]
+		if v == nil {
+			return nil, errorx.New("invalid input: nil variable")
+		}
+
+		if v.Sensitive && v.Value == "" {
+			v.Value = h.modelClient.Variables().Query().
```
gitlawr, line 167: Can we query vars once instead of querying one for each var?

shanewxy: updated

gitlawr (approved):
LGTM

pkg/apis/environment/extension.go
```diff
@@ -150,3 +154,44 @@ func getCaps(entities model.Services) (int, int) {
...
+	}
+
+	// Fetch and fill the value of sensitive variables from the cloned environment.
+	if len(variableNames) > 0 {
+		vs := h.modelClient.Variables().Query().
+			Where(
+				variable.EnvironmentID(req.CloneEnvironmentId),
+				variable.NameIn(variableNames...),
+			).
+			AllX(req.Context)
```
alexcodelf, line 182: using All(req.Context) instead of AllX

pkg/apis/environment/extension_view.go
```diff
@@ -27,3 +33,91 @@ type (
...
+
+		CloneEnvironmentId object.ID `path:"environment"`
+
+		model.EnvironmentCreateInput `path:",inline" json:",inline"`
+	}
+
+	RouteCloneEnvironmentResponse = model.EnvironmentOutput
+)
+
+func (r *RouteCloneEnvironmentRequest) Validate() error {
```
alexcodelf, line 49: It seems that the clone request validate is similar to the create environment validate, can you using a common validate func for two request validation?

pkg/apis/environment/extension.go
shanewxy: updated

pkg/apis/environment/extension_view.go
shanewxy: updated

alexcodelf (approved):
LGTM

Issue #1193, opened 2023-08-24
Support non-catalog template option to the catalog filter of Templates tab


linyinli:
**Is your enhancement related to a problem? Please describe.**
When catalog templates and non-catalog templates are mixed together, it's difficult to find the non-catalog templates.

**Describe the solution you'd like**
Support for adding non-catalog template options to the catalog filter.

Finenyaco:
**Environment**
- Walrus version: main 068e70303c857a262eb48aa2f7f59bec0073015b
- UI version: dev-3542d28

**Test Result:** pass

<img width="1851" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/3388b145-707a-4a2a-b5a4-4fbc1db8de7f">

PR #1301, opened 2023-09-27
feat: support non-catalog template filter

shanewxy:
**Problem:**

When catalog templates and non-catalog templates are mixed together, it's difficult to find the non-catalog templates. 

**Solution:**

Since not providing `catalogId` is already used for querying all templates, add a `nonCatalog` param to fetch non-catalog templates only.

**Related Issue:**
#1193


gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #606, opened 2023-05-08
Support project scope templates


gitlawr:
So that different projects can use different set of templates.

Finenyaco:
Verified in main 3a128a4de94303c68057ac66aabedc3806cc6040

PR #1305, opened 2023-10-08
feat: support project scope template

shanewxy:
**Problem:**

Support project scope catalogs and templates.

**Solution:**

- Add `project_id` field to catalogs and templates, empty for global.
- Add `withGlobal` param in template querying to include global templates used for service creation.
- Add `template_id` to template_versions index constraint, as name and version may be duplicated across projects.

**Related Issue:**
#606


aiwantaozi (approved):

gitlawr (approved):
LGTM

Issue #1296, opened 2023-09-25
Failed to initialize the catalog when the certificate of catalog server is not valid 


linyinli:
**Describe the bug**
Failed to initialize the catalog when the certificate of catalog server is not valid.

**Steps to reproduce**
1. Go to 'Operation Hub -> Catalogs', add a catalog, using Gitlab type.

**Result**
Failed to initialize the catalog, as shown below:
<img width="1448" alt="image" src="https://github.com/seal-io/walrus/assets/42955482/db0b06bd-d9a5-47e8-a37e-5e7c3624985f">

**Environment**
- Walrus version: v0.3.1

Finenyaco:
The gitlab server with self-signed certificates is not currently supported.
<img width="2483" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/fae3cad4-5f21-4c63-9fc5-0e1b700b3967">

Finenyaco:
**Environment**
- Walrus version: main 700176d7801e24b89a6f2f308db96436be6be8e5
- UI version: dev-43c5e24

**Test Result:** pass

After enabling `Skip certificate authentication` in system settings, it can initialize the catalog with a self-signed certificate.
<img width="557" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/57dbb0ed-f8da-4957-816f-7ee868becdaf">

PR #1310, opened 2023-10-13
fix: self-signed certificates catalog not supported

shanewxy:
**Problem:**

Self-signed certificates is not supported in both catalog fetching and terraform deploying.

**Solution:**

- Support configuring skip SSL certificates verification in global settings `SSLSkipVerify`.
- Config InsecureSkipVerify to the `http.Client` fetches catalogs if `SSLSkipVerify` is set.
- Set `GIT_SSL_NO_VERIFY` environment variable for local template cloning and service deploying job if `SSLSkipVerify` is set.

**Related Issue:**
#1296

### DIFF
--- a/pkg/deployer/terraform/deployer.go
+++ b/pkg/deployer/terraform/deployer.go
@@ -234,7 +234,7 @@ func (d Deployer) createK8sJob(ctx context.Context, opts createK8sJobOptions) er
 		return err
 	}
 
-	jobEnv, err := d.getProxyEnv(ctx)
+	jobEnv, err := d.getEnv(ctx)
 	if err != nil {
 		return err
 	}
@@ -250,7 +250,7 @@ func (d Deployer) createK8sJob(ctx context.Context, opts createK8sJobOptions) er
 	return CreateJob(ctx, d.clientSet, jobOpts)
 }
 
-func (d Deployer) getProxyEnv(ctx context.Context) ([]corev1.EnvVar, error) {
+func (d Deployer) getEnv(ctx context.Context) ([]corev1.EnvVar, error) {
 	var env []corev1.EnvVar
 
 	allProxy, err := settings.DeployerAllProxy.Value(ctx, d.modelClient)
@@ -298,6 +298,13 @@ func (d Deployer) getProxyEnv(ctx context.Context) ([]corev1.EnvVar, error) {
 		env = append(env, corev1.EnvVar{
 			Name:  "NO_PROXY",
 			Value: noProxy,
+		})
+	}
+
+	if settings.SkipRemoteTLSVerify.ShouldValueBool(ctx, d.modelClient) {
+		env = append(env, corev1.EnvVar{
+			Name:  "GIT_SSL_NO_VERIFY",
+			Value: "true",
 		})
 	}
 
--- a/pkg/settings/settings.go
+++ b/pkg/settings/settings.go
@@ -135,6 +135,12 @@ var (
 		editable,
 		initializeFrom("true"),
 		modifyWith(notBlank))
+	// SkipRemoteTLSVerify indicates whether skip SSL verification when accessing remote server.
+	SkipRemoteTLSVerify = newValue(
+		"SkipRemoteTLSVerify",
+		editable,
+		initializeFrom("false"),
+		modifyWith(notBlank))
 )
 
 // the built-in settings for server cron jobs.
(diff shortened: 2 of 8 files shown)


pkg/templates/git.go
```diff
@@ -456,3 +461,46 @@ func getValidVersions(
...
+		err := SetGitCAEnvVar(ctx, m.TransactionalModelClient)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+// SetGitCAEnvVar sets environment variable GIT_SSL_CAINFO from settings.
```
gitlawr, line 486: Does it work cloning from github.com when GIT_SSL_CAINFO is set?

shanewxy: Yes, setting `GIT_SSL_CAINFO` extends the CA certificate trust chain used by Git, but it doesn't override the system's default certificates. GitHub's SSL certificates are signed by a public CA and are generally trusted by default.

shanewxy:
Updated based on today's discussion

pkg/deployer/terraform/deployer.go
```diff
@@ -239,6 +241,8 @@ func (d Deployer) createK8sJob(ctx context.Context, opts createK8sJobOptions) er
 		return err
 	}
 
+	jobEnv = d.addGitEnv(ctx, jobEnv)
+
```
thxCode, line 245: can we merge the `getProxyEnv` and `addGitEnv` together? like `getEnv`?

pkg/templates/git.go
```diff
@@ -6,6 +6,9 @@ import (
 	"os"
 	"path/filepath"
 
+	settingbus "github.com/seal-io/walrus/pkg/bus/setting"
+	"github.com/seal-io/walrus/pkg/settings"
+
```
thxCode, line 11: fix `goimport` please, and move the new imports to the bottom.

pkg/settings/settings.go
```diff
@@ -135,6 +135,12 @@ var (
 		editable,
 		initializeFrom("true"),
 		modifyWith(notBlank))
+	// SSLSkipVerify indicates whether skip SSL verification when accessing remote server.
+	SSLSkipVerify = newValue(
```
thxCode, line 139: it seems we are doing skip "insecure" tls cert verification, can we use something like`DisableInsecureVerify` or `SkipInsecureVerify`?

pkg/templates/git.go
```diff
@@ -456,3 +461,37 @@ func getValidVersions(
...
+	// Set environment variable GIT_SSL_NO_VERIFY.
+	err := os.Setenv(GitSSLSkipVerifyKey, settings.SSLSkipVerify.ShouldValue(ctx, mc))
+	if err != nil {
+		return err
+	}
+
+	log.WithName("git").Infof("set environment variable %s", GitSSLSkipVerifyKey)
+
+	return nil
+}
```
thxCode, line 497: the getter client has an option for insecure verification skipping, but the GitGetter implementation does not respect this option. I think we can fork the [go-getter](https://github.com/hashicorp/go-getter) from hashicorp,  implement that and push to upstream.

https://github.com/hashicorp/go-getter/blob/0298a221674f629339295fa8a1e6a938e28506e0/get_git.go#L206-L207

pkg/deployer/terraform/deployer.go
shanewxy: updated

pkg/templates/git.go
shanewxy: updated

pkg/settings/settings.go
shanewxy: updated

pkg/templates/git.go
shanewxy: updated

thxCode (approved):
LGTM

pkg/settings/settings.go
```diff
@@ -135,6 +135,12 @@ var (
 		editable,
 		initializeFrom("true"),
 		modifyWith(notBlank))
+	// SkipInsecureVerify indicates whether skip SSL verification when accessing remote server.
+	SkipInsecureVerify = newValue(
```
gitlawr, line 139: What about change to SkipRemoteTLSVerify

shanewxy: updated

aiwantaozi (approved):

Issue #1297, opened 2023-09-25
Failed to reset tag when sync template


linyinli:
**Describe the bug**
An error was reported while refreshing the catalog:
```
Failed to reset to tag <some_tag>: object not found
```

**Steps to reproduce**
1. Go to 'Operation Hub -> Catalogs', add a catalog by using `https://github.com/terraform-aws-modules`;
2. Watching the log of walrus container.

**Result**
There are some errors occur:
<img width="522" alt="image" src="https://github.com/seal-io/walrus/assets/42955482/8b8ac929-2833-47da-892d-744e3723ddf7">

**Environment**
- Walrus version: v0.3.1

Finenyaco:
**Environment**
- Walrus version: main 96c9574ea9fee1bad195827be7db27d629f416a6
- UI version: dev-5976040

**Test Result:** pass

PR #1311, opened 2023-10-17
fix: failed to reset to tag when syncing templates

shanewxy:
**Problem:**

The hash reference of the tag sometimes points to the tag itself, instead of the commit. 

**Solution:**

Obtain an `*object.Tag`, then find the target commit in the Target field.

**Related Issue:**
#1297

### DIFF
--- a/pkg/templates/git.go
+++ b/pkg/templates/git.go
@@ -428,8 +428,16 @@ func getValidVersions(
 			continue
 		}
 
+		hash := resetRef.Hash()
+
+		// If tag is not a commit hash, get commit hash from tag object target.
+		object, err := r.TagObject(hash)
+		if err == nil {
+			hash = object.Target
+		}
+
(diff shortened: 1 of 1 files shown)


gitlawr (approved):
LGTM

aiwantaozi:
Can you provide more details on what situations will point to tag itself?

shanewxy:
> Can you provide more details on what situations will point to the tag itself?

Tags can be of two types: lightweight tags and annotated tags.

Lightweight tags are simply references to specific commits.The hash of a lightweight tag points directly to the commit. 

Annotated tags contain additional information such as tag messages, the name of the creator, and a creation date. The hash of an annotated tag points to the tag object itself, which contains the additional information. The tag object, in turn, points to the specific commit it is tagging.

alexcodelf (approved):
LGTM

Issue #1302, opened 2023-09-27
Failed to validate list(may(any)) type


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Add a template with source `https://github.com/terraform-google-modules/terraform-google-kubernetes-engine`.
2. Using this template to create a service. Fill in the required fields and save.

**Result**

1. Returned an error: `invalid variables: unexpected value parsed: error parsing property \"node_pools\": missing expected {`
<img width="2537" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/ce1b6148-6d25-44bd-b0b1-4f6c11c44ad0">

2. After changing the type to `list(may(string))`, the service was created successfully.

**Expected behavior**

Create the service successfully.

**Environment**
- Walrus version:  main 597f55b631e6dcbd9725c4b140a9dda22eaaf035
- UI version: dev-281154c

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main aeb8ea4107b44284c3485fab8c60668c0b128716
- UI version: dev-5976040

Test Result: pass

PR #1312, opened 2023-10-17
fix: failed to validate (any) elements nested in compound types

shanewxy:
**Problem:**

When the property type is a nested compound type, it fails to validate elements with type `any`.

**Solution:**

Use the builtin function of go-cty to check dynamic types recursively.

**Related Issue:**
#1302

### DIFF
--- a/pkg/dao/types/property/field.go
+++ b/pkg/dao/types/property/field.go
@@ -425,20 +425,6 @@ func (i Property) Cty() (cty.Type, cty.Value, error) {
 	}
 
 	switch {
-	// From Terraform expression guide,
-	// https://developer.hashicorp.com/terraform/language/expressions/type-constraints#collection-types,
-	// map(any), list(any), set(any) are the compatible expansions for map, list, set.
-	case (i.Type.IsMapType() && cty.DynamicPseudoType.Equals(*i.Type.MapElementType())) ||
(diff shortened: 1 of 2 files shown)


pkg/dao/types/property/field.go
```diff
@@ -425,29 +425,24 @@ func (i Property) Cty() (cty.Type, cty.Value, error) {
...
 	switch {
-	// From Terraform expression guide,
-	// https://developer.hashicorp.com/terraform/language/expressions/type-constraints#collection-types,
-	// map(any), list(any), set(any) are the compatible expansions for map, list, set.
-	case (i.Type.IsMapType() && cty.DynamicPseudoType.Equals(*i.Type.MapElementType())) ||
-		(i.Type.IsListType() && cty.DynamicPseudoType.Equals(*i.Type.ListElementType())) ||
-		(i.Type.IsSetType() && cty.DynamicPseudoType.Equals(*i.Type.SetElementType())):
-		// When the property type is very loose,
-		// only perform conversion checks on its value.
+	case i.Type == cty.NilType || i.Type == cty.DynamicPseudoType:
```
gitlawr, line 428: Can we add unit tests to ensure it works in different cases?

shanewxy: updated

thxCode (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1329, opened 2023-10-26
Fail to create service when project & global templates share the same name


gitlawr:
**Steps to reproduce**
1. Add `https://github.com/walrus-catalog` catalog to the default project
2. Create a webservice

**Result**
Pop the error `model: template_version not singular`

**Environment**
- Walrus version: main

Finenyaco:
**Environment**
- Walrus version: main 9c5ea546b1918a53aaf1a3173a5f43b10dd5fe3e
- UI version: dev-19c51ab

**Test Result:** pass

No error returned.
<img width="1697" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f808a5ed-3b67-4e6f-9bbd-241b9b03dad8">

PR #1332, opened 2023-10-27
fix: create service fail to validate when using templates with conflict names

shanewxy:
**Problem:**

TemplatedVersion's unique constraint used to be on `name` and `version`, modified to `template_id` when supporting project templates, which causes requests not providing `template_id` fail to validate.

**Solution:**

Modify unique constraint to be `name`, `version` and `project_id` where `project_id` is not null, and `name`, `version` when it is null. 

Service creation request format:
```
...
template:
  name: webservice
  version: v0.0.3
  project:
    name: dev
...
```
or
```
...
template:
  name: webservice
  version: v0.0.3
  project:
    id: xxx
...
```
**Related Issue:**
#1329


gitlawr (approved):
LGTM

thxCode (approved):
LGTM

Issue #1306, opened 2023-10-09
Support editing template schema in the UI


gitlawr:
**Is your enhancement related to a problem? Please describe.**
Now we need to do this in version control system. If we can edit the schema in the UI, it's simpler to tweak exposed capabilities of a template.

gitlawr:
Verified in main

PR #1334, opened 2023-10-27
feat: support custom service UI

aiwantaozi:
**Feature:**
1. Allow the user to customize the service and deploy display UI by editing the schema.
2. The UI is somewhat dependent on Terraform's types during deployment. 

**Solution:**
1. Add openapi schema to define the variable and UI display configuration, allowing the user to customize the service deploy display UI by editing the schema
2. Decoupled the terraform types dependency, support UI render based on the openapi schema and without requiring knowledge of Terraform types.

**Related Issue:**
#1306


pkg/apis/setup.go
```diff
@@ -95,6 +96,7 @@ func (s *Server) Setup(ctx context.Context, opts SetupOptions) (http.Handler, er
 		r.Routes(setting.Handle(opts.ModelClient))
 		r.Routes(subject.Handle(opts.ModelClient))
 		r.Routes(template.Handle(opts.ModelClient))
+		r.Routes(templateversion.Handle(opts.ModelClient))
```
shanewxy, line 99: Now we've supported project scope template, may need add it to project's SubResourceHandler as well?

pkg/templates/git.go
```diff
@@ -138,14 +142,21 @@ func syncTemplateFromRef(
 		return err
 	}
 
-	schema, err := loadTerraformTemplateSchema(w.Filesystem.Root())
+	// Try to get version from reference.
+	ref := repo.Reference
+	if v, err := version.NewVersion(repo.Reference); err == nil {
+		ref = v.Original()
```
shanewxy, line 148: What's this operation used for? `v.Original()` seems to be the same as the reference passed in.

aiwantaozi: Use the original ref to prevent duplicated, this is the original PR add this change the first time. https://github.com/seal-io/walrus/pull/1135

pkg/dao/types/template_version.go
```diff
@@ -0,0 +1,150 @@
...
+	return s.Schema.Components.Schemas["variables"].Value
+}
+
+// InternalTemplateVersionSchema include the internal template variables schema and template data.
+type InternalTemplateVersionSchema struct {
+	// Schema specifies the openapi schema of input variables.
+	Schema *openapi3.T `json:"schema"`
+	// Data specifies the data of this template.
+	Data *TemplateVersionSchemaData `json:"data"`
+}
```
gitlawr, line 56: Can we flatten the fields to 
```
type Schema struct{
    OpenAPIV3Schema *openapi3.T
    
    Readme string `json:"readme"`
    RequiredProviders []ProviderRequirement `json:"requiredProviders"`
    
}
```

pkg/apis/connector/basic_view.go
```diff
@@ -25,6 +29,13 @@ type (
 )
 
 func (r *CreateRequest) Validate() error {
+	// Set default configData schema.
+	if r.ConfigData.GetSchema() == nil {
+		if s := getConfigDataSchema(r.Category, r.Type); s != nil {
+			r.ConfigData.Schema = *s
+		}
+	}
```
gitlawr, line 37: Why so many changes are needed around connectors? How do we use configdata schema? In my understanding, configData are basically map[string]string with sensitive bool.

pkg/templates/loader/testdata/full_schema/schema.yaml
```diff
@@ -0,0 +1,76 @@
+openapi: 3.0.3
+info:
+  title: dev
+  version: v1.0.0
+components:
+  schemas:
+    variables:
```
gitlawr, line 7: image

CRD schema is abbreviated to description and properties. Do you think we can do in similar way? Does it affect UI rendering?

pkg/templates/openapi/ext.go
```diff
@@ -0,0 +1,320 @@
...
+	ExtUIHidden    = "hidden"
+	ExtUIImmutable = "immutable"
+	ExtUIWidget    = "widget"
+	ExtUISequence  = "sequence"
+
+	// Extension for original value.
+	ExtOriginal                = "x-walrus-original"
+	ExtOriginalLanguage        = "language"
+	ExtOriginalType            = "type"
+	ExtOriginalValueExpression = "value-expression"
```
gitlawr, line 19: Do you think we can put everything in the same ext field like:
```yaml
x-walrus:
  ui:
    group: xxx
  original:
    type: xx
    language: xxx
```

How are language/type/value-expression being used?

pkg/dao/schema/connector.go
```diff
@@ -69,12 +69,9 @@ func (Connector) Fields() []ent.Field {
...
 			Annotations(
 				entx.Input(entx.WithQuery())),
-		field.String("config_version").
-			Comment("Connector config version.").
-			NotEmpty(),
-		crypto.PropertiesField("config_data").
+		crypto.PropertyField("config_data").
 			Comment("Connector config data.").
-			Default(crypto.Properties{}).
+			Default(crypto.Property{}).
```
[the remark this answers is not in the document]
aiwantaozi: After the updated, the value could be map[string]any, and the schema could describe the value is a map, so we could replace the map[string]Property with Property.

```go
	// Property holds the schema and underlay value of the property.
	Property struct {
		// Schema specifies the schema of this property.
		Schema openapi3.Schema `json:"schema"`
		// RawValue specifies the value of this property.
		RawValue Value `json:"value,omitempty"`
	}
```

pkg/dao/schema/template_version.go
```diff
@@ -61,9 +61,14 @@ func (TemplateVersion) Fields() []ent.Field {
...
-			Comment("Schema of the template.").
-			Default(&types.TemplateSchema{}),
+		field.JSON("internalSchema", types.InternalTemplateVersionSchema{}).
+			Comment("Generated schema and data of the template.").
+			Default(types.InternalTemplateVersionSchema{}),
+		field.JSON("externalSchema", types.ExternalTemplateVersionSchema{}).
+			Comment("User customized schema of the template.").
+			Default(types.ExternalTemplateVersionSchema{}).
+			Annotations(
+				entx.Input(entx.WithUpdate())),
```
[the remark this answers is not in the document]
aiwantaozi: updated to schema and customizeOpenAPISchema

pkg/dao/types/template_version.go
aiwantaozi: updated

pkg/apis/connector/basic_view.go
aiwantaozi: updated

pkg/templates/loader/testdata/full_schema/schema.yaml
aiwantaozi: after communicating, keep the openAPI format and metadata

pkg/templates/openapi/ext.go
aiwantaozi: Prefer to keep the original and UI separated, so UI and backend could focus on the extension they care. Several extensions with the same prefix are common way. https://swagger.io/docs/specification/openapi-extensions/

pkg/dao/schema/template_version.go
```diff
@@ -61,9 +61,14 @@ func (TemplateVersion) Fields() []ent.Field {
 			Comment("Source of the template.").
 			NotEmpty().
 			Immutable(),
-		field.JSON("schema", &types.TemplateSchema{}).
-			Comment("Schema of the template.").
-			Default(&types.TemplateSchema{}),
+		field.JSON("schema", types.Schema{}).
+			Comment("Generated schema and data of the template.").
+			Default(types.Schema{}),
+		field.JSON("customizeOpenAPISchema", types.CustomizeOpenAPISchema{}).
```
gitlawr, line 67: s/customize/customized/

pkg/apis/template/extension.go
```diff
@@ -84,5 +84,22 @@ func (h Handler) RouteGetVersions(
...
+		if !v.CustomizeOpenAPISchema.IsEmpty() || v.Schema.IsEmpty() {
+			continue
+		}
+
+		entities[i].CustomizeOpenAPISchema = entities[i].Schema.
+			Expose()
+	}
+
+	return model.ExposeTemplateVersions(entities)
 }
```
gitlawr, line 105: Can we merge CustomizedOpenAPISchema to entity.schema, and skip output of CustomizedOpenAPISchema in the API?

pkg/deployer/terraform/deployer.go
```diff
@@ -806,85 +808,98 @@ func getModuleConfig(
...
-	sensitiveVariableRegex, err := matchAnyRegex(sensitiveVariables.UnsortedList())
-	if err != nil {
-		return nil, err
+		for n, v := range variablesSchema.Properties {
+			// Add sensitive from schema variable.
+			if v.Value.WriteOnly {
+				sensitiveVariables.Insert(fmt.Sprintf(`var\.%s`, n))
+			}
+
+			// Add seal metadata.
```
gitlawr, line 849: walrus metadata

pkg/templates/openapi/ext.go
```diff
@@ -0,0 +1,315 @@
+package openapi
+
+import (
+	"github.com/getkin/kin-openapi/openapi3"
+)
+
+const OpenAPIVersion = "3.0.3"
```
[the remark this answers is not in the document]
aiwantaozi: 3.0.x is more widely used because most of the tools have not yet released stable support for version 3.1.0. Swagger only started supporting version 3.1 this year, and the OpenAPI Generator has not yet released stable support for 3.1.0.

pkg/apis/template/extension.go
aiwantaozi: Prefer using separated schemas and customizing the OpenAPI schema. 

At the API level, for list/get operations, you can simply ignore the details by using 'extract=-schema' and focus on viewing the customized OpenAPI schema. For the update operation, user should only set the 'customizeOpenAPISchema' and disregard the others. Both the list and put APIs use the same object, and the openapi schema is sufficient, eliminating the need for additional documentation.

At the database level, there is no need to modify the schema when updating the 'customizeOpenAPISchema'.

pkg/dao/schema/template_version.go
```diff
@@ -61,9 +61,14 @@ func (TemplateVersion) Fields() []ent.Field {
...
 			NotEmpty().
 			Immutable(),
-		field.JSON("schema", &types.TemplateSchema{}).
-			Comment("Schema of the template.").
-			Default(&types.TemplateSchema{}),
+		field.JSON("schema", types.Schema{}).
+			Comment("Generated schema and data of the template.").
+			Default(types.Schema{}),
+		field.JSON("uiSchema", types.UISchema{}).
+			Comment("User customized schema of the template.").
```
gitlawr, line 68: update the comment

pkg/dao/types/template_version.go
```diff
@@ -0,0 +1,153 @@
...
+	"fmt"
+
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/hashicorp/terraform-config-inspect/tfconfig"
+
+	"github.com/seal-io/walrus/pkg/templates/openapi"
+)
+
+// UISchema include the schema of template variables user customize.
+type UISchema struct {
```
gitlawr, line 14: `type UISchema openapi3.T`
And update comment

gitlawr (approved):
LGTM

shanewxy (approved):

gitlawr (approved):

Issue #965, opened 2023-07-10
[Feature Request]Support customized service icons in dependency graph using template icons


linyinli:
Version: v0.2.0

For user request, it's recommended to support customized service icons in dependency graph using template icons, to show server services, middleware services, database services and other services clearly.

<img width="1589" alt="image" src="https://github.com/seal-io/seal/assets/42955482/f8c8ca0b-6c0e-4521-b093-0724d84eda7d">

Finenyaco:
**Environment**
- Walrus version: main 6d4903951fc8b6284473bfbae8fe8f0fea34ecd4
- UI version: dev-d45e562

**Test Result:** pass
<img width="1566" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/13f719ff-cd3e-4d9a-a479-eda2d23f74d1">

PR #1346, opened 2023-11-01
feat: return template icons in dependency graph response

shanewxy:
**Problem:**

The service icons show in dependency graph are always the default icon.

**Solution:**

Support service customized icon by using the corresponding template icon.

**Related Issue:**
#965


alexcodelf (approved):
LGTM

Issue #1355, opened 2023-11-06
Object type variables do not display properly on forms


Finenyaco:
**Describe the bug**

Object type variables do not display properly on forms.

**Steps to reproduce**
1. Add a template with source `https://github.com/unitong/test-variable`. The variable `test_object` is defined as follows,

```
variable "test_object" {
  type          = object({
    name        = string
    age         = number
    email       = optional(object({
        address = string
        domain  = string
    }))
  })
  default       = {
    name        = "Bob"
    age         = 23
    email       = {
        address = "bob"
        domain  = "example.com"
    }
  }
}
```

**Result**

1. Properties in the object are displayed directly on the form and it is not clear which field they belong to. And the default values are not displayed.
<img width="2515" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/722c2b69-54a9-468f-90e7-350719c360c3">

2. Checked the schema.
```
        test_object:
          default:
            age: 23
            email:
              address: bob
              domain: example.com
            name: Bob
          properties:
            age:
              title: age
              type: number
            email:
              properties:
                address:
                  title: address
                  type: string
                domain:
                  title: domain
                  type: string
              required:
                - address
                - domain
              title: email
              type: object
            name:
              title: name
              type: string
          required:
            - age
            - name
          title: test_object
          type: object
```

**Expected behavior**

1. User can find the `test_object` field and clearly see the structure of the object type variables.

**Environment**
- Walrus version: main 6d4903951fc8b6284473bfbae8fe8f0fea34ecd4
- UI version: dev-d45e562

**Additional context**

Finenyaco:
It cannot display the complex type `list(map(any)` as well.

Variable definition:
```
variable "list_map_any" {
  type        = list(map(any))
  description = "List of maps containing node pools"

  default = [
    {
      name = "default-node-pool"
    },
  ]
}
```

Display on the form：
<img width="1247" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/ed71f579-cca5-463f-af33-b4fb91203747">

Checked the api response:
<img width="998" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/7cb13599-9bae-4571-80fd-0cd5d17f3075">

shanewxy:
**Environment**

Walrus version: v0.4.0-rc2 af772a7ae049670658fdb2d2249cfde51277db8e
UI version: v0.4.0-rc2

**Test Result**: pass

[...]

PR #1357, opened 2023-11-06
support cli for generate schema and fix bugs

aiwantaozi:
**Problem:**
1. Bug: Could not import template include option(type, default)
2. Bug: Need to support the template only define default without type
3. Bug: Sync template from catalog renders incorrect schema
4. Support CLI for generating schema.yaml 

**Solution:**
1. Support import template with option(type, default)
2. Support generate type from default
3. Skip the incorrectly cached schema while syncing templates from catalog
4. Support CLI command ui-schema generate

**Related Issue:**
https://github.com/seal-io/walrus/issues/1355
https://github.com/seal-io/walrus/issues/1362
#1365


pkg/templates/loader/loader.go
```diff
@@ -8,15 +8,31 @@ import (
 
 // SchemaLoader define the interface for loading schema from template.
 type SchemaLoader interface {
-	Load(rootDir, templateName, templateVersion string) (*types.Schema, error)
+	Load(rootDir, templateName string, mode Mode) (*types.Schema, error)
```
gitlawr, line 11: Whether it's from schema file or not is the implementation detail of SchemaLoader. Why do we expose it to the caller

aiwantaozi: Expose to support different modes.
- Load from file: mainly used in walrus server. While walrus sync template schema, when there is a schema.yaml, will use the schema user defined in schema.yaml directly.
- Merge: mainly used in CLI, while user already defined a customized schema.yaml, and then updated the varaibles.tf this mode will regenerate the schema from variables.tf and merge the result with the schema from schema.yaml, this mode will include both new generated and user customized.
- Original: only return the generated, ignore the customized.

cmd/cli/command.go
```diff
@@ -146,6 +154,50 @@ func NewConfigCmd() *cobra.Command {
 	return configCmd
 }
 
+// NewUISchemaCmd generate ui schema command.
+func NewUISchemaCmd() *cobra.Command {
```
gitlawr, line 158: Why UISchema instead of Schema?

aiwantaozi: updated

pkg/templates/loader/loader.go
```diff
@@ -26,5 +42,13 @@ func LoadSchema(rootDir, templateName, templateVersion string) (s *types.Templat
...
 	// Continue with other loaders in the future.
-	return nil, fmt.Errorf("no supported schema found for template %s version %s", templateName, templateVersion)
+	return nil, fmt.Errorf("no supported schema found for template %s", templateName)
 }
+
+type Mode string
+
+const (
+	ModePreferFromFile Mode = "preferFromFile"
+	ModeFromOriginal   Mode = "fromOriginal"
```
gitlawr, line 52:
```
ModeSchemaFile
ModeOriginal
ModeMerge
```

gitlawr (dismissed):
LGTM

pkg/templates/loader/loader.go
```diff
@@ -26,5 +42,13 @@ func LoadSchema(rootDir, templateName, templateVersion string) (s *types.Templat
 	}
 
 	// Continue with other loaders in the future.
-	return nil, fmt.Errorf("no supported schema found for template %s version %s", templateName, templateVersion)
+	return nil, fmt.Errorf("no supported schema found for template %s", templateName)
 }
+
+type Mode string
+
+const (
```
alexcodelf, line 50: can we make the mode define style as same as https://github.com/seal-io/walrus/blob/main/pkg/apis/server.go#L52-L56?

alexcodelf (approved):
LGTM

Issue #1374, opened 2023-11-08
Support for stop running workflows and deleting records


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

**Describe the solution you'd like**

Currently, it can not abort running workflows and delete records.

<img width="2501" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/46ebc7a7-d7f0-40d2-a353-7d240e5a97a5">

**Describe alternatives you've considered**

**Additional context**

shanewxy:
**Environment**

Walrus version: v0.4.0-rc2
UI version: v0.4.0-rc2

**Test Result**: pass

PR #1380, opened 2023-11-09
support workflow parameters and add workfow controller

alexcodelf:
**Problem:**
- Support parameters when running a workflow
- The workflow running is slow due to too much http step to sync status.
- Approval results are not reset after re-run workflow (Reason: the approval data not cleared)
- Cannot re-approve when approval fails  (Reason: using model.Client not mode.Tx to do transaction)
- Add APIs to Support for stop running workflows and deleting records
- Unable to get log of completed tasks before workflow ends(status changed completed before logs set to database)

**Solution:**
- Support setting workflow execution parameters when running a workflow.
- Using workflow controller to update workflow execution status.
- Fix using wrong model client to commit transaction
- Add apis to support enhance functions.
- record step logs before status chagned.

**Related Issue:**
#1374 #1382 #1389 #1403


pkg/dao/types/workflow.go
```diff
@@ -17,9 +18,50 @@ const (
...
+		return errors.New("invalid input: empty name")
+	}
+
+	if c.Value == "" {
+		return errors.New("invalid input: empty value")
+	}
+
+	// Mustache syntax is not allowed.
+	// Argo workflow will try parse the value as a template parameter like
+	// {{workflow.parameters.name}}.
```
gitlawr, line 47: Can we call it variables and use it in the format ${workflow.var.name}?

pkg/workflow/reconcile.go
```diff
@@ -0,0 +1,79 @@
...
+	err := r.KubeClient.Get(ctx, req.NamespacedName, wf)
+	if err != nil {
+		if kerrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+
+		return ctrl.Result{}, err
+	}
+
+	r.Logger.Info("workflow", "name", wf.Name, ",status", wf.Status.Phase)
```
gitlawr, line 38: Should we use Debug

pkg/dao/types/workflow.go
alexcodelf: updated

pkg/workflow/reconcile.go
alexcodelf: updated

pkg/dao/schema/workflow.go
```diff
@@ -58,6 +59,9 @@ func (Workflow) Fields() []ent.Field {
 			Default(0).
 			Annotations(
 				entx.SkipInput()),
+		field.JSON("variable_configs", types.VariableConfigs{}).
```
gitlawr, line 62: What about name this field "variables" and the type "WorkflowVariable"? I wonder if we need another concept since we already used this term in other places like template schema.

alexcodelf: updated to 
```
field.JSON("variables", types.WorkflowVariables{}).
			Comment("Configs of workflow variables.").
			Optional(),

```

gitlawr (approved):
LGTM

Issue #1389, opened 2023-11-09
Cannot re-approve when approval fails


Finenyaco:
**Describe the bug**

Cannot re-approve when approval fails

**Steps to reproduce**
1. Create a workflow and add an approval task.
2. Execute the workflow.
3. Click Approve as soon as it is ready for approval.

**Result**

1. Failed to approve and returned an error. 

```
2023-11-09T10:27:18.549Z	W	Failed to resume	{"name": "flow-id-20231109180351-486530644638171138", "error": "currently, set only targets suspend nodes: no suspend nodes matching nodeFieldSelector: templateName=suspend-486530644655013890"}
2023-11-09T10:27:18.549Z	E	api	error requesting POST /v1/projects/486372827121844226/workflows/486530637608517634/executions/486530644638171138/stage-executions/486530644654948354/step-executions/486530644655013890/approve: 1 error occurred:
	* rpc error: code = Internal desc = currently, set only targets suspend nodes: no suspend nodes matching nodeFieldSelector: templateName=suspend-486530644655013890


github.com/seal-io/walrus/pkg/apis/runtime.erroring
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_error.go:45
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.recovering
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_recovery.go:39
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.observing
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_observation.go:89
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/gin-gonic/gin.(*Engine).handleHTTPRequest
	github.com/gin-gonic/gin@v1.9.1/gin.go:620
github.com/gin-gonic/gin.(*Engine).ServeHTTP
	github.com/gin-gonic/gin@v1.9.1/gin.go:576
github.com/seal-io/walrus/pkg/apis/runtime.(*Router).ServeHTTP
	github.com/seal-io/walrus/pkg/apis/runtime/router.go:191
net/http.serverHandler.ServeHTTP
	net/http/server.go:2947
[...]

Issue #1385, opened 2023-11-09
Incorrect status of workflow step when a user doesn't have permission to upgrade the service


Finenyaco:
**Describe the bug**

Incorrect status of workflow step when a user doesn't have permission to upgrade the service

**Steps to reproduce**
1. Create a workflow and add a task to deploy a service to a production environment.
2. Execute workflow by admin user to create the service in production environment.
3. Execute workflow by a general user to upgrade the service.

<img width="2422" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/294c7403-a04d-4f02-b8de-37d8464b8cf8">
<img width="2454" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/7e45380b-0dc2-443f-ba45-716d58410857">

**Result**

1. The step status is `Completed`, but actually the deployment failed and there was no permission to upgrade that service.
<img width="2087" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/43c197a4-685b-4256-a546-daeeb037db5a">

**Expected behavior**

1. Display the correct status for workflow step.
2. Return more messages.

**Environment**
- Walrus version: main 065ee92e55e14044fa0777be4c631f7082fadc0e
- UI version: dev-81ad833

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main a3e5c0343d3e831ca30e4fd76a160212bc5bba7f
- UI version: dev-0f40ad5

**Test Result:** pass

<img width="1256" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/72fe7860-b50d-47a4-8b57-2bbf037e96b1">

Issue #1378, opened 2023-11-08
Project members in the Member role do not have permission to create workflows and approve tasks


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create user `test1` and add `test1` as a project member with the role `Member`.
<img width="2472" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/6d863f88-ba57-465c-b27e-ca0b0bbb29ee">

2. Create a workflow, add an approval step and add the user to approver.
3. Log in walrus as `test1`, go to the workflow and approve the task.

**Result**

1. Failed to approve, permission denied.
<img width="2555" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/1c9665bd-19ff-400b-ba52-f1cd74e0f225">

2. Project members in the Member role cannot create workflows.
<img width="2509" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/d774fabc-4e39-41e1-b83d-327a4de2be45">

**Expected behavior**

**Environment**
- Walrus version: main 065ee92e55e14044fa0777be4c631f7082fadc0e
- UI version: dev-b9a6538

**Additional context**

Finenyaco:
- Walrus version: main a3e5c0343d3e831ca30e4fd76a160212bc5bba7f
- UI version: dev-2338df1

**Test Result:** pass

Issue #1375, opened 2023-11-08
Rejection of approval task does not take effect


Finenyaco:
**Describe the bug**

Rejection of approval task does not take effect

**Steps to reproduce**
1. Create a workflow add an approval task.
2. Reject the approval task.

**Result**

1. It didn't work after clicking  on `Reject`. No request was sent.
<img width="2304" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/7b072ee6-b4d7-4c92-b069-d12ffc3ef056">

**Expected behavior**

Reject the approval task and the workflow status changes to failed.

**Environment**
- Walrus version: main 065ee92e55e14044fa0777be4c631f7082fadc0e
- UI version: dev-b9a6538

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main a3e5c0343d3e831ca30e4fd76a160212bc5bba7f
- UI version: dev-aab74c2

**Test Result:** pass

<img width="319" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/ed9caeb4-a9b3-455c-8616-56a29444406f">

PR #1396, opened 2023-11-10
Fix project members rbacs and approval step

alexcodelf:
**Problem:**
- workflow approvel steps for rejection does not have effect.
- project member do not have permission to manage workflows.
- workflow create resource step have special characters makes create or update resource failed.

**Solution:**

- fix approval rejection logic and record rejected users, using transaction to approve steps. #1389 #1375
- add project member rbacs . #1378
- update resource step to record failed information to mark step execution failed. #1385 #1373

**Related Issue:**
#1373, #1375, #1378, #1385, #1389


gitlawr (dismissed):
LGTM

pkg/dao/types/workflow.go
```diff
@@ -83,26 +86,46 @@ func NewWorkflowStepApprovalSpec(spec map[string]any) (*WorkflowStepApprovalSpec
...
-		users, err := toObjectIDs(v)
-		if err != nil {
-			return nil, err
+	for i := range userIndexes {
+		if _, ok := spec[userIndexes[i]]; !ok {
+			continue
+		}
+
+		if v, ok := spec[userIndexes[i]].([]any); ok {
+			users, err := toObjectIDs(v)
```
aiwantaozi, line 101: could you merge func toObjectIDs and removeDuplicatedUsers, since they always used together

pkg/dao/types/workflow.go
```diff
@@ -83,26 +86,46 @@ func NewWorkflowStepApprovalSpec(spec map[string]any) (*WorkflowStepApprovalSpec
...
 }
 
+func (s *WorkflowStepApprovalSpec) IsRejected() bool {
+	return len(s.RejectedUsers) > 0
+}
+
 func (s *WorkflowStepApprovalSpec) IsApproved() bool {
+	if s.IsRejected() {
+		return false
+	}
```
aiwantaozi, line 127: When user config or here, then one user approved and one user rejected, it should be approved right? But based on the code it would return not approved

alexcodelf: If one user reject the step, the step is rejected, no matter how many other users approved.

pkg/dao/types/workflow.go
alexcodelf: updated

aiwantaozi (approved):

Issue #964, opened 2023-07-10
[Feature Request] Show service number in environment list view


pengjiang80:
Version: v0.2
Description: Currently we only show env name, description, create time in env list view, which is not very helpful. User may want to know how many services have been deployed in each environment and the status of them. It will be helpful If we can show this overview information in the env list view. Currently users can only click and enter each env to check.
![image](https://github.com/seal-io/seal/assets/16434640/0b4cafe8-6c4e-4034-8c01-1723a4051cc1)

Finenyaco:
**Environment**
- Walrus version: 26ec3d124e49c0d0834e48b4e1e974d689c6a0e9
- UI version: dev-e291910

**Test Result:** 

<img width="2487" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/51b41c7b-ec6c-462c-aab2-28a3c334a4ee">

But the statistics cannot be automatically updated https://github.com/seal-io/walrus/issues/1454

PR #1406, opened 2023-11-13
feat: add resource status summary for environment

aiwantaozi:
**Problem:**

need to show resource/service status in the environment list

**Related Issue:**

#964

### DIFF
--- a/pkg/apis/environment/basic_view.go
+++ b/pkg/apis/environment/basic_view.go
@@ -8,6 +8,7 @@ import (
 	"github.com/seal-io/walrus/pkg/dao/model/environment"
 	"github.com/seal-io/walrus/pkg/dao/model/predicate"
 	"github.com/seal-io/walrus/pkg/dao/types/object"
+	"github.com/seal-io/walrus/pkg/dao/types/status"
 	"github.com/seal-io/walrus/utils/validation"
 )
 
@@ -26,9 +27,11 @@ func (r *CreateRequest) Validate() error {
 type (
 	GetRequest struct {
 		model.EnvironmentQueryInput `path:",inline"`
+
+		IncludeSummary bool `query:"includeSummary,omitempty"`
 	}
 
-	GetResponse = *model.EnvironmentOutput
+	GetResponse = *environmentOutput
 )
 
 type UpdateRequest struct {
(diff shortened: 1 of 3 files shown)


gitlawr:
How does it look with both resource & service now?

aiwantaozi:
> How does it look with both resource & service now?

Should like below, group by ready, transitioning and error
image

pkg/apis/environment/basic_view.go
```diff
@@ -65,9 +68,46 @@ type (
...
 )
 
 type CollectionDeleteRequest = model.EnvironmentDeleteInputs
+
+type environmentOutput struct {
+	model.EnvironmentOutput `json:",inline"`
+	StatusSummary           status.Count `json:"statusSummary"`
+}
+
+func exposeEnvironment(entity *model.Environment) *environmentOutput {
```
gitlawr, line 85: Can we move it to pkg/dao/expose.go

pkg/apis/environment/basic_view.go
```diff
@@ -65,9 +68,46 @@ type (
...
+
+	if len(entity.Edges.Resources) > 0 {
+		for _, v := range entity.Edges.Resources {
+			switch {
+			case v.Status.Error:
+				output.StatusSummary.Error++
+			case v.Status.Transitioning:
+				output.StatusSummary.Transitioning++
+			default:
+				output.StatusSummary.Ready++
```
gitlawr, line 98: A stopped resource is neither in error nor transitioning status, is it expected to count it as ready?

aiwantaozi: The stopped resource count as "ready" is confusing. What do you think about changing "ready" to "healthy" or "stable"?

pkg/apis/environment/basic_view.go
aiwantaozi: `func exposeEnvironment` only changes the expose API response object, and no other places need to use this function, prefer to keep it in the API level, like other expose functions (exposeSetting, exposeVariable)

pkg/apis/environment/basic_view.go
gitlawr: Ready, Error and Other?

gitlawr (approved):
LGTM

alexcodelf (approved):

Issue #1316, opened 2023-10-20
Support service draft and resource recycle


linyinli:
At present, we are facing the following problems:
- Currently, when deploying a service, we need to configure all of its variables at once before delivery. If there are numerous parameters or various details that require confirmation, it can become quite complex.
- During development, testing, or demonstrations, it is not possible to recycle a service instance and preserve the service's configuration for future reuse without the need for reconfiguration.
I believe that we need support for service draft and resource recycle to resolve these problems.

linyinli:
For more valuable use cases, such as rapidly setting up multiple environments for various tests within resource constraints, we need to enhance the service draft by introducing the capability to clone the draft only or clone and deploy simultaneously, regardless of whether the source of the clone is a draft or has already been deployed.

Finenyaco:
Verified in main d96e8a0e7bf4e2a7c68285cb47b204f4aa0c714d.

PR #1416, opened 2023-11-13
Support resource draft, stop and start

gitlawr:
**Related Issue:**
https://github.com/seal-io/walrus/issues/1316

### DIFF
--- a/pkg/apis/environment/extension_view.go
+++ b/pkg/apis/environment/extension_view.go
@@ -1,9 +1,22 @@
 package environment
 
 import (
+	"fmt"
+	"net/http"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/seal-io/walrus/pkg/dao"
 	"github.com/seal-io/walrus/pkg/dao/model"
+	"github.com/seal-io/walrus/pkg/dao/model/environment"
+	"github.com/seal-io/walrus/pkg/dao/model/project"
+	"github.com/seal-io/walrus/pkg/dao/model/resource"
+	"github.com/seal-io/walrus/pkg/dao/model/templateversion"
 	"github.com/seal-io/walrus/pkg/dao/types"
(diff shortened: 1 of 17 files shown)


pkg/apis/environment/extension_view.go
```diff
@@ -54,3 +71,125 @@ type (
...
+	for _, r := range resources {
+		if pkgresource.IsStoppable(r) {
+			stoppableResources = append(stoppableResources, r)
+			stoppableResourceIDs = append(stoppableResourceIDs, r.ID)
+		}
+	}
+
+	r.stoppableResources = stoppableResources
+
+	return apiresource.ValidateRevisionsStatus(r.Context, r.Client, stoppableResourceIDs...)
```
alexcodelf, line 194: What is the difference between start and upgrade when resource is Ready or Failed in this request?

pkg/apis/environment/extension_view.go
```diff
@@ -54,3 +71,125 @@ type (
...
+		All(r.Context)
+	if err != nil {
+		return err
+	}
+
+	stoppableResources := make(model.Resources, 0, len(resources))
+	stoppableResourceIDs := make([]object.ID, 0, len(resources))
+
+	for _, r := range resources {
+		if pkgresource.IsStoppable(r) {
```
alexcodelf, line 113: Are the Failed resources can be stopped?

pkg/apis/environment/extension_view.go
gitlawr: Update validation so that start only applies to undeployed/stopped resources.

pkg/apis/environment/extension_view.go
gitlawr: Update status validation for stop action

alexcodelf (approved):
LGTM

aiwantaozi (approved):

Issue #1400, opened 2023-11-10
Matching the wrong rule when matching multiple rules at the same time


Finenyaco:
**Describe the bug**

Matching the wrong rule when matching multiple rules at the same time

**Steps to reproduce**
1. New a resource definition add add two match rules. The selector of the first rule is is that the environment type is `Production`.
<img width="1091" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/6c7e6e7b-546b-4abc-9d56-f77ed6ebcb83">

2. The selector of the second rule is that the environment is `k18-prod`.

<img width="965" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/1e1ebc2f-f44c-43f4-95d5-e9b0b982c75d">

3. Create an environment called `k18-prod` with the `Production` environment type.
<img width="2503" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/92ecc7e6-30c2-48d9-bb29-c8057ec68ed2">

4. Click on env `k18-prod` and create a resource using the resource definition created in step 1.

**Result**

1. It matched the second rule.
<img width="1386" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/07637a3c-0ad1-40cf-bf63-b9757cec2065">

**Expected behavior**

1. When the resource matches more than one rule, it is expected to match the first rule.

**Environment**
- Walrus version: main 9e14c25ece772fbb582b0be93a0fdb5e486fd53e
- UI version: dev-e15ef2e
**Additional context**

Finenyaco:
**Environment**
- Walrus version: 4465aa8deb63ff1ceb270b6cbc1c5ae2852d8242
- UI version: dev-7b0f2a2

**Test Result:** pass

PR #1421, opened 2023-11-14
Preserve Resource Definition Matching Rule order

gitlawr:
**Problem:**
Matching rules' order is not kept.

**Solution:**
Store order in relation table and query matching rules by order.

**Related Issue:**
https://github.com/seal-io/walrus/issues/1400


aiwantaozi:
How was the scene handled :
1. User create definition with rule1(order is 1), rule2(order is 2), saved.
2. Edit the definition, delete rule1, add another rule A, and what is the order for rule A.
3. Need UI to handle the order or backend should handle it? If this is handled by UI, while several users editing the definition at the same time, the order may misconfigure

gitlawr:
> How was the scene handled :
> 
> 1. User create definition with rule1(order is 1), rule2(order is 2), saved.
> 2. Edit the definition, delete rule1, add another rule A, and what is the order for rule A.
> 3. Need UI to handle the order or backend should handle it? If this is handled by UI, while several users editing the definition at the same time, the order may misconfigure

1. rule1 is order 0, rule2 is order 1
2. rule2 is order 0, rule A is order 1
3. It's irrelevant to the API, only the underlying implementation detail.

aiwantaozi (approved):

Issue #1432, opened 2023-11-15
Failed to deploy services/resources: An argument definition must end with a newline


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create a service using `webservice` template.
2. Check the deployment results.

**Result**

1. Returned an error: An argument definition must end with a newline
<img width="936" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/a48d7fa4-a87b-4abf-9c4f-bcc045575f1d">

2. Check the main.tf. Output `service1_Service Ip` contains a space.

```
terraform {
  required_providers {
    kubectl = {
      source  = "gavinbunney/kubectl"
      version = "1.14.0"
    }
    kubernetes = {}
  }

  backend "http" {
    address                = "https://18.167.117.84/v1/projects/487397017002180610/environments/487397878596108290/resources/487398158960164866/revisions/487398158976942082/terraform-states"
    password               = "oa8YbJ8q1DBObW62SBwcPXDFLSYFlabGGHvBBHEol6TAcg6JU-1A3ual0BIs2OFDq9Z15tRlVzCuW0D3Ua8hKMMm0Ul88TZBJYJ74A8IfT-nc5943SEe3sUiOTSNisZeQzE"
    retry_max              = 10
    retry_wait_max         = 5
    skip_cert_verification = true
    update_method          = "PUT"
    username               = "seal"
  }
}

provider "kubectl" {
  alias       = "connector--487397823784943618"
  config_path = "/var/terraform/secrets/config487397823784943618"
}

provider "kubernetes" {
  alias       = "connector--487397823784943618"
  config_path = "/var/terraform/secrets/config487397823784943618"
}

module "service1" {
  image        = "nginx"
  limit_cpu    = ""
  limit_memory = ""
  name         = ""
  namespace    = ""
  ports        = [80]
  providers = {
    kubectl    = kubectl.connector--487397823784943618
    kubernetes = kubernetes.connector--487397823784943618
  }
  replicas                       = 1
  request_cpu                    = "0.001"
  request_memory                 = "10Mi"
  source                         = "git::https://github.com/walrus-catalog/webservice?ref=v0.0.4"
  walrus_metadata_namespace_name = ""
  walrus_metadata_service_name   = ""
}

output "service1_Ports" {
  sensitive = false
  value     = module.service1.Ports
}

output "service1_Service Ip" {
  sensitive = false
  value     = module.service1.Service Ip
}

```

**Expected behavior**

[...]

PR #1433, opened 2023-11-15
fix: generate output contains invalid character

alexcodelf:
**Problem:**
Can not apply service, due to invalid  terraform configs.

**Solution:**
Using output raw key to generate config. 

**Related Issue:**

#1432


gitlawr (approved):
LGTM

aiwantaozi (approved):

Issue #1339, opened 2023-10-31
Support adding templates in a subdirectory


gitlawr:
e.g., https://github.com/terraform-aws-modules/terraform-aws-cloudwatch/tree/master/modules/cis-alarms.

gitlawr:
Only apply to isolated templates, not for templates from a catalog.

Finenyaco:
**Environment**
- Walrus version: 3ccf6b03dc821e799b35b45d41e07165821beb7f
- UI version: dev-e291910

**Test Result:** pass
<img width="2492" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/d60ed072-231b-4e28-ba25-c5b5ac69d4a5">
<img width="1371" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f6d4531e-eb31-4b67-b148-d0d1a1ed6320">

PR #1434, opened 2023-11-16
feat: support subpath module import

alexcodelf:
**Problem:**
Can not import subdirectory template from remote repository.

**Solution:**
Support subdirectory template import.
Example:
`https://github.com/alexcodelf/webservice//sub`
`https://github.com/alexcodelf/webservice//sub?ref=main`

The subdirectory is following double slashes.
**Related Issue:**
#template
#1339

### DIFF
--- a/pkg/vcs/repo.go
+++ b/pkg/vcs/repo.go
@@ -31,9 +31,11 @@ type Repository struct {
 	Topics []string `json:"topics"`
 	// Driver is the driver of the repository. E.G: github.
 	Driver string `json:"driver"`
+	// SubPath is the sub path of the repository.
+	SubPath string `json:"subPath"`
 }
 
-// ParseURLToGit parses a raw URL to a git repository.
+// ParseURLToRepo parses a raw URL to a git repository.
 // Since the return repository only contains the namespace and name,
 // It only used for create template from catalog.
 func ParseURLToRepo(rawURL string) (*Repository, error) {
@@ -42,6 +44,7 @@ func ParseURLToRepo(rawURL string) (*Repository, error) {
 	ref := ""
 	name := ""
 	namespace := ""
+	subPath := ""
 
 	endpoint, err := transport.NewEndpoint(rawURL)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


pkg/vcs/repo.go
```diff
@@ -58,6 +61,15 @@ func ParseURLToRepo(rawURL string) (*Repository, error) {
 		rawURL = strings.TrimSuffix(rawURL, "?ref="+ref)
 	}
 
+	// Get sub path from path.
+	if strings.Contains(path, "//") {
+		paths := strings.Split(path, "//")
+		subPath = paths[1]
+
+		path = strings.TrimSuffix(path, "//"+subPath)
+		rawURL = strings.TrimSuffix(rawURL, "//"+subPath)
```
gitlawr, line 70: How do we handle sth like https://github.com/foo/bar//one//two?

alexcodelf: In what cases will use this case? If it is a deep dir, it should be `https://github.com/alexcodelf/webservice//dep/sub`

gitlawr: Not a valid use case. I ask because it will fail the parsing logic here. Do we have any validation to avoid this?

alexcodelf: updated, create template will check git source parse.

gitlawr (approved):
LGTM

aiwantaozi (dismissed):

pkg/vcs/repo.go
```diff
@@ -31,6 +31,8 @@ type Repository struct {
 	Topics []string `json:"topics"`
 	// Driver is the driver of the repository. E.G: github.
 	Driver string `json:"driver"`
+	// SubPath is the sub path of the repository.
+	SubPath string `json:"subPath"`
 }
 
 // ParseURLToGit parses a raw URL to a git repository.
```
aiwantaozi, line 38: ParseURLToGit -> ParseURLToRepo

aiwantaozi (approved):

Issue #1405, opened 2023-11-13
[CLI] Should not should maxprocs log in cli


aiwantaozi:
**Is your enhancement related to a problem? Please describe.**
<img width="1058" alt="image" src="https://github.com/seal-io/walrus/assets/5936271/5cf1ba31-47bf-4ef5-b174-8e7382f6f324">

We use maxprocs package, it will always show init logs, should not show these logs while use cli.

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

shanewxy:
**Environment**

Walrus version: v0.4.0-rc2 https://github.com/seal-io/walrus/commit/af772a7ae049670658fdb2d2249cfde51277db8e

**Test Result**: pass


![Image](https://github.com/seal-io/walrus/assets/45061862/0533b582-b68d-4a36-bc55-16987fddf554)

Issue #1015, opened 2023-08-04
[CLI] Hide backend logs on error


Finenyaco:
**Version:** main
**UI:** dev-1497e86
**Commit ID:** 7346fa3e2f32c2e7a19d3454f50e7842a11cfe1b
**Env:** AWS EC2 Ubuntu 22.04 amd64

The CLI command currently prints too many backend logs when an error occurs.

```
root@ip-192-168-1-101:~/.seal# seal -v
seal version main (7346fa3e2f32c2e7a19d3454f50e7842a11cfe1b)
```

```
root@ip-192-168-1-101:~/.seal# seal config setup --project-name pro2
2023-08-04T09:33:25.386Z	D	Error Stack: goroutine 1 [running]:
runtime/debug.Stack()
	runtime/debug/stack.go:24 +0x65
main.Run.func1()
	github.com/seal-io/seal/cmd/cli/main.go:32 +0x46
panic({0x962c60, 0xc0001c7180})
	runtime/panic.go:884 +0x212
main.NewConfigCmd.func2(0xc0001c5200?, {0x9f9d5f?, 0x2?, 0x2?})
	github.com/seal-io/seal/cmd/cli/command.go:108 +0x127
github.com/spf13/cobra.(*Command).execute(0xc0001c5200, {0xc000035040, 0x2, 0x2})
	github.com/spf13/cobra@v1.7.0/command.go:944 +0x847
github.com/spf13/cobra.(*Command).ExecuteC(0xc0001c4c00)
	github.com/spf13/cobra@v1.7.0/command.go:1068 +0x3bd
github.com/spf13/cobra.(*Command).Execute(...)
	github.com/spf13/cobra@v1.7.0/command.go:992
main.Run()
	github.com/seal-io/seal/cmd/cli/main.go:55 +0xee
main.main()
	github.com/seal-io/seal/cmd/cli/main.go:23 +0x77

error while execute: endpoint is required
```

```
root@ip-192-168-1-101:~/.seal# seal config setup -s https://xxx --project-name pro2 --token xxx
2023-08-04T09:35:39.481Z	D	Error Stack: goroutine 1 [running]:
runtime/debug.Stack()
	runtime/debug/stack.go:24 +0x65
main.Run.func1()
	github.com/seal-io/seal/cmd/cli/main.go:32 +0x46
panic({0x98fa60, 0xc00036ac30})
	runtime/panic.go:884 +0x212
[...]

PR #1437, opened 2023-11-16
fix: remove cli useless logs

aiwantaozi:
**Problem:**
CLI shows useless logs.

**Solution:**
Remove CLI useless logs

**Related Issue:**
#1015 
https://github.com/seal-io/walrus/issues/1405

### DIFF
--- a/staging/utils/runtimex/runtime.go
+++ b/staging/utils/runtimex/runtime.go
@@ -8,7 +8,7 @@ import (
 	"github.com/seal-io/walrus/utils/log"
 )
 
-func init() {
+func Init() {
 	_, _ = maxprocs.Set(maxprocs.Logger(log.WithName("maxprocs").Printf))
 }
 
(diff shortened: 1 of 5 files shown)


staging/utils/runtimex/runtime.go
```diff
@@ -8,7 +8,7 @@ import (
 	"github.com/seal-io/walrus/utils/log"
 )
 
-func init() {
+func Init() {
```
gitlawr, line 11: It is implicitly invoked before. Any impact of this change? Why do we need the change in runner.go?

aiwantaozi: This implicitly invoked will always print the log inside the Set function while using log package, like below:
`maxprocs        maxprocs: Leaving GOMAXPROCS=%s: CPU quota undefined`

gitlawr: Can we change Printf to Debugf?

gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #1436, opened 2023-11-16
Failed to update resource definition


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create a resource definition.
2. Edit the resource definition and save.
3. Check the configuration.

**Result**

1. There's one more rule. The old rule still stands.
<img width="2509" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/02a96e88-31cc-4c61-b9ec-4829cf2408cf">

2. Delete the old rule and save. But no response after clicking on save.

**Expected behavior**

Update resource definition successfully.

**Environment**
- Walrus version: 4465aa8deb63ff1ceb270b6cbc1c5ae2852d8242
- UI version: dev-d246b6d

**Additional context**

Finenyaco:
**Environment**
- Walrus version: 4465aa8deb63ff1ceb270b6cbc1c5ae2852d8242
- UI version: dev-7b0f2a2

**Test Result:** 
- Check the PUT resource-definitions request. There are two rules.
<img width="783" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/4bfd60f7-df17-4f1a-9158-05a853ecc656">

- Check the GET resource-definition response. There are four rules, including two old ones.
<img width="772" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f0c8bb3a-bb46-47dd-b392-42bca2fd4763">

Finenyaco:
**Environment**
- Walrus version: 3ccf6b03dc821e799b35b45d41e07165821beb7f
- UI version: dev-e291910

**Test Result:** failed

1. Create a resource definition, and add a webservice template with version 0.0.4.
2. Edit the resource definition and change template version to 0.0.5.
3. Check the rules after updating. Failed to update. The version is still 0.0.4.

<img width="903" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/76813874-0b5f-4e90-9f48-baf874208b95">

4. Check the PUT resource-definitions request.
<img width="781" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/a71a5cd1-41d4-4561-abee-593b8e8bf701">

5. Check the GET resource-definitions response.
<img width="876" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/3866bf4f-15c7-44cc-abe0-00a45e97dedd">

Finenyaco:
**Environment**
[...]

PR #1439, opened 2023-11-16
fix: update matching rule index

gitlawr:
**Problem:**
Definition ID + rule name is a not unique constraint. users may save a def with multiple rules using the same name.

**Solution:**
Update indexes.

**Related Issue:**

https://github.com/seal-io/walrus/issues/1436


alexcodelf (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #1435, opened 2023-11-16
Failure to upgrade service using deploy-source-code template when first deployment fails


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Use  template `deploy-source-code` version 0.0.6 to deploy a service and configure an incorrect registry username.
2. Correct the user name and upgrade.
3. Check the deployment results.

**Result**

1. Failed to deploy.
<img width="2432" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/b454415e-8268-4c69-a5ad-c20bf59b223f">

```
2023-11-16T06:45:31.640Z	E	api	error requesting PUT /v1/projects/487397017002180610/environments/487397878596108290/resources/487524735370919938/revisions/487524897606598658/terraform-states: 1 error occurred:
	* model: validator failed for field "ResourceComponentRelationship.resource_component_id": value is blank


github.com/seal-io/walrus/pkg/apis/runtime.erroring
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_error.go:45
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.recovering
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_recovery.go:39
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.observing
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_observation.go:89
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/gin-gonic/gin.(*Engine).handleHTTPRequest
	github.com/gin-gonic/gin@v1.9.1/gin.go:620
github.com/gin-gonic/gin.(*Engine).ServeHTTP
	github.com/gin-gonic/gin@v1.9.1/gin.go:576
github.com/seal-io/walrus/pkg/apis/runtime.(*Router).ServeHTTP
	github.com/seal-io/walrus/pkg/apis/runtime/router.go:191
net/http.serverHandler.ServeHTTP
	net/http/server.go:2947
net/http.initALPNRequest.ServeHTTP
	net/http/server.go:3556
net/http.(*http2serverConn).runHandler
	net/http/h2_bundle.go:5910
[...]

Issue #1351, opened 2023-11-06
Deploying an existing resource, manually deleting it in target cluster and redeploying it shows deployment failure


Finenyaco:
**Describe the bug**

Deploying an existing resource, manually deleting it in target cluster and redeploying it shows deployment failure

**Steps to reproduce**
1. Create a service that exists in target cluster.
<img width="1100" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/a0c3ff21-7cd9-4b0f-a46b-4f7156c49777">

2. Manually delete resources in the target cluster.
3. Redeploy the service and check the results.

**Result**

1. Deployment failed and no logs.
<img width="1502" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/81d65669-e632-4f89-847f-ec2ef227e531">

2. Checked walrus logs.
```
	* model: validator failed for field "ServiceResourceRelationship.service_resource_id": value is blank


github.com/seal-io/walrus/pkg/apis/runtime.erroring
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_error.go:45
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.recovering
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_recovery.go:39
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.observing
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_observation.go:89
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/gin-gonic/gin.(*Engine).handleHTTPRequest
	github.com/gin-gonic/gin@v1.9.1/gin.go:620
github.com/gin-gonic/gin.(*Engine).ServeHTTP
	github.com/gin-gonic/gin@v1.9.1/gin.go:576
github.com/seal-io/walrus/pkg/apis/runtime.(*Router).ServeHTTP
	github.com/seal-io/walrus/pkg/apis/runtime/router.go:191
net/http.serverHandler.ServeHTTP
	net/http/server.go:2947
net/http.initALPNRequest.ServeHTTP
	net/http/server.go:3556
net/http.(*http2serverConn).runHandler
[...]

PR #1444, opened 2023-11-17
fix: failure to upgrade service when first deployment fails

shanewxy:
**Problem:**

When redeploying previously failed service, new instance dependencies are created before the instances are created, so the id field is blank, causing deployment fail.

**Solution:**

Save the new instances first before update dependencies.

**Related Issue:**
#1435 
#1351


gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

PR #1446, opened 2023-11-17
refactor: support generate span and order in schema

aiwantaozi:
**Problem:**
Need to add order and span for better UI render

**Solution:**
Support  ui extension configure order and colSpan

**Related Issue:**
#1306

### DIFF
--- a/pkg/templates/loader/terraform_loader.go
+++ b/pkg/templates/loader/terraform_loader.go
@@ -205,12 +205,16 @@ func (l *TerraformLoader) applyMissingConfig(generated, customized *openapi3.Sch
 		}
 
 		// Extensions.
-		if len(in.Value.Extensions) != 0 {
-			if len(v.Value.Extensions) == 0 {
-				s.Properties[n].Value.Extensions = make(map[string]any)
-			}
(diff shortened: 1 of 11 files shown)


pkg/templates/loader/terraform_loader.go
```diff
@@ -205,12 +205,24 @@ func (l *TerraformLoader) applyMissingConfig(generated, customized *openapi3.Sch
...
+
+		order := openapi.GetUIOrder(v.Value.Extensions)
+		genOrder := openapi.GetUIOrder(in.Value.Extensions)
+
+		if order == 0 && genOrder != 0 {
+			ext.SetUIOrder(genOrder)
 		}
+
+		colSpan := openapi.GetUIColSpan(v.Value.Extensions)
+		genColSpan := openapi.GetUIOrder(in.Value.Extensions)
```
gitlawr, line 219: Should we define the UIExtension structure so that it's easier and clearer for parsing and maintaining?

aiwantaozi: updated

gitlawr (approved):
LGTM

thxCode (approved):
LGTM

Issue #1442, opened 2023-11-16
Failed to delete a resource that does not match any resource definition rules


Finenyaco:
**Describe the bug**

Failed to delete a resource that does not match any resource definition rules

**Steps to reproduce**
1. Create a resource definition.
2. Create a resource doesn't match any resource type.
3. Delete the resource created in step 2.

**Result**

1. Prompts for succeed but the actual deletion failed.
2. Check walrus logs.

```
2023-11-16T12:15:58.240Z	E	fail to destroy resource: %!w(*errors.errorString=&{resource definition  does not match resource })
github.com/seal-io/walrus/utils/log.DelegatedLogger.Errorf
	github.com/seal-io/walrus/utils@v0.0.0-00010101000000-000000000000/log/delegate.go:168
github.com/seal-io/walrus/utils/log.Errorf
	github.com/seal-io/walrus/utils@v0.0.0-00010101000000-000000000000/log/logger.go:174
github.com/seal-io/walrus/pkg/resource.Destroy
	github.com/seal-io/walrus/pkg/resource/resource.go:164
github.com/seal-io/walrus/pkg/apis/resource.Handler.CollectionDelete.func1
	github.com/seal-io/walrus/pkg/apis/resource/basic.go:304
github.com/seal-io/walrus/pkg/dao/model.(*Client).WithTx
	github.com/seal-io/walrus/pkg/dao/model/client.go:558
github.com/seal-io/walrus/pkg/apis/resource.Handler.CollectionDelete
	github.com/seal-io/walrus/pkg/apis/resource/basic.go:271
reflect.Value.call
	reflect/value.go:584
reflect.Value.Call
	reflect/value.go:368
github.com/seal-io/walrus/pkg/apis/runtime.(*Router).Routes.func1
	github.com/seal-io/walrus/pkg/apis/runtime/router_route.go:150
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/auths.Account.Filter
	github.com/seal-io/walrus/pkg/auths/account.go:56
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.If.func1
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_filter.go:54
github.com/gin-gonic/gin.(*Context).Next
[...]

PR #1447, opened 2023-11-17
refactor: do straight deletion when no components exist

gitlawr:
**Problem:**
Now deletion is done by completion of a revision. It's not feasible for a resouce failing to match a definition. It cannot build a revision whose templateID is required.

**Solution:**
Do straight deletion instead of creating a revision and delete via revision status sync if nothing needs to be destroyed.

**Related Issue:**
https://github.com/seal-io/walrus/issues/1442


alexcodelf (approved):
LGTM

thxCode (approved):
LGTM

PR #1450, opened 2023-11-17
fix: make template version in definition rule mutable

gitlawr:
**Problem:**
fail to update template version of a matching rule.

**Solution:**
update schema and make it mutable

**Related Issue:**
https://github.com/seal-io/walrus/issues/1436


alexcodelf (approved):
LGTM

aiwantaozi (approved):

Issue #1451, opened 2023-11-17
error adding a template when the template repo has no tags


gitlawr:
**Steps to reproduce**
1. Add template from https://github.com/walrus-catalog-sandbox/terraform-kubernetes-redis

**Result**
<img width="742" alt="image" src="https://github.com/seal-io/walrus/assets/5697937/c7546a38-9cc9-4672-b313-d3474d2d4b54">

**Environment**
- Walrus version: dev-3ccf6b0

gitlawr:
The cause is no tags are present in the repo. We can improve the error message to reveal the cause

gitlawr:
Verified in v0.4.0-rc2

PR #1453, opened 2023-11-17
fix: error adding a template when the template repo has no tags

shanewxy:
**Problem:**

When project_id is nil, the sql query fails to parse this field.

**Solution:**

Add the condition ProjectIDIsNil.

**Related Issue:**
#1451


alexcodelf (approved):
LGTM

aiwantaozi (approved):

Issue #1456, opened 2023-11-17
Connectors of non-Kubernetes type were shown in cost perspective select list


aiwantaozi:
**Describe the bug**

Connectors of non-Kubernetes type were shown in perspective select list

**Steps to reproduce**
1. check the cost perspective for cluster
2. it also returns non-Kubernetes connector
3. abc is a version control type connector

<img width="920" alt="image" src="https://github.com/seal-io/walrus/assets/5936271/f5a92e59-1d39-495a-9943-a9d830171bdb">

**Environment**
- Walrus version: 1744a05efacd67e9bb10ea6b032677ab930d4881

**Additional context**

shanewxy:
**Environment**
- Walrus version: v0.4.0-rc2 af772a7ae049670658fdb2d2249cfde51277db8e
- UI version: v0.4.0-rc2

Test Result: pass

<img width="1244" alt="image" src="https://github.com/seal-io/walrus/assets/45061862/d913a46b-fa35-44e4-84e3-d5ba561a637b">

PR #1457, opened 2023-11-17
fix: connectors of non-Kubernetes type were shown in perspective select list

aiwantaozi:
**Problem:**
Connectors of non-Kubernetes type were shown in the perspective select list

**Solution:**
Should not show non-Kubernetes type connector in perspective select list

**Related Issue:**
#1456


gitlawr (approved):
LGTM

shanewxy (approved):

PR #1459, opened 2023-11-17
fix: tuple endpoint return invalid error

aiwantaozi:
**Problem:**
Tuple endpoint returns an invalid error

**Solution:**
Should not return invalid error while 

**Related Issue:**

### DIFF
--- a/pkg/apis/resource/extension.go
+++ b/pkg/apis/resource/extension.go
@@ -304,10 +304,20 @@ func (h Handler) getEndpointsFromOutput(ctx context.Context, id object.ID) ([]Ac
 				Name:      v.Name,
 			})
 		case v.Schema.Type == openapi3.TypeArray:
-			if v.Schema.Items == nil || v.Schema.Items.Value == nil ||
-				v.Schema.Items.Value.Type != openapi3.TypeString {
-				// For list and set: all elements are the same type.
-				return nil, invalidTypeErr
+			if v.Schema.Items != nil && v.Schema.Items.Value != nil {
+				// Array.
+				if v.Schema.Items.Value.Type != "" && v.Schema.Items.Value.Type != openapi3.TypeString {
+					return nil, invalidTypeErr
+				}
+
+				// Tuple.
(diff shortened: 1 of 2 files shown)


pkg/terraform/parser/parser.go
```diff
@@ -225,15 +225,17 @@ func ParseStateOutput(revision *model.ResourceRevision) ([]types.OutputValue, er
 				val = []byte(`"<sensitive>"`)
 			}
 
+			s := translator.SchemaOfType(
```
alexcodelf, line 228: What is the different of this change between before and after?

thxCode (approved):
LGTM

pkg/apis/resource/extension.go
```diff
@@ -304,10 +304,20 @@ func (h Handler) getEndpointsFromOutput(ctx context.Context, id object.ID) ([]Ac
 				Name:      v.Name,
 			})
 		case v.Schema.Type == openapi3.TypeArray:
-			if v.Schema.Items == nil || v.Schema.Items.Value == nil ||
-				v.Schema.Items.Value.Type != openapi3.TypeString {
-				// For list and set: all elements are the same type.
-				return nil, invalidTypeErr
+			if v.Schema.Items != nil && v.Schema.Items.Value != nil {
+				// Array.
+				if v.Schema.Items.Value.Type != "" && v.Schema.Items.Value.Type != openapi3.TypeString {
```
alexcodelf, line 309: When `v.Schema.Items.Value.Type != openapi3.TypeString`, it must not equal to `""`, So it just need to change to 
```
if v.Schema.Items.Value.Type != openapi3.TypeString {
    xxx
}
```

pkg/terraform/parser/parser.go
aiwantaozi: just makes the code read easier

pkg/apis/resource/extension.go
aiwantaozi: when the type isn't set, also allows to pass

PR #1480, opened 2023-11-20
refactor: support generate order of terraform object

aiwantaozi:
**Problem:**
While generating schema from variables.tf, can't support generating the order of object's properties

**Solution:**
Support generating object properties's order

**Related Issue:**
#1306


gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #1474, opened 2023-11-20
Resources will be displayed in service list after rollback


Finenyaco:
**Describe the bug**

Rollback a resource and it will be displayed in the service list too.

**Steps to reproduce**
1. Create a resource called `test`.
2. Rollback resource `test`.
3. Check the service list.

**Result**

1. 'test' appears in both the resource list and the service list.
<img width="2502" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/72ca0904-64c1-4131-b37c-049cdea57ad3">

2. Deleting these two records at the same will return an error.
<img width="2512" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/408e4a25-154b-4a4d-9e70-78d7e39997e4">

**Expected behavior**

1. Resource should not be displayed in service list.

**Environment**
- Walrus version: f94d57de903a925f2bd72d34b5aa1eaf8213f6ad
- UI version: dev-e4baff9

**Additional context**

shanewxy:
**Environment**

Walrus version: v0.4.0-rc2
UI version: v0.4.0-rc2

**Test Result**: Pass
<img width="1439" alt="image" src="https://github.com/seal-io/walrus/assets/45061862/322685d9-89d9-43ae-ad82-80b449d3c3a9">

PR #1489, opened 2023-11-21
fix: rollback resource

gitlawr:
Fix the problem that templateID of resource using definition is set on rollback.

**Related Issue:**
https://github.com/seal-io/walrus/issues/1474


alexcodelf (approved):
LGTM

aiwantaozi (approved):

Issue #1488, opened 2023-11-21
Failed to reference a complex type variable


Finenyaco:
**Describe the bug**

Failed to reference a complex type variable

**Steps to reproduce**
1. Create a service `test1` using `webservice` template.
2. Create another service `test2` using `webservice` template and reference the output `ports` of `test1`.
3. Check the deployment results.

**Result**

1. Deploy failed, the variable type was incorrect.
<img width="799" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/422d8f06-2e43-4521-a86c-aa78d696a7d5">

**Expected behavior**

Complex type variable can be referenced successfully.

**Environment**
- Walrus version: f94d57de903a925f2bd72d34b5aa1eaf8213f6ad
- UI version: dev-ab43587

**Additional context**

PR #1499, opened 2023-11-21
fix(terraform): failed to reference a complex type variable

alexcodelf:
**Problem:**
The reference of resource outputs is set to string when generate terraform configs

Failed to reference a complex type variable #1488

**Solution:**
keep the output variable type when generate the config

**Related Issue:**
#1488

### DIFF
--- a/pkg/terraform/config/config_test.go
+++ b/pkg/terraform/config/config_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/zclconf/go-cty/cty"
 
 	"github.com/seal-io/walrus/pkg/terraform/block"
 )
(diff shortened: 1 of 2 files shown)


shanewxy (dismissed):

pkg/terraform/config/config_test.go
```diff
@@ -203,3 +204,70 @@ func TestToModuleBlock(t *testing.T) {
...
+			ctyType:  cty.DynamicPseudoType,
+			expected: "any",
+		},
+	}
+
+	for _, tt := range cases {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := typeExprTokens(tt.ctyType)
+			if err != nil {
+				t.Errorf("unexpected error: %s", err)
```
gitlawr, line 265: Can you add some tests for the unhappy paths?

gitlawr (approved):

PR #1515, opened 2023-11-22
fix: prevent crash while no variable and add generate groupOrder

aiwantaozi:
**Problem:**
1. The template without variable will crash
2. while calling of schema expose function will reset the schema
3. add group order while generating

**Issue**
#1518
#1517


thxCode (approved):
LGTM

PR #1526, opened 2023-11-23
fix: mirror image

alexcodelf:
**Problem:**
without new line the last image can not be mirrored

**Solution:**
Add condition `[[ -n "$line" ]]` checks if the $line variable is non-empty after the read command fails to read a new line.

**Related Issue:**


aiwantaozi (approved):

gitlawr (approved):
LGTM

aiwantaozi (approved):

Issue #1532, opened 2023-11-23
[cli] Cannot upgrade a resource


gitlawr:
**Steps to reproduce**
1. Create a resource.
2. Try to upgrade it using cli.

**Result**
```
$ w resource upgrade from-cli --attributes '{"image": "nginx"}' --template null
unexpected status code 400: template or resource definition is required
$ w resource upgrade from-cli --attributes '{"image": "nginx"}' --template null --type "nginx"
Error: unknown flag: --type
Usage:
  walrus resource upgrade <resource> [flags]

Flags:
      --attributes json          (default {}...)
      --description string
      --labels stringToString    (default [])
      --template json            (default {}...)

Global Flags:
  -d, --debug           Enable debug log
  -h, --help            Help for this command
  -o, --output string   Output format [table, json, yaml] (default "table")
  -v, --version         Version for CLI

unknown flag: --type
```

**Expected behavior**
Be able to upgrade a resource using CLI.

**Environment**
- Walrus version: dev-9ddbecf

alexcodelf:
Env: walrus-server: faeb6f61f232c5a68e3638c09299e4e19547bbf3
Test Result: PASS

resource can be upgraded through cli.


![Image](https://github.com/seal-io/walrus/assets/85595852/7c36383e-3c0b-477e-a3d1-b6bb5388866b)

PR #1534, opened 2023-11-23
refactor: do not require type on resource upgrade

gitlawr:
**Problem:**
Type is required when upgrading resource with definition. But it's immutable and is not configurable in generated upgrade command.

**Solution:**
Do not require type on resource upgrade

**Related Issue:**
https://github.com/seal-io/walrus/issues/1532


thxCode (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #1492, opened 2023-11-21
Add containerservice template


gitlawr:
A phased enhancement of https://github.com/seal-io/walrus/issues/1322

shanewxy:
<img width="1410" alt="image" src="https://github.com/seal-io/walrus/assets/45061862/6e88a28c-ce8e-44db-aa92-b43a64cc9c31">

thxCode:
@aiwantaozi 
request payload
```
[...]

PR #1536, opened 2023-11-23
fix: donot respect ui schema

thxCode:
**Problem:**

validate the resource variables with the original schema but not the (customized) UI schema.

**Solution:**

prioritize using UI schema verification if found.

**Related Issue:**
partial fix for https://github.com/seal-io/walrus/issues/1492


gitlawr (approved):
LGTM

aiwantaozi (approved):

PR #1541, opened 2023-11-23
refactor: skip set groupOrder while exist

aiwantaozi:
**Problem:**
groupOrder reset.

**Solution:**
Skip set groupOrder while exist


gitlawr (approved):
LGTM

shanewxy (approved):
LGTM

Issue #1485, opened 2023-11-21
Show empty group when all properties are hidden


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Go to `deploy-source-code` template detail page.
2. Check the UI schema preview.

**Result**

1. There is an empty `Basic` group.
<img width="897" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f9ab6055-519a-470e-b8e5-26e5163e1a68">

2. Check the UI schema. The property of basic group is hidden.
```
components:
  schemas:
    variables:
      properties:
        context:
          additionalProperties:
            type: object
          default: {}
          description: >
            Receive contextual information. When Walrus deploys, Walrus will
            inject specific contextual information into this field.


            Examples:

            ```

            context:
              project:
                name: string
                id: string
              environment:
                name: string
                id: string
              resource:
                name: string
                id: string
            ```
          title: Context
          type: object
          x-walrus-ui:
            colSpan: 12
            group: Basic
            hidden: true
            order: 23
```
**Expected behavior**

Do not show the group if all properties are hidden.

**Environment**
- Walrus version: f94d57de903a925f2bd72d34b5aa1eaf8213f6ad
- UI version: dev-ab43587

**Additional context**

gitlawr:
@aiwantaozi Can we remove the basic group in the schema?

gitlawr:
Update schema of ports & envvars properties to make it look better
<img width="1402" alt="image" src="https://github.com/seal-io/walrus/assets/5697937/d6ce4ebf-9f1a-4606-bd68-b5f280335ad9">

aiwantaozi:
![Image](https://github.com/seal-io/walrus/assets/5936271/13ce291a-7f0a-4eae-91bc-9556bab1a49f)
The Order does not work as expected.
UI version: https://github.com/seal-io/walrus-ui/commit/26a7d60d9ee093b2050c2501bb42822034a908d8

aiwantaozi:
Update the ports's colSpan to 6 and update the environment colSpan to 6, still the same image as above, the span and order not work as expected.
UI version: https://github.com/seal-io/walrus-ui/commit/6b316d39797a9bb76af6eec65d28de85e76be1dd

alexcodelf:
Env:  UI dev-8067cc7
Test Result: PASS
Case of former: PASS
1. Order. Rendered as expected.
[...]

PR #1551, opened 2023-11-23
refactor: don't set order and colspan while schema.yaml existed

aiwantaozi:
**Problem:**
Server will set the default order and colSpan

**Solution:**
Should not set default order and colSpan while existed

**Related Issue:**
#1485


alexcodelf (approved):
LGTM

Issue #1579, opened 2023-11-24
[v0.4.0-rc3] failed to use repo subpath importing template


thxCode:
**Describe the bug**

Failed to use a template imported by repo subpath to create a service.

**Steps to reproduce**
1. Prepare an Environment connected with a Kubernetes connector.
2. Import a template from a repo's subpath, like: https://github.com/thxCode/walrus-testing-template//modules/aws-vpc?ref=main
3. Create a service with the template created by step 2.

**Result**

Failed to use the real module under the subpath.

<img width="697" alt="image" src="https://github.com/seal-io/walrus/assets/3394724/42f31e10-1e30-4e71-aa97-a2e2e1dd5432">

<img width="1325" alt="image" src="https://github.com/seal-io/walrus/assets/3394724/63ddc30d-b0d8-4975-934f-66ad84b4b1a2">

<img width="653" alt="image" src="https://github.com/seal-io/walrus/assets/3394724/18d97d1b-153a-49d0-8991-57084054c46f">

**Expected behavior**

Work with the template imported by subpath.

**Environment**
- Walrus version:

**Additional context**
None

shanewxy:
Verified in main.


![Image](https://github.com/seal-io/walrus/assets/45061862/168f212c-c0f4-4f45-97af-bd655fcde7dc)

![Image](https://github.com/seal-io/walrus/assets/45061862/7e8c672d-adfc-4ed1-b5cd-99282258d929)

PR #1589, opened 2023-11-24
fix: template version mission subpath

alexcodelf:
**Problem:**

template version missing subpath.

**Solution:**

add subpath for template version when import a template with subpath

**Related Issue:**

#1579

### DIFF
--- a/pkg/templates/git.go
+++ b/pkg/templates/git.go
@@ -210,6 +210,10 @@ func syncTemplateFromRef(
 		)
 	}
 
+	if repo.SubPath != "" {
+		source += "//" + repo.SubPath
+	}
+
 	// Create or update a template version.
 	return mc.TemplateVersions().Create().
 		Set(&model.TemplateVersion{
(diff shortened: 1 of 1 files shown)


pkg/templates/git.go
```diff
@@ -345,6 +350,7 @@ func GetTemplateVersions(
 	entity *model.Template,
 	newVersions []*version.Version,
 	versionSchema map[*version.Version]types.TemplateVersionSchema,
+	subPath string,
```
gitlawr, line 353: Is subPath included in `entity.Source`?

pkg/templates/git.go
```diff
@@ -33,6 +33,7 @@ func CreateTemplateVersionsFromRepo(
 	entity *model.Template,
 	versions []*version.Version,
 	versionSchema map[*version.Version]types.TemplateVersionSchema,
+	subPath string,
```
gitlawr, line 36: Is subPath included in entity.Source?

gitlawr (approved):
LGTM

thxCode (approved):
LGTM

Issue #1560, opened 2023-11-24
Cannot run workflow with non-configurable variable


shanewxy:
**Describe the bug**

**Steps to reproduce**
1. Create a workflow.
2. Create two variables, one configurable and the other one non-configurable.
3. Run the workflow.

**Result**

Cannot run workflow. Log shows the non-configurable variable is invalid.
<img width="1236" alt="image" src="https://github.com/seal-io/walrus/assets/45061862/a98354fa-34df-4c04-8c19-04198a42846e">

**Expected behavior**

Run the workflow successfully.

**Environment**
- Walrus version: v0.4.0-rc3

**Additional context**

shanewxy:
Verified in main.

PR #1590, opened 2023-11-24
fix: cannot run workflow with non-configurable variable

alexcodelf:
**Problem:**

Bug: Workflow params not set overwritable will make valid failed.
Reason: The input variable map is overwirte to request when validate.

**Solution:**

Aviod overwrite the raw variables when validate run workflow input.

**Related Issue:**

#1560

### DIFF
--- a/pkg/dao/workflow.go
+++ b/pkg/dao/workflow.go
@@ -377,33 +377,35 @@ func parseWorkflowVariables(
 
 // OverwriteWorkflowVariables merges the variables into the config.
 func OverwriteWorkflowVariables(vars map[string]string, config types.WorkflowVariables) (map[string]string, error) {
-	configKeys := sets.NewString()
+	mergedVars := make(map[string]string, len(config)+len(vars))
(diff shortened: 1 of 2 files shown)


pkg/dao/workflow.go
```diff
@@ -394,16 +395,18 @@ func OverwriteWorkflowVariables(vars map[string]string, config types.WorkflowVar
...
 	}
 
 	for i := range config {
 		name, value := config[i].Name, config[i].Value
-		if _, ok := vars[name]; !ok {
-			vars[name] = value
+		if _, ok := mergedVars[name]; !ok {
+			mergedVars[name] = value
 		}
 	}
```
gitlawr, line 407: Can we set these in L387-L392 so that we only loop config once?

alexcodelf: updated

gitlawr (approved):
LGTM

thxCode (approved):
LGTM

Issue #1586, opened 2023-11-24
Trim "terraform-" template prefix


gitlawr:

gitlawr:
Verified in main

PR #1592, opened 2023-11-25
feat: trim "terraform-" template prefix

shanewxy:
**Problem:**

Trim template's "terraform-" prefix.

**Related Issue:**
#1586

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -3,6 +3,7 @@ package catalog
 import (
 	"context"
 	"fmt"
+	"strings"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


pkg/catalog/catalog.go
```diff
@@ -146,7 +147,7 @@ func SyncTemplates(ctx context.Context, mc model.ClientSet, c *model.Catalog) er
 				repo.Driver = c.Type
 
 				t := &model.Template{
-					Name:        repo.Name,
+					Name:        strings.TrimPrefix(repo.Name, "terraform-"),
```
gitlawr, line 150: Let's use a function like `normalizeTemplateName` to make the code self-descriptive.

gitlawr (approved):
LGTM

thxCode (approved):
LGTM

Issue #1580, opened 2023-11-24
list out the mismatched resource type


thxCode:
**Describe the bug**

Under a mismatched resource rules' environment, we still can get the resource type.

**Steps to reproduce**
1. Prepare project `default` with a testing environment `local` and a Kubernetes connector.
2. Prepare project `manager` with a testing environment `user`, a production environment `production`, a Kubernetes connector, and an AWS connector.
3. Create a resource definition, named `mysql`, and type it as `mysql`.
4. Create two rules for the res def created by step 3.
    1. If the environment type is testing and the project name is manager, select Kubernetes MySQL template.
    2. If the environment type is production, select AWS MySQL template.

**Result**

Under environment `local`, can list out the mysql resource type.

<img width="638" alt="image" src="https://github.com/seal-io/walrus/assets/3394724/72f3ad1c-0bd6-4348-868f-4c3a9d8fb7ba">

<img width="2500" alt="image" src="https://github.com/seal-io/walrus/assets/3394724/24cfdf12-95f3-4a55-b2f2-1de6b1972de9">

**Expected behavior**

the environment cannot list out the resource types that do not match the rules.

**Environment**
- Walrus version: v0.4.0-rc3

**Additional context**
None

shanewxy:
Verified in main.


![Image](https://github.com/seal-io/walrus/assets/45061862/dbd82784-6ba9-4199-9cd8-c84253007645)

PR #1594, opened 2023-11-25
fix: add missing selector in matching rule query

gitlawr:
**Problem:**
Always reutrns all resource definitions for an environment regardless of the selectors because selector is not queried before the matching.

**Related Issue:**
https://github.com/seal-io/walrus/issues/1580


thxCode (approved):
LGTM

aiwantaozi (approved):

Issue #1555, opened 2023-11-24
incorrect service final status


thxCode:
**Describe the bug**

Completed Kubernetes Job summarise as NotReady.

**Steps to reproduce**
1. Prepare an Environment with a Kubernetes kind of Connector.
2. Choose walrus-catalog-sandbox/terraform-kubernetes-containertask template. (Just deploy a Kubernetes Job)
3. Simply do something echo to make the Job finished.

**Result**

<img width="1666" alt="image" src="https://github.com/seal-io/walrus/assets/3394724/77d626aa-fdf3-4541-b016-1ea28fa1a2f7">

<img width="1647" alt="image" src="https://github.com/seal-io/walrus/assets/3394724/6a61e997-bf35-4bc5-ad11-d623845d1baf">

<img width="500" alt="image" src="https://github.com/seal-io/walrus/assets/3394724/d05f823e-ba35-495c-bc2a-ff42fb2eb4d1">

**Expected behavior**

Be Ready or InActive.

**Environment**
- Walrus version: v0.4.0-rc3

**Additional context**
https://github.com/walrus-catalog-sandbox/terraform-kubernetes-containertask

gitlawr:
Verified in main
Completed Job:
<img width="1230" alt="image" src="https://github.com/seal-io/walrus/assets/5697937/f3e6efd9-fa5f-49cb-8a3c-4c2fb71c2109">

Failed Job:
<img width="1229" alt="image" src="https://github.com/seal-io/walrus/assets/5697937/5891229d-08ff-4980-b14a-4ea3b33f56b6">

PR #1602, opened 2023-11-25
fix: cannot identify completed kube job components

thxCode:
**Problem:**

Do not identify the completed condition of the Pod created by a Job.

**Solution:**

In the status walker, give two decisions to solve this case.

**Related Issue:**
#1555


gitlawr (approved):
LGTM

shanewxy (approved):
LGTM

Issue #1623, opened 2023-11-27
View api detail in swagger return error


aiwantaozi:
**Describe the bug**

**Steps to reproduce**
1. Go to ${walrus_server}/swagger to view the swagger api
2. View the details of API `/v1/dashboards/_/latest-resource-revisions`
3. Will get error

**Result**

**Expected behavior**
Should not return error

**Environment**
- Walrus version:
v0.4.0

**Additional context**

PR #1624, opened 2023-11-27
fix: still generate schema for object without properties

aiwantaozi:
**Problem:**
View some of the API in openAPI swagger return error could not found schema

**Solution:**
reset the `visited` to prevent get doesn't exit schema

**Related Issue:**
#1623


thxCode (approved):
LGTM

gitlawr (approved):

PR #1631, opened 2023-11-28
refactor(deployer): add revision manager and plan

alexcodelf:
- extract revision create to revision manager from terraform deployer

- generate revision plan configs with revision plan

- terraform deplyer using revision create job

**Problem:**

The deployer and the revision are too tightly coupled.

**Solution:**

Change infomation is in commit message.

**Related Issue:**


alexcodelf:
Close as #2131 address the refactor issue

Issue #1599, opened 2023-11-25
Could not import template with revision hash


aiwantaozi:
**Describe the bug**
Could not import template with revision hash

**Steps to reproduce**
1. Import template with hash https://github.com/walrus-catalog/aws-rds?ref=4abe0942fe95f60c24c4188b98add9616390df0c
2. failed to import
3. Based on the terraform doc, revision is allowed, https://developer.hashicorp.com/terraform/language/modules/sources#selecting-a-revision

**Result**
<img width="700" alt="image" src="https://github.com/seal-io/walrus/assets/5936271/d30c4dd4-feed-41b4-b3ad-02b2215bd03e">

**Expected behavior**
should success import

**Environment**
- Walrus version:
Server f3ba1403e0a9c2545220715867724e8c4eb5cf14
**Additional context**

PR #1635, opened 2023-11-28
fix: could not import template with revision hash

alexcodelf:
**Problem:**

Could not import template with hash ref.

**Solution:**

Parse and reset hash when import template.

**Related Issue:**
#1599


pkg/templates/git.go
```diff
@@ -214,6 +215,10 @@ func syncTemplateFromRef(
 		source += "//" + repo.SubPath
 	}
 
+	if plumbing.IsHash(ref) {
+		ref = ref[:7]
```
thxCode, line 219: make sure it does not happen out-of-range.

alexcodelf: Try to get hash of reference from repo, and trim it as template version.

pkg/templates/git.go
```diff
@@ -214,6 +214,23 @@ func syncTemplateFromRef(
...
+	switch {
+	case reference.Name().IsBranch():
+	case reference.Name().IsTag():
+	case reference.Name().IsRemote():
+	case reference.Name().IsNote():
+	default:
+		ref = reference.Hash().String()
+		if len(ref) > 7 {
+			ref = ref[:7]
+		}
```
gitlawr, line 231: Can you explain the cause of the problem? What happen if we don't trim the hash?

alexcodelf: There are two consideration:

- Our version list query will try to compare version order by convert version tag to int array(compare every digits). If we keep the full hash, it will convert to integer  that out of range of integer. 
- The short hash like github will make it easier to select template version.
image

alexcodelf: Updated. Modify version order sql, only sort semver version.

pkg/apis/template/extension.go
```diff
@@ -63,11 +63,17 @@ func (h Handler) RouteGetVersions(
 		query.Select(fields...)
 	}
 
+	//nolint:lll
+	// From https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string.
+	semverExpression := `^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`
```
gitlawr, line 68: Do you think we can simply filter out sortable versions by the presence of the `.` charactor?

pkg/templates/git.go
```diff
@@ -214,6 +214,20 @@ func syncTemplateFromRef(
...
+	}
+
+	switch {
+	case reference.Name().IsBranch():
+	case reference.Name().IsTag():
+	case reference.Name().IsRemote():
+	case reference.Name().IsNote():
+	default:
+		ref = reference.Hash().String()
+	}
```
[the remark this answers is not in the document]
alexcodelf: removed

pkg/apis/template/extension.go
alexcodelf: Updated to simplied expression. The short one may parse some invalid semver but is acceptable in our case. From semver offical site it can cover all cases if we want strictly filter semver tags.

Issue #1644, opened 2023-11-29
Walrus crashed and restarted because argo workflows failed to install


linyinli:
**Describe the bug**
Walrus crashed and restarted because argo workflows failed to install.

**Steps to reproduce**
Deploy Walrus by High Availability Installation: https://seal-io.github.io/docs/deploy/replication

**Result**
Walrus crashes and restarts:
```
2023-11-28T10:48:17.710Z        E       workflow        failed to deploy argo workflow: error get release walrus-system:walrus-workflow, error get release walrus-system:walrus-workflow, Kubernetes cluster unreachable: Get "https://192.168.0.1:443/version": x509: certificate signed by unknown authority
github.com/seal-io/walrus/pkg/server.applyWorkflowDeployment.func1
        github.com/seal-io/walrus/pkg/server/init_workflow_runtime.go:254
github.com/alitto/pond.(*WorkerPool).executeTask
        github.com/alitto/pond@v1.8.3/pond.go:454
github.com/alitto/pond.worker
        github.com/alitto/pond@v1.8.3/worker.go:32
```

**Expected behavior**
1. High availability installation works fine;
2. A failed argo workflow installation should not affect walrus.

**Environment**
- Walrus version: v0.4.0

Finenyaco:
**Environment**
- Walrus version: main fc24716b7a80c2db17ff0558b63c83f36db031de
- UI version: dev-0b2f2dc

**Test Result:** pass
<img width="1616" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/d0e4d559-63a8-4282-ad10-d7a6bfb3d347">

```
root@walrus-kind-control-plane:/# k get all -n walrus-system
NAME                                                                                        READY   STATUS      RESTARTS      AGE
pod/database-55fd47f7bc-c6hxb                                                               1/1     Running     0             63m
pod/identity-access-manager-978f95c68-4vcns                                                 1/1     Running     0             63m
pod/test-workflow-490181505738342435-step-secret-490181505738342435-3744763495              0/1     Completed   0             109s
pod/test-workflow-490181505738342435-workflow-step-execution-490181505738342435-989507416   0/2     Completed   0             99s
pod/tf-job-apply-490181421567049763-h8qh5                                                   0/1     Completed   0             2m39s
pod/tf-job-apply-490181525954887714-d4wvt                                                   0/1     Completed   0             97s
[...]

PR #1651, opened 2023-11-30
fix: argo workflows failed to install

alexcodelf:
**Problem:**

Walrus crashed and restarted because argo workflows failed to install.

**Solution:**

Add permission for apply walrus in k8s, using `rest.Config` directly instead of converting to`clientcmdapi.Config`

**Related Issue:**

#1644

### DIFF
--- a/pkg/k8s/deploy/helm.go
+++ b/pkg/k8s/deploy/helm.go
@@ -4,13 +4,17 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path"
 	"time"
 
 	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/chart/loader"
-	"helm.sh/helm/v3/pkg/cli"
 	"helm.sh/helm/v3/pkg/release"
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/client-go/discovery"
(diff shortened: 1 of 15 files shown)


pkg/k8s/deploy/helm.go
```diff
@@ -4,13 +4,18 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path"
 	"time"
 
+	"k8s.io/client-go/restmapper"
+
```
gitlawr, line 10: remove this line

pkg/workflow/client.go
```diff
@@ -406,8 +406,8 @@ type ArgoAPIClient struct {
 }
 
 func NewArgoAPIClient(restCfg *rest.Config) *ArgoAPIClient {
-	apiConfig := k8s.ToClientCmdApiConfig(restCfg)
-	clientConfig := clientcmd.NewDefaultClientConfig(apiConfig, nil)
+	apiConfig, _ := k8s.ToClientCmdApiConfig(restCfg)
```
gitlawr, line 409: Why do we ignore the error?

alexcodelf: The reason is we follow the pattern of init API handler not return error. Let expose it when init API server.

gitlawr (dismissed):
LGTM

Issue #1603, opened 2023-11-25
Definition schema showing template specific variables


gitlawr:
**Steps to reproduce**
1. Add a mysql definition
2. Add two rules with k8s-mysql and aws-mysql
3. Save.
4. Check definition schema

**Result**
It contains variables specific to aws
<img width="907" alt="image" src="https://github.com/seal-io/walrus/assets/5697937/c8fe21b9-4536-48f1-a327-0bea7c5f3c7d">

**Environment**
- Walrus version: main

PR #1652, opened 2023-11-30
refactor: generate ui schema for res def

thxCode:
**Problem:**

- keep the keys have the same name, but do not consider the type.
- have not shrunk the range, including description, default, enum, maximum, minimum, and so on.

**Solution:**

shrink the first-level properties of the schemas.

**Related Issue:**
#1603

### DIFF
--- a/pkg/apis/resourcedefinition/helper.go
+++ b/pkg/apis/resourcedefinition/helper.go
@@ -1,0 +1,524 @@
+package resourcedefinition
+
+import (
+	"context"
+	"fmt"
+	"math"
+	"reflect"
+
+	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/tidwall/gjson"
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
+	"k8s.io/apimachinery/pkg/util/sets"
+
+	"github.com/seal-io/walrus/pkg/dao/model"
+	"github.com/seal-io/walrus/pkg/dao/model/templateversion"
+	"github.com/seal-io/walrus/pkg/dao/types"
+	"github.com/seal-io/walrus/pkg/dao/types/object"
+	"github.com/seal-io/walrus/pkg/templates/openapi"
+	"github.com/seal-io/walrus/utils/json"
+	"github.com/seal-io/walrus/utils/pointer"
+)
+
+const (
+	variablesSchemaKey = types.VariableSchemaKey
+	outputsSchemaKey   = types.OutputSchemaKey
+
+	originalExtensionKey = openapi.ExtOriginalKey
+)
+
+// generateSchema generates definition schema with inputs/outputs intersection of matching template versions.
+func generateSchema(ctx context.Context, mc model.ClientSet, df *model.ResourceDefinition) error {
+	// Prepare the match rule schemas.
+	var scss []openapi3.Schemas
+	{
+		// Map default variables.
+		defaultsMapping := make(map[object.ID][]byte, len(df.Edges.MatchingRules))
+		for _, mr := range df.Edges.MatchingRules {
+			defaultsMapping[mr.TemplateID] = json.ShouldMarshal(mr.Attributes)
+		}
+
+		// Map schemas.
+		schemasMapping := make(map[object.ID]openapi3.Schemas, len(df.Edges.MatchingRules))
+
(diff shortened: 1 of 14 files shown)


pkg/apis/resourcedefinition/helper.go
```diff
@@ -0,0 +1,528 @@
...
+				if period+1 < len(d) {
+					lv.Description = string(d[:period+1])
+				}
+			} else {
+				var rt string
+				if t := strings.ToLower(lv.Title); t != strs.Pluralize(t) {
+					rt = "Specify the " + t + "."
+				} else {
+					rt = "Specify " + t + "."
+				}
```
gitlawr, line 291: The logic depending the peroid seems fragile. What if the description uses other terminal punctuation like `?/!` or anthing else? Is there a more universal approach, like getting the longest common substring?

pkg/apis/resourcedefinition/helper.go
```diff
@@ -0,0 +1,528 @@
...
+		if (lv.Default != nil && rv.Default == nil) ||
+			(lv.Default != nil && rv.Default != nil && !reflect.DeepEqual(lv.Default, rv.Default)) {
+			lv.Default = nil
+
+			setSchemaRefMutuallyExclusive(lsr, mef)
+		}
+	}
+
+	// Get the optimal range.
+	// If found exclusion case, it falls back to blank range.
```
gitlawr, line 311: The blank range seems invalid, can we remove the property in this case?

pkg/apis/resourcedefinition/helper.go
```diff
@@ -0,0 +1,528 @@
...
+	"github.com/seal-io/walrus/utils/pointer"
+	"github.com/seal-io/walrus/utils/strs"
+)
+
+const (
+	variablesSchemaKey = "variables"
+	outputsSchemaKey   = "outputs"
+
+	originalExtensionKey          = "x-walrus-original"
+	mutuallyExclusiveExtensionKey = "x-walrus-mutually-exclusive"
```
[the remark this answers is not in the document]
thxCode: do not need to be aware of this, but we need a place to record the state during shrinking.

pkg/apis/resourcedefinition/helper.go
thxCode: updated

pkg/apis/resourcedefinition/helper.go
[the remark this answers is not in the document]
thxCode: use a memory notebook to replace this extension key.

gitlawr (dismissed):
LGTM

pkg/apis/resourcedefinition/helper.go
```diff
@@ -0,0 +1,506 @@
...
+	"github.com/seal-io/walrus/pkg/templates/openapi"
+	"github.com/seal-io/walrus/utils/json"
+	"github.com/seal-io/walrus/utils/pointer"
+)
+
+const (
+	variablesSchemaKey = "variables"
+	outputsSchemaKey   = "outputs"
+
+	originalExtensionKey = "x-walrus-original"
```
aiwantaozi, line 28: Can we use the existed extension key?
https://github.com/seal-io/walrus/blob/6200ad68e4725668a3ab4c9f355cd95fe4b37f51/pkg/templates/openapi/ext.go#L89

pkg/apis/resourcedefinition/helper.go
```diff
@@ -0,0 +1,506 @@
...
+	if lv.Format == "" && rv.Format != "" {
+		lv.Format = rv.Format
+	}
+
+	if !lv.WriteOnly && rv.WriteOnly {
+		lv.WriteOnly = rv.WriteOnly
+	}
+
+	// Find the most complete sentence.
+	// If not found, it cleans up the description.
```
aiwantaozi, line 266: Can these logics be refactored into smaller functions? like alignDescription, alignDefault, alignEnum..

thxCode: although we can do this, I wouldn't say I like a lot of little pieces that function everywhere.

the independent functions are usually used for reusing, but alignDescription/alignDefault/... only works in one place.

aiwantaozi (approved):

Issue #1625, opened 2023-11-27
While the pod is being initialized, Walrus displays InitializeFailed


linyinli:
**Describe the bug**
While the pod is being initialized, walrus displays InitializeFailed.

**Steps to reproduce**
1. Create a container service include init container;
2. Check the status of container.

**Result**
While the pod is being initialized, Walrus displays InitializeFailed.
<img width="1132" alt="image" src="https://github.com/seal-io/walrus/assets/42955482/943ce0c5-a9e1-48f7-9ebb-64e99f345b3f">

**Environment**
- Walrus version: v0.4.0

PR #1654, opened 2023-11-30
fix: display initializeFailed when the pod is initializing

shanewxy:
**Problem:**

When a container service include init container, initialized is false before the init container completed.

**Solution:**

Change the display status to `Initializing` when reason is `ContainersNotInitialized`.

**Related Issue:**
#1625

### DIFF
--- a/pkg/operator/k8s/kubestatus/walker.go
+++ b/pkg/operator/k8s/kubestatus/walker.go
@@ -49,7 +49,24 @@ var podStatusPaths = status.NewWalker(
 		},
 	},
 	func(d status.Decision[core.PodConditionType]) {
-		const reasonPodCompleted = "PodCompleted"
+		const (
+			reasonContainersNotInitialized = "ContainersNotInitialized"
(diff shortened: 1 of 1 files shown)


pkg/operator/k8s/kubestatus/walker.go
```diff
@@ -49,6 +49,20 @@ var podStatusPaths = status.NewWalker(
 		},
 	},
 	func(d status.Decision[core.PodConditionType]) {
+		d.Make(core.PodInitialized,
+			func(st status.ConditionStatus, reason string) (display string, isError, isTransitioning bool) {
+				switch st {
+				case status.ConditionStatusTrue:
+					return "Initialized", false, false
+				case status.ConditionStatusFalse:
+					if reason == "ContainersNotInitialized" {
```
thxCode, line 58: make `ContainerNotInitialized` as constant value, like the `reasonPodCompleted`

thxCode (approved):
LGTM

gitlawr (approved):

Issue #1640, opened 2023-11-29
Fail to import some templates: no supported schema found


gitlawr:
**Describe the bug**
Fail to import some templates: no supported schema found.

**Steps to reproduce**
1. Add a template using source `https://gitlab.com/seal-eyod/walrus-on-aws?ref=v0.0.1`

**Result**
The template is not imported.
<img width="800" alt="image" src="https://github.com/seal-io/walrus/assets/5697937/e1535b28-b34d-4653-b308-86350f03711e">

**Expected behavior**

**Environment**
- Walrus version: v0.4.0

PR #1657, opened 2023-11-30
refactor: load terraform schema doesn't need to match main.tf

aiwantaozi:
**Problem:**
Failed to load template without main.tf

**Solution:**
load terraform schema doesn't need to match main.tf

**Related Issue:**
#1640


alexcodelf (approved):
LGTM

Issue #1597, opened 2023-11-25
be able to refresh template schema of a force-pushed tag


thxCode:
**Is your enhancement related to a problem? Please describe.**

Some scenarios want to avoid version upgradation but be able to refresh its schema. Although we do not update the schema of an existing tag, the new change in the template is still available, which results in inconsistency between what is seen and what is deployed.

**Describe the solution you'd like**

Be able to refresh template schema, no matter from the refreshing of one template or batch refreshing from the catalog.

**Describe alternatives you've considered**

None

**Additional context**

None

Finenyaco:
**Environment**
- Walrus version: main ce9f6f458d0e8facdc2448ccfe78b0ee434b301b
- UI version: dev-e0bc105

**Test Result:** pass

PR #1658, opened 2023-12-01
feat: support force reset template versions

alexcodelf:
**Problem:**

Can not force refresh template or catalog template version with tag updated.

**Solution:**

Make update template version as default behavior. If user hard reset tag, it will update template version.

**Related Issue:**
#1597

### DIFF
--- a/pkg/templates/git.go
+++ b/pkg/templates/git.go
@@ -10,7 +10,6 @@ import (
 	"github.com/go-git/go-git/v5"
 	"github.com/go-git/go-git/v5/plumbing/transport"
 	"github.com/hashicorp/go-version"
-	"k8s.io/apimachinery/pkg/util/sets"
 
 	"github.com/seal-io/walrus/pkg/dao/model"
 	"github.com/seal-io/walrus/pkg/dao/model/template"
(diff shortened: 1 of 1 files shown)


pkg/templates/git.go
```diff
@@ -74,29 +45,17 @@ func CreateTemplateVersionsFromRepo(
...
-		).
+	return mc.TemplateVersions().CreateBulk(templateVersionCreates...).
+		OnConflict(conflictOptions...).
+		UpdateNewValues().
 		Exec(ctx)
-
-	return err
 }
 
 // SyncTemplateFromRef clones a git repository create template and template version with reference.
```
thxCode, line 61: syncTemplateFromRef

pkg/templates/git.go
```diff
@@ -210,23 +192,7 @@ func syncTemplateFromRef(
 		)
 	}
```
thxCode, line 193:
```
if entity.ProjectID == "" {
  return ...
}

return ...
```

pkg/templates/git.go
```diff
@@ -186,6 +145,29 @@ func syncTemplateFromRef(
...
+			Schema:     *schema,
+			ProjectID:  entity.ProjectID,
+		}).
+		OnConflict(conflictOptions...).
+		UpdateNewValues().
+		Exec(ctx)
+}
+
+// templateVersionConflictOptions returns template version conflict options with template project id.
+func templateVersionConflictOptions(entity *model.Template) []sql.ConflictOption {
```
thxCode, line 170: getTemplateVersionUpsertConfilctOptions

thxCode (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1622, opened 2023-11-27
Zombie process appears if walrus can't access Github


orangedeng:
**Describe the bug**
Git zombie process appears if Walrus can't access Github(gfw problem)

**Steps to reproduce**
1. Running the walrus in docker and get into `Operation hub` -> `Catalogs`
2. `Refresh` the `builtin` catalog
3. The catalog shows `Initializing` because Walrus can't access Github

**Result**
1. Exex into Walrus container and run `ps -ef |grep git | wc -l`
2. gets ~50 process for git operations for a long time. And the count will keep increasing over time.

**Expected behavior**
1. return only 1 process count

**Environment**
- Walrus version: v0.4.0

**Additional context**
-  After the catalog status turns into `InitializeFailed`, the git process count won't go down over time.

![截屏2023-11-27 14 36 55](https://github.com/seal-io/walrus/assets/15376837/57ce809d-c6e3-4d3a-a226-c7c149e1527f)

Finenyaco:
**Environment**
- Walrus version: main 8fc1ee8f23f152bbeaf1aec685c03d15ffbccae1
- UI version: dev-199faf2

**Test Result:** Failed

Wait 15 minutes after refreshing the built-in catalogue.

```
root@b187505aae6a:/# ps aux | grep git | wc -l
33
```

Finenyaco:
**Environment**
- Walrus version: main 2ef0df227f0580311686bc6ae2ab19a9548bee94
- UI version: dev-6e415a8

**Test Result:** pass

PR #1668, opened 2023-12-04
fix: zombie processes

alexcodelf:
**Problem:**

Git will create zombie process when sync template or catalog in poor network.

**Solution:**

Using [`tini`](https://github.com/krallin/tini) to manage zombie processes and signal forwarding.

**Related Issue:**

#1622


gitlawr (approved):
LGTM

thxCode (approved):
LGTM

Issue #1454, opened 2023-11-17
Support for automatically updating service/resource status statistics on the environment list page


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

Create a service and go back to the environment list. The service/resource status statistics cannot be updated unless the page is manually refreshed.

<img width="2479" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/64817d4a-95ac-4857-b1fb-ae4bf53864da">

<img width="2479" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/5dab93a2-e621-4dc5-87fe-103d8247dbbc">

**Describe the solution you'd like**

Update the resource/service status statics automatically

**Describe alternatives you've considered**

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 6d6811539cfc22ea3ae13429ab572c63a7908f0f
- UI version: dev-43aa2a1

**Test Result:** pass

Finenyaco:
**Environment**
- Walrus version: main fcec81df221b37754b6be6b5657bbbfab845fabf
- UI version: dev-33d2ecd

**Test Result:** Resource status statistics will be updated after deleting a resource/service, but sometimes the statistics are wrong. https://github.com/seal-io/walrus/issues/1752

PR #1669, opened 2023-12-04
fix: service deletion not update on environment list

shanewxy:
**Problem:**

The database record has already been removed when a resource deletion notified, so it cannot locate the corresponding environment to notify statistics update.

**Solution:**

Add common foreign keys(`environment_id`, `project_id`) to database change trigger notification. Directly use this id retrieved from the old table instead of querying after deletion.

**Related Issue:**
#1454

### DIFF
--- a/pkg/dao/migration/trigger_modelchange.go
+++ b/pkg/dao/migration/trigger_modelchange.go
@@ -37,40 +37,53 @@ CREATE OR REPLACE FUNCTION {{ .Function }}() RETURNS TRIGGER
 AS 
 $$
 DECLARE
-    ids        json;
-    ids_length integer;
-    payload    text;
+    ids                 jsonb;
+    ids_length          int;
+    column_count        int;
+    payload             text;
 BEGIN
-    
-    -- Select ID list from transition table,
-    -- and convert the result to a JSON string array.
+
+    -- Check if the table has project_id and environment_id columns. 
+    -- For entity tables, project_id is always present when environment_id is present.
+    SELECT count(1) FROM information_schema.columns 
+    WHERE table_name = TG_TABLE_NAME 
+        AND column_name IN ('project_id', 'environment_id') 
+    INTO column_count;
+
+    -- Build the ID list.
+    ids := '[]'::jsonb;
+
     CASE TG_OP
-    WHEN 'INSERT' THEN
-        SELECT array_to_json(array(SELECT id::text FROM new_table)) INTO ids;
-    WHEN 'UPDATE' THEN
-        SELECT array_to_json(array(SELECT id::text FROM new_table)) INTO ids;
+    WHEN 'INSERT', 'UPDATE' THEN
+        EXECUTE 'SELECT jsonb_agg(jsonb_build_object(''id'', id::text' ||
+            CASE WHEN column_count >= 1 THEN ', ''project_id'', project_id::text' ELSE '' END ||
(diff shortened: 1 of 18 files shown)


pkg/dao/migration/trigger_modelchange.go
```diff
@@ -37,20 +37,49 @@ CREATE OR REPLACE FUNCTION {{ .Function }}() RETURNS TRIGGER
...
+                environment_ids := null;
+        END;
+        BEGIN
+            EXECUTE 'SELECT array_to_json(array(SELECT project_id::text FROM old_table))' INTO project_ids;
+        EXCEPTION
+            WHEN undefined_column THEN
+                -- Handle the case when the columns don't exist
+                project_ids := null;
+        END;
     ELSE
```
thxCode, line 83: begin...exception... is a way to achieve our goal, but there are three times DSL here, can we fetch the column info from info_schema at first and then select id, project_id, environment_id together?

pkg/dao/migration/trigger_modelchange.go
```diff
@@ -37,20 +37,49 @@ CREATE OR REPLACE FUNCTION {{ .Function }}() RETURNS TRIGGER
...
 $$
 DECLARE
-    ids        json;
-    ids_length integer;
-    payload    text;
+    ids             json;
+    ids_length      integer;
+    environment_ids json;
+    project_ids     json;
+    foreign_keys 	jsonb;
```
thxCode, line 44: align `foreign_keys   jsonb`

pkg/dao/migration/trigger_modelchange.go
thxCode: if the above suggestion is available, can we make the payload looks like this?
```
Timestamp   time.Time    `json:"ts"`
Operation   operation    `json:"op"`
TableSchema string       `json:"tb_s"`
TableName   string       `json:"tb_n"`
Rows      []struct{
  ID          object.ID `json:"id"`
  ProjectID   object.ID `json:"project_id"`
  Environment object.ID `json:"environment_Id"`
}  `json:"ids"`
```

pkg/dao/migration/trigger_modelchange.go
aiwantaozi, line 44: align

aiwantaozi (dismissed):

pkg/dao/migration/trigger_modelchange.go
```diff
@@ -37,40 +37,59 @@ CREATE OR REPLACE FUNCTION {{ .Function }}() RETURNS TRIGGER
...
+        SELECT 1 FROM information_schema.columns WHERE table_name = TG_TABLE_NAME AND column_name = 'project_id'
+    ) THEN
+        has_project_id := true;
+    END IF;
+
+    IF EXISTS (
+        SELECT 1 FROM information_schema.columns WHERE table_name = TG_TABLE_NAME AND column_name = 'environment_id'
+    ) THEN
+        has_environment_id := true;
+    END IF;
```
thxCode, line 60: select count(1) from ... where .... and column_name in ('project_id', 'environment_id')

-- 0: nothing
-- 1: project_id
-- 2: both

shanewxy: updated

thxCode (approved):
LGTM

aiwantaozi (approved):

Issue #1641, opened 2023-11-29
Each time the service is updated, the access URL is displayed as the NodePort endpoint


linyinli:
**Describe the bug**
Each time the service is updated, the access URL is displayed as the NodePort endpoint, and replace it with the contents of endpoints output after refreshing.

**Steps to reproduce**
1. Create a kubernetes-mysql service with NodePort service type;
2. Check the access URL;
3. Refreshing and check the access URL again.

**Result**
<img width="1685" alt="image" src="https://github.com/seal-io/walrus/assets/42955482/c56e166e-c133-40cc-a9bc-a64b99d06923">
<img width="1688" alt="image" src="https://github.com/seal-io/walrus/assets/42955482/3d46bc9f-67c5-4d87-9e50-6509cff16183">

**Expected behavior**
Both should be displayed.

**Environment**
- Walrus version: v0.4.0

Finenyaco:
**Environment**
- Walrus version: main 2ef0df227f0580311686bc6ae2ab19a9548bee94
- UI version: dev-6e415a8

**Test Result:** pass

<img width="1616" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/d81f4450-f92d-4a26-a302-749bd3ef9767">

```
service/port                        NodePort    10.96.65.128   <none>        80:30614/TCP,8080:31008/TCP   12m
```

Issue #1443, opened 2023-11-17
Cannot automatically update access endpoints


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create a service.
2. Go to the service details page and watch the access endpoints list.

**Result**

1. The access endpoints list is not updated automatically.

<img width="2490" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/5443a2d5-671b-4a63-952b-849a682fac1b">

2. Check the chunk request messages, no access endpoints updates.
<img width="1877" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/2a2e19eb-f088-49e3-96d5-c3e468c6d382">

3. After refreshing the page manually.
<img width="2486" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/c3bbf43b-4cde-417b-b05a-ba6378b846a3">

**Expected behavior**

**Environment**
- Walrus version: 4465aa8deb63ff1ceb270b6cbc1c5ae2852d8242
- UI version: dev-4eab774

**Additional context**

alexcodelf:
The Reason is that endpoints of resource is fetched by operator when execute cronjob to discover resource components. But the endpoint stream subscribe `ResourceRevision`, once the revision output does not contains any output and resource conponment update task finished after contab job, the output will not update.

gitlawr:
It worked in v0.3.1

Finenyaco:
**Environment**
- Walrus version: main 2ef0df227f0580311686bc6ae2ab19a9548bee94
- UI version: dev-6e415a8

**Test Result:** pass

PR #1670, opened 2023-12-04
refactor: return both user defined endpoints and endpoints from resource

aiwantaozi:
**Problem:**
When user-defined endpoints exist, the endpoints from resources are ignored.

**Solution:**
Grab both user-defined endpoints and resources

**Related Issue:**
#1641 
#1443

### DIFF
--- a/pkg/apis/resource/extension.go
+++ b/pkg/apis/resource/extension.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/http"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/getkin/kin-openapi/openapi3"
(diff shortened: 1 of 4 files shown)


pkg/apis/resource/extension.go
```diff
@@ -242,25 +243,16 @@ func (h Handler) RouteGetAccessEndpoints(req RouteGetAccessEndpointsRequest) (Ro
 					continue
 				}
 
-				var resp *runtime.ResponseCollection
-
-				switch dm.Type {
-				case modelchange.EventTypeCreate:
-					// While create new resource revision,
-					// the previous endpoints from outputs and resources need to be deleted.
-					resp = runtime.TypedResponse(modelchange.EventTypeDelete.String(), eps)
```
gitlawr, line 251: Do we need to coordinate with UI for this change?

aiwantaozi: Yes, UI is already updated with this change.

pkg/apis/resource/extension.go
```diff
@@ -6,8 +6,10 @@ import (
 	"net/http"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/getkin/kin-openapi/openapi3"
```
alexcodelf, line 11: remove empty line.

pkg/apis/resource/extension.go
```diff
@@ -347,36 +347,97 @@ func (h Handler) getEndpointsFromOutput(ctx context.Context, id object.ID) ([]Ac
...
+	timeout := time.Until(
+		expr.
+			Next(time.Now()).
+			Add(10 * time.Second))
+
+	if timeout > 180*time.Second {
+		timeout = 180 * time.Second
+	}
+
+	waitCtx, cancel := context.WithTimeout(ctx, timeout)
```
alexcodelf, line 420: What is the user experience for user if wait timeout?

aiwantaozi: Need refresh to load the access endpoint. In our framework, updating events are generated through changes in database data. It is challenging to support both high availability (HA) and custom triggering simultaneously. Modifying the revision to trigger data changes when updating the component endpoint in a scheduled task is not appropriate. Create another task to support trigger customized event https://github.com/seal-io/walrus/issues/1682

alexcodelf (dismissed):
LGTM

pkg/apis/resource/extension.go
```diff
@@ -271,17 +267,24 @@ func (h Handler) RouteGetAccessEndpoints(req RouteGetAccessEndpointsRequest) (Ro
...
+	// Endpoints from resources.
+	reps, err := h.getEndpointsFromResources(ctx, id)
+	if err != nil {
+		return nil, err
 	}
 
-	// Endpoints from resources.
-	return h.getEndpointsFromResources(ctx, id)
+	eps := oeps
+	eps = append(eps, reps...)
```
gitlawr, line 282: eps := append(oeps, reps...)

aiwantaozi: Because Go append manipulates underlying array, so the `eps := append(oeps, reps...)` will failed with lint below.
`appendAssign: append result not assigned to the same slice (gocritic)`

gitlawr (approved):
LGTM

Issue #1431, opened 2023-11-15
Support for configuring the maximum number of websocket connections


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

Currently, only 10 terminals can be connected at the same time with one source ip. If all devices have the same egress address, this limitation will be obvious.

<img width="2545" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/2339c900-e5d3-440a-864f-6ddf82c2bb07">

**Describe the solution you'd like**

Support for configuring the maximum number of websocket connections

**Describe alternatives you've considered**

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 2c03e4c222ca1fef240e2fd8ef081e6f4ac6fa3d
- UI version: dev-e0bc105

**Test Result:** 

- Verify that the default value for `--websocket-conn-max-per-ip` is 25 --> pass
- Set `--websocket-conn-max-per-ip` to 10 and verify the maximum number of terminal connections. --> pass

PR #1683, opened 2023-12-05
feat: support configuring max websocket connections

shanewxy:
**Problem:**

The maximum number of websocket connections is fixed. Only 10 terminals can be connected at the same time with one source ip.

**Solution:**

Support for configuring the maximum number of websocket connections by startup command flag.

**Related Issue:**
#1431

### DIFF
--- a/pkg/server/runner.go
+++ b/pkg/server/runner.go
@@ -48,17 +48,18 @@ import (
 type Server struct {
 	Logger clis.Logger
 
-	BindAddress        string
-	BindWithDualStack  bool
-	EnableTls          bool
-	TlsCertFile        string
-	TlsPrivateKeyFile  string
-	TlsCertDir         string
-	TlsAutoCertDomains []string
-	BootstrapPassword  string
-	ConnQPS            int
-	ConnBurst          int
-	GopoolWorkerFactor int
+	BindAddress           string
+	BindWithDualStack     bool
+	EnableTls             bool
+	TlsCertFile           string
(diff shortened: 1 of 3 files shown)


pkg/server/runner.go
```diff
@@ -58,6 +58,7 @@ type Server struct {
 	BootstrapPassword  string
 	ConnQPS            int
 	ConnBurst          int
+	WebsocketMaxConn   int
```
thxCode, line 61: how about `WebsocketConnMaxPerIP`?

pkg/server/runner.go
```diff
@@ -230,6 +232,12 @@ func (r *Server) Flags(cmd *cli.Command) {
 			Destination: &r.ConnBurst,
 			Value:       r.ConnBurst,
 		},
+		&cli.IntFlag{
+			Name:        "websocket-max-conn-per-ip",
```
thxCode, line 236: websocket-conn-max-per-ip?

pkg/server/runner.go
```diff
@@ -93,6 +94,7 @@ func New() *Server {
 		TlsCertDir:             apis.TlsCertDirByK8sSecrets,
 		ConnQPS:                10,
 		ConnBurst:              20,
+		WebsocketMaxConn:       10,
```
thxCode, line 97: according to the issue, we can make this larger to 25.

thxCode (approved):
LGTM

gitlawr (approved):

Issue #1666, opened 2023-12-04
Support variable references when the variable type is an array


linyinli:
**Is your enhancement related to a problem? Please describe.**
Now arrays cannot reference the output of other services/resources when creating services/resources.

**Describe the solution you'd like**
Support variable references when the variable type is an array.

<img width="1293" alt="image" src="https://github.com/seal-io/walrus/assets/42955482/e61fc424-fc97-4a8c-a526-725f9d928a3f">

Finenyaco:
**Environment**
- Walrus version: main fcec81df221b37754b6be6b5657bbbfab845fabf
- UI version: dev-33d2ecd

**Test Result:** pass
<img width="1033" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/0e467e19-1beb-4f2f-b85f-0cf88a444ecf">

PR #1691, opened 2023-12-05
refactor: modify resource reference parser

alexcodelf:
- support variable reference in array
- short resource reference

**Problem:**

- Can not parse variables and resource reference in array.
- Normal shell variable like `${MYSQL_HOST}` in attribute will make deployment failed.
- Reference of `resource` or `service` make it inconvenient to use.

**Solution:**

- Support parse reference in array of attributes.
- Auto convert to [escape-sequences](https://developer.hashicorp.com/terraform/language/expressions/strings#escape-sequences-1) from `${` to `$${`  if attributes contains `${` that is not walrus variables. ~~User may manually use `$${MYSQL_HOST}` directly if they want, parser will keep it if found double dollar of sequence.~~
- Use short reference `svc` and `res`, now we can use `${svc.service1.output}` and `${res.resource1.output}`

@hibig UI need to change.

**Related Issue:**
https://github.com/seal-io/walrus/issues/1492#issuecomment-1825025846
#1666

### DIFF
--- a/pkg/deployer/terraform/deployer.go
+++ b/pkg/deployer/terraform/deployer.go
@@ -62,14 +62,21 @@ const (
 	_variablePrefix = "_walrus_var_"
 
 	// _resourcePrefix the prefix of the service output name.
-	_resourcePrefix = "_walrus_resource_"
+	_resourcePrefix = "_walrus_res_"
 )
 
 var (
 	// _variableReg the regexp to match the variable.
 	_variableReg = regexp.MustCompile(`\${var\.([a-zA-Z0-9_-]+)}`)
 	// _resourceReg the regexp to match the service/resource output.
-	_resourceReg = regexp.MustCompile(`\${(service|resource)\.([^.}]+)\.([^.}]+)}`)
+	_resourceReg = regexp.MustCompile(`\${(svc|res)\.([^.}]+)\.([^.}]+)}`)
+	// _interpolationReg is the regular expression for matching non-reference or non-variable expressions.
+	// Reference: https://developer.hashicorp.com/terraform/language/expressions/strings#escape-sequences-1
+	// To handle escape sequences, ${xxx} is converted to $${xxx}.
+	// If there are more than two consecutive $ symbols, like $${xxx}, they are further converted to $$${xxx}.
+	// During Terraform processing, $${} is ultimately transformed back to ${},
+	// this interpolation is used to ensuring a WYSIWYG user experience.
+	_interpolationReg = regexp.MustCompile(`\$\{((var\.)?([^.}]+)(?:\.([^.}]+))?)[^\}]*\}`)
 )
 
 // Deployer terraform deployer to deploy the resource.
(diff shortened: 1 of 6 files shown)


pkg/deployer/terraform/deployer.go
```diff
@@ -606,7 +611,7 @@ func (d Deployer) loadConfigsBytes(ctx context.Context, opts createK8sSecretsOpt
 		EnvironmentID:    opts.Context.Environment.ID,
 	}
 	// Parse module attributes.
-	variables, dependencyOutputs, err := ParseModuleAttributes(
+	attr, variables, dependencyOutputs, err := ParseModuleAttributes(
```
thxCode, line 614: attrs

thxCode:
> - Auto convert to [escape-sequences](https://developer.hashicorp.com/terraform/language/expressions/strings#escape-sequences-1) from `${` to `$${`  if attributes contains `${` that is not walrus variables. User may manually use `$${MYSQL_HOST}` directly if they want, parser will keep it if found double dollar of sequence.

so `$${MYSQL_HOST}` is the same with `${MYSQL_HOST}`?

walrus input | terraform file | shell script result
---|---|---
`${MYSQL_HOST}` | `$${MYSQL_HOST}` | `${MYSQL_HOST}` // what i set, what i get
`$${MYSQL_HOST}` | `$${MYSQL_HOST}` | `${MYSQL_HOST}` // can we avoid this case?

alexcodelf:
> > * Auto convert to [escape-sequences](https://developer.hashicorp.com/terraform/language/expressions/strings#escape-sequences-1) from `${` to `$${`  if attributes contains `${` that is not walrus variables. User may manually use `$${MYSQL_HOST}` directly if they want, parser will keep it if found double dollar of sequence.
> 
> so `$${MYSQL_HOST}` is the same with `${MYSQL_HOST}`?
> 
> walrus input	terraform file	shell script result
> `${MYSQL_HOST}`	`$${MYSQL_HOST}`	`${MYSQL_HOST}` // what i set, what i get
> `$${MYSQL_HOST}`	`$${MYSQL_HOST}`	`${MYSQL_HOST}` // can we avoid this case?
> @alexcodelf

Let's parse `$${MYSQL_HOST}` to `$$${MYSQL_HOST}`, terraform will convert `$${}` to`${}`. It will follow WYSIWYG for user experience.

pkg/deployer/terraform/deployer.go
```diff
@@ -62,14 +62,19 @@ const (
...
 var (
 	// _variableReg the regexp to match the variable.
 	_variableReg = regexp.MustCompile(`\${var\.([a-zA-Z0-9_-]+)}`)
 	// _resourceReg the regexp to match the service/resource output.
-	_resourceReg = regexp.MustCompile(`\${(service|resource)\.([^.}]+)\.([^.}]+)}`)
+	_resourceReg = regexp.MustCompile(`\${(svc|res)\.([^.}]+)\.([^.}]+)}`)
+	// _interpolationReg the regexp to match the non-reference or non-variable.
+	// Due to https://developer.hashicorp.com/terraform/language/expressions/strings#escape-sequences-1,
+	// ${xxx} will convert to $${xxx} to avoid escape sequences.
+	// If attributes contains $${xxx}, it will not be converted and still keep $${xxx}.
```
thxCode, line 76: why we still keep $${xxx}?

alexcodelf: updated

thxCode (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1693, opened 2023-12-06
Panic as runtime error when adding catalog


pengjiang80:
Failed to add external catalog such as AWS or GCP terraform modules, error message is "panic as runtime error: invalid memory address or nil pointer dereference." 

Version: v0.4.1-rc1
1. Go to "operation hub" -> catalogs
2. Click add catalog and input "https://github.com/terraform-aws-modules"
3. The catalog will be added and failed with "InitializeFailed" status with part of the templates added.

Finenyaco:
**Environment**
- Walrus version: main da3e39d9242713e4b0785ff8578f1bd0441cc314
- UI version: dev-91b63b5

**Test Result:** No panic occurred. But now there is a high probability that synchronizing the catalog will fail.

The success rates of catalog synchronization for different versions are compared below:

<img width="632" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/31433b77-fdba-4f18-ad26-9fd719e16d95">

Finenyaco:
**Environment**
- Walrus version: main 4390f83cc107fffa74428dc93a36de2d8799e317
- UI version: dev-7cc976f

**Test Result:**
Refresh the catalog `https://github.com/terraform-aws-modules
` 20 times with one concurrency, and there is no occurrence of 'error running /usr/bin/git' error.

But with 10 and 20 concurrent instances, this issue is prone to occur easily. https://github.com/seal-io/walrus/issues/1711

Finenyaco:
**Environment**
- Walrus version:  v0.4.1-rc12
- UI version: v0.4.1-rc2

**Test Result:**

Can be reproduced in v0.4.1-rc2
<img width="2485" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/4c215d40-a555-45a8-9851-ae50bed0df10">

```
2023-12-12T08:22:25.702Z	E	gopool	panic observing: panic as runtime error: invalid memory address or nil pointer dereference
github.com/seal-io/walrus/utils/gopool.(*waitGroup).Go.func1.1
	github.com/seal-io/walrus/utils@v0.0.0-00010101000000-000000000000/gopool/waitgroup.go:57
runtime.gopanic
	runtime/panic.go:884
runtime.panicmem
	runtime/panic.go:260
runtime.sigpanic
	runtime/signal_unix.go:839
github.com/seal-io/walrus/pkg/templates.isConstraintSatisfied
[...]

PR #1694, opened 2023-12-06
fix: import catalog failed

alexcodelf:
**Problem:**

Sometimes import catalog failed. Catalog contains invalid templates may not have schema.

**Solution:**

Template schema will be set to variables and outputs not exist.

**Related Issue:**
#1693


thxCode (approved):
LGTM

Issue #1692, opened 2023-12-06
Displays the Evicting status while the Pod is being created


linyinli:
**Describe the bug**
Displays the Evicting status while the Pod is being created.

**Steps to reproduce**
1. Create a service using the kubernetes-mysql template;
2. Check the status of pod.

**Result**
![image](https://github.com/seal-io/walrus/assets/42955482/fc3319c5-2461-4264-ab72-2058c6ccc189)
<img width="723" alt="image" src="https://github.com/seal-io/walrus/assets/42955482/9e4af453-b836-4db4-9f6a-2444e509d404">
<img width="1506" alt="image" src="https://github.com/seal-io/walrus/assets/42955482/984ab450-1298-44fc-80f5-30fa93bd7929">

**Environment**
- Walrus version: v0.4.1-rc1
- Kubernetes version: v1.27.6+k3s1

linyinli:
<img width="865" alt="image" src="https://github.com/seal-io/walrus/assets/42955482/9f111e0a-3d48-442b-a6bb-a2e76a72cb22">

Finenyaco:
**Environment**
- Walrus version: dev-da3e39d
- UI version: dev-91b63b5

**Test Result:** pass

<img width="1590" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/51b73e65-f4bf-4801-88d1-649ca9e054a6">

PR #1695, opened 2023-12-06
fix: displays Evicting status while no summary

shanewxy:
**Problem:**

When the pod has no condition yet, the default summary status uses the last step in the path, which is `DisruptionTarget` in this case.

**Solution:**

Move `DisruptionTarget` to be in front of `Ready`.

**Related Issue:**
#1692


gitlawr (approved):
LGTM

thxCode (approved):
LGTM

PR #1697, opened 2023-12-06
fix: sync catalog git clone failed

alexcodelf:
**Problem:**

Sometines catalog sync template will failed, the getter's context be replaced everytime will cause exit for cloning.

**Solution:**

New getter for each getter.

**Related Issue:**


gitlawr (approved):
LGTM

thxCode (approved):
LGTM

PR #1704, opened 2023-12-08
fix: child process unexpected exit

alexcodelf:
**Problem:**

Our signal handler will clean exited processes before subprocess exited normally.

**Solution:**

- Still introduce tini and hack copy cgroup to make sure embedded k3s start up normally.
- Remove reap handler, using tini to handler zombie process. Let's post it if we really need reap child process.

**Related Issue:**
#1693

### DIFF
--- a/cmd/server/server.go
+++ b/cmd/server/server.go
@@ -1,11 +1,7 @@
 package main
 
 import (
-	"context"
-	"errors"
 	"os"
-	"os/signal"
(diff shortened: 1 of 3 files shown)


thxCode:
https://github.com/seal-io/walrus/pull/1706

alexcodelf:
close as 1706 merged

PR #1690, opened 2023-12-05
refactor: reap child processes if run as pid 1

thxCode:
**Problem:**

further fix after PR https://github.com/seal-io/walrus/pull/1679 to avoid ISSUE #1678 and address ISSUE #1622.

**Solution:**

reap child processes if run as PID 1.

**Related Issue:**
https://github.com/seal-io/walrus/issues/1622

PR #1706, opened 2023-12-10
fix: syscall wait4 result be stolen

thxCode:
**Problem:**

PR https://github.com/seal-io/walrus/pull/1690 intends to use [syscall wait4](https://linux.die.net/man/2/wait4) to reap zombie process, but accidentally stole the result from the normal exec command.

the root cause is that, on linux, the command waiting implementation of exec uses [syscall waitid with WNOWAIT](https://github.com/golang/go/blob/619b8fd7d2c94af12933f409e962b99aa9263555/src/os/exec_unix.go#L21-L25) to confirm whether the process is ready, and then to call [syscall wait4](https://github.com/golang/go/blob/619b8fd7d2c94af12933f409e962b99aa9263555/src/os/exec_unix.go#L42-L47) to fetch the process complete state.

the time gap between ready checking and state fetching can make the process state stoles by reaper, including [tini](https://github.com/krallin/tini/blob/0b44d3665869e46ccbac7414241b8256d6234dc4/src/tini.c#L546).

**Solution:**

- Move the reaper from `main.go` to the util signals package.
- Use waitid to get an exited process but leave its state for the next wait.
- Sleep a while to yield reap's processor.
- Confirm the stat from the system proc file.

> This PR addresses the linux platform only, for freebsd/dragonfly we need further implementation with [syscall wait6](https://man.freebsd.org/cgi/man.cgi?query=wait6).

**Related Issue:**
#1693


staging/utils/signals/handler.go
```diff
@@ -0,0 +1,140 @@
...
+					wpid, err = syscall.Wait4(pid, nil, syscall.WNOHANG, nil)
+				}
+
+				if wpid <= 0 {
+					continue // Cannot recycle the candidate, go ahead to confirm next.
+				}
+
+				logger.Infof("reaped child process %d", wpid)
+			}
+		}
```
alexcodelf, line 125: can we extra to a `reapZombie` or other name func to make it clear?

staging/utils/signals/handler.go
```diff
@@ -0,0 +1,140 @@
...
+
+			for {
+				runtime.Gosched()
+
+				// Use waitid to get an exited child process,
+				// not like waitpid, waitid can leave the child process in waitable state.
+				var si Siginfo
+
+				err := syscallWaitid(P_ALL, 0, &si, syscall.WEXITED|syscall.WNOWAIT, nil)
+				for errors.Is(err, syscall.EINTR) { // Try again if got interrupted.
```
alexcodelf, line 85: What will happen if error is not syscall.EINTR

staging/utils/signals/handler.go
```diff
@@ -0,0 +1,140 @@
...
+						continue // Go ahead to confirm next child process.
+					}
+
+					// Wait again if the process is still a zombie process,
+					// so that the process can be recycled by its parent.
+					time.Sleep(zombieTimeWait)
+				}
+
+				// Use wait4(waitpid) to reap the abnormal exited child process.
+				wpid, err := syscall.Wait4(pid, nil, syscall.WNOHANG, nil)
```
alexcodelf, line 114: Suppose there have multi zombie process. As we sleep finished and want to reap exited child process. At the same time, cmd.Command finished and wait to call `syscall.Wait4(pid, nil, syscall.WNOHANG, nil)`. Will this func will clean normal child process as before?

staging/utils/signals/handler.go
```diff
@@ -0,0 +1,140 @@
...
+	"time"
+
+	"github.com/seal-io/walrus/utils/log"
+)
+
+var registered = make(chan struct{})
+
+// Handler registers for signals and returns a context.
+func Handler() context.Context {
+	close(registered) // Panics when called twice.
```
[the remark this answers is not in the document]
thxCode: sync.Once usually workes the case that reconfigure but not be applied , but in here, we should never let caller regain a context from singals handler.

staging/utils/signals/handler.go
thxCode: yes

staging/utils/signals/handler.go
thxCode: we already gave a specific pid for this func. the specific pid comes from waitid. technically, we use time waiting and proc state validation to yield the routine to avoid reap that kind exited command.

staging/utils/signals/handler.go
thxCode: do not make our funcs to be piecemeal.

gitlawr (approved):
LGTM

Issue #1711, opened 2023-12-11
Simultaneously refreshing the catalog may occasionally result in errors


Finenyaco:
**Describe the bug**

Simultaneously refreshing the catalog may occasionally result in errors

**Steps to reproduce**
1. Add a template with source `https://github.com/terraform-aws-modules`.
2. Refresh the catalog 10/20 times continuously.
3. Check walrus logs.

**Result**

1. There are some errors, for examples,
```
2023-12-11T07:33:51.061Z	E	template	failed to get https://github.com/terraform-aws-modules/terraform-aws-lambda: error downloading 'https://github.com/terraform-aws-modules/terraform-aws-lambda': error running /usr/bin/git:
github.com/seal-io/walrus/pkg/vcs.CloneGitRepo
	github.com/seal-io/walrus/pkg/vcs/repo.go:203
github.com/seal-io/walrus/pkg/templates.SyncTemplateFromGitRepo
	github.com/seal-io/walrus/pkg/templates/git.go:269
github.com/seal-io/walrus/pkg/catalog.SyncTemplates.func1
	github.com/seal-io/walrus/pkg/catalog/catalog.go:165
github.com/seal-io/walrus/utils/gopool.(*waitGroup).Go.func1
	github.com/seal-io/walrus/utils@v0.0.0-00010101000000-000000000000/gopool/waitgroup.go:61
github.com/seal-io/walrus/utils/gopool.(*waitGroup).Go.func2
	github.com/seal-io/walrus/utils@v0.0.0-00010101000000-000000000000/gopool/waitgroup.go:68
github.com/alitto/pond.(*WorkerPool).executeTask
	github.com/alitto/pond@v1.8.3/pond.go:454
github.com/alitto/pond.worker
	github.com/alitto/pond@v1.8.3/worker.go:32
```

**Expected behavior**

**Environment**
- Walrus version: main 4390f83cc107fffa74428dc93a36de2d8799e317
- UI version: dev-7cc976f

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main ed4438198eb885792f5de4162d934ed509e361eb
- UI version: v0.4.1-rc2

Refreshed the catalog `https://github.com/terraform-aws-modules` 20/30 times continuously, this issue was not reproduced and no zombie processes.

**Test Result:** pass

PR #1712, opened 2023-12-11
fix: reap child process before cmd wait finished

alexcodelf:
- enlarge wait time to wait child process to call `syscall.Wait4` normally

**Problem:**

As #1706 reap zombie process, we sometimes still encounter `no child process` error for child process.

**Solution:**

Enlarge wait time to wait child process to call [syscall wait4](https://github.com/golang/go/blob/619b8fd7d2c94af12933f409e962b99aa9263555/src/os/exec_unix.go#L42-L47) and  reap zombie after that.

**Related Issue:**
#1711

### DIFF
--- a/staging/utils/signals/handler.go
+++ b/staging/utils/signals/handler.go
@@ -72,7 +72,7 @@ func Handler() context.Context {
 	// Reap child processes.
 	go func() {
 		for range reapChan {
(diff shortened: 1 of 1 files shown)


thxCode (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1676, opened 2023-12-04
Workflow reports duplicated resource when set the retry times


linyinli:
**Describe the bug**
Workflow reports duplicated resource when set the retry times.

**Steps to reproduce**
1. Create a workflow and some deployment tasks, set retry times;
2. When a deployment fails, it reports duplicated resource.

**Result**
<img width="992" alt="image" src="https://github.com/seal-io/walrus/assets/42955482/7d8b35e3-336f-405a-97e4-1bd1e076baaf">
<img width="898" alt="image" src="https://github.com/seal-io/walrus/assets/42955482/59999f8a-71ec-44b6-bd6c-be7b23a69d08">

**Environment**
- Walrus version: v0.4.0

Issue #1520, opened 2023-11-22
Workflow variables can be created with the same name


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create a workflow.
2. Add two variables with the same.

**Result**

<img width="521" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/9738e606-91fa-4e54-89bf-ecb227870989">

**Expected behavior**

**Environment**
- Walrus version: d96e8a0e7bf4e2a7c68285cb47b204f4aa0c714d
- UI version: dev-9cbb6ba

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main ce9f6f458d0e8facdc2448ccfe78b0ee434b301b
- UI version: dev-e0bc105

**Test Result:**

- Please add validation on the frontend @hibig 
<img width="504" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/ad7fe37b-c574-40bf-a88d-b2c845865229">

- Check the backend.
<img width="524" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/e7d1a92b-b33a-4872-aa09-04610bdaa471">

PR #1714, opened 2023-12-11
fix: workflow issues

alexcodelf:
**Problem:**

- workflow can define duplicate variables
- service create duplicated when retry times over 1. 
    The reason is we have already defined step type to `create`, when create a new servcie failed in step, it actually create a new service, but retry in step will not change jobType dynamically. In this case, step will use `create` API to interact with server will return error 409.

**Solution:**

- add variable validation for workflow create or update
- get service step type with in step, if servcie is created then try to upgrade service in servcie step.

**Related Issue:**
#1676 #1520


thxCode (approved):
LGTM

gitlawr (approved):
LGTM

PR #1721, opened 2023-12-12
fix: check template constraint nil error

alexcodelf:
**Problem:**

Panic occur when import invalid template. The reason that `main(dev)` version does not have this issue is we ignore walrus version contraint check in `main(dev)` version.

**Solution:**

Check variable before get schema ext walrus version

**Related Issue:**
#1693


aiwantaozi (approved):

thxCode (approved):
LGTM

Issue #1719, opened 2023-12-12
Cannot create a resource using the resource type that contains a resource label selector


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create a resource definition with resource labels selector.

<img width="1190" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/aa4cb886-2b8d-4ff1-a9a1-578a85a4feed">

2. Create a resource using the resource definition added above.

**Result**

1. The newly created resource definition is not displayed.

<img width="881" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/08c0dbfa-6237-4d9c-83df-913e1e115b68">

**Expected behavior**

1. Display the type of the newly created resource definition when creating a resource.

**Environment**
- Walrus version: v0.4.1-rc2

**Additional context**

It may be introduced by changes related to issue https://github.com/seal-io/walrus/issues/1580 , but resource labels selector cannot be listed out in advance.

Finenyaco:
**Environment**
- Walrus version: main 0ce7516077eac143abadf67373426215f867d593
- UI version: dev-90d701d

**Test Result:** pass

PR #1728, opened 2023-12-13
fix: available resource types for envs

gitlawr:
**Problem:**
When finding available resource types for an environment, it compares with empty resource label so it never match if there's resource label selector.

**Solution:**
Ignore resource label selector in environment level matching.

**Related Issue:**
https://github.com/seal-io/walrus/issues/1719


alexcodelf (approved):
LGTM

aiwantaozi (approved):

Issue #1523, opened 2023-11-22
Optimize workflow states after stopping it


Finenyaco:
**Environment**
- Walrus version: d96e8a0e7bf4e2a7c68285cb47b204f4aa0c714d
- UI version: dev-9cbb6ba

After stopping the workflow, the workflow status will change to "Failed". I think it would be more accurate to change the status to "Aborted" or "Cancelled".

<img width="1038" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/fb0ad27d-9918-4371-a77a-d98025818050">

Finenyaco:
**Environment**
- Walrus version: main 645047fbb578508e9b8ab46353195be7de52f22e
- UI version: dev-b511a81

Track in https://github.com/seal-io/walrus/issues/1841

PR #1733, opened 2023-12-13
fix: workflow status 

alexcodelf:
**Problem:**

- manual delete workflow from k8s will make workflow status stuck.
- after stopping the workflow, the workflow status will change to "Failed"

**Solution:**

- When terminate or resume workflow, check workflow exist in cluster, if not mark workflow status to failed.
- add status `Canceled`

**Related Issue:**
#1523

@hibig UI need adjust status of workflow.

### DIFF
--- a/pkg/dao/types/status/status_workflow_execution.go
+++ b/pkg/dao/types/status/status_workflow_execution.go
@@ -1,8 +1,9 @@
 package status
 
 const (
-	WorkflowExecutionStatusPending ConditionType = "Pending"
-	WorkflowExecutionStatusRunning ConditionType = "Running"
+	WorkflowExecutionStatusPending  ConditionType = "Pending"
+	WorkflowExecutionStatusRunning  ConditionType = "Running"
+	WorkflowExecutionStatusCanceled ConditionType = "Canceled"
 )
 
 // workflowExecutionStatusPaths makes the following decision.
@@ -14,11 +15,15 @@ const (
 //	| Running          | Unknown                 | Running               | Transitioning         |
 //	| Running          | False                   | Failed                | Error                 |
 //	| Running          | True                    | Completed             | Completed             |
+//	| Canceled         | Unknown                 | Canceling             | Transitioning         |
+//	| Canceled         | False                   | CancelFailed          | Error                 |
+//	| Canceled         | True                    | Canceled              | Canceled              |
 var workflowExecutionStatusPaths = NewWalker(
 	[][]ConditionType{
 		{
 			WorkflowExecutionStatusPending,
 			WorkflowExecutionStatusRunning,
+			WorkflowExecutionStatusCanceled,
 		},
 	},
 )
@@ -28,8 +33,9 @@ func WalkWorkflowExecution(st *Status) *Summary {
 }
 
 const (
-	WorkflowStageExecutionStatusPending ConditionType = "Pending"
-	WorkflowStageExecutionStatusRunning ConditionType = "Running"
+	WorkflowStageExecutionStatusPending  ConditionType = "Pending"
+	WorkflowStageExecutionStatusRunning  ConditionType = "Running"
+	WorkflowStageExecutionStatusCanceled ConditionType = "Canceled"
 )
 
 // workflowStageExecutionStatusPaths makes the following decision.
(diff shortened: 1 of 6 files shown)


pkg/dao/types/status/status_workflow_execution.go
```diff
@@ -14,11 +15,15 @@ const (
...
 //	| Running          | True                    | Completed             | Completed             |
+//	| Stopped          | Unknown                 | Stopping              | Transitioning         |
+//	| Stopped          | False                   | Failed                | Error                 |
+//	| Stopped          | True                    | Stopped               | Completed             |
 var workflowExecutionStatusPaths = NewWalker(
 	[][]ConditionType{
 		{
 			WorkflowExecutionStatusPending,
 			WorkflowExecutionStatusRunning,
+			WorkflowExecutionStatusStopped,
```
gitlawr, line 26: Is it inactive? ref: https://github.com/seal-io/walrus/pull/1737

pkg/workflow/client.go
```diff
@@ -304,9 +313,32 @@ func (s *ArgoWorkflowClient) Terminate(ctx context.Context, opts TerminateOption
 		return err
 	}
 
+	if err == nil {
+		return nil
+	}
+
+	if err := s.setNotFoundStatus(ctx, opts.WorkflowExecution); err != nil {
```
gitlawr, line 320: Why do we handle not-found differently comparing to the Terminate() error?

alexcodelf: When a workflow is not found in Kubernetes, the status of the workflow remains unchanged even when a user attempts to stop it, resulting in a perpetual 'running' status. We require a method to address this situation.

The reason we don't need to handle a `not found` error in `resubmit`, `submit` or `delete` is that the absence of a workflow in Kubernetes does not impact the outcome.

pkg/dao/types/status/status_workflow_execution.go
alexcodelf: Is it inappropriate to use the `inactive` status to indicate an stopped workflow?

pkg/workflow/client.go
gitlawr: What about we update WorkflowExecutionStatusRunning when handling Terminate() error?

pkg/dao/types/status/status_workflow_execution.go
gitlawr: How do we show the status in the UI?

pkg/workflow/status.go
```diff
@@ -123,8 +124,14 @@ func (m *StatusSyncer) SyncWorkflowExecutionStatus(ctx context.Context, wf *wfv1
 		m.Logger.Debugf("workflow %s is succeeded", we.ID.String())
 
 	case wfv1.WorkflowFailed, wfv1.WorkflowError:
-		status.WorkflowExecutionStatusPending.True(we, "")
-		status.WorkflowExecutionStatusRunning.False(we, wf.Status.Message)
+		switch {
+		case status.WorkflowExecutionStatusStopped.IsUnknown(we):
+			status.WorkflowExecutionStatusStopped.True(we, wf.Status.Message)
```
gitlawr, line 129: What does the message look like in this case

pkg/dao/types/status/status_workflow_execution.go
gitlawr: How do we show the status in the UI?

pkg/workflow/status.go
alexcodelf: The message here will be `Stopped with strategy 'Terminate'`

pkg/workflow/client.go
alexcodelf: Updated.

pkg/workflow/status.go
alexcodelf: Let's using gray color to represent `stopped` status.

pkg/dao/types/status/status_workflow_execution.go
alexcodelf: Let's using `Canceled` for stop a workflow.

gitlawr (approved):
LGTM

pkg/dao/types/status/status_workflow_execution.go
```diff
@@ -14,11 +15,15 @@ const (
 //	| Running          | Unknown                 | Running               | Transitioning         |
 //	| Running          | False                   | Failed                | Error                 |
 //	| Running          | True                    | Completed             | Completed             |
+//	| Canceled         | Unknown                 | Canceling             | Transitioning         |
+//	| Canceled         | False                   | Failed                | Error                 |
```
shanewxy, line 19: Should it be consistent with the summary status in walker, i.e. CancelFailed?

shanewxy (approved):
LGTM

gitlawr (approved):

Issue #1636, opened 2023-11-28
Fail to delete  services containing dependencies in undeployed or stopped state


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create an environment and add a two services `app-1` and `app-2`. `app-1` references a project variable and `app-2` is dependent on `app-1`.

<img width="2491" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/632f6b55-a879-4c63-9640-a87ed8a28753">

<img width="2480" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/1bf58fcc-fe03-420a-ba6e-1b1b9c1317e3">

2. Clone the environment and save draft. Delete these two services and check the result.

**Result**

In step 2, services cannot be deleted in `Undeployed` state.
<img width="2553" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f006a9ad-9b7d-43a6-a73a-4c355951ab84">

**Expected behavior**

Successfully delete services containing dependencies in undeployed or stopped state.

**Environment**
- Walrus version: v0.4.0

**Additional context**

Sometimes cannot delete services in stopped status.
<img width="2488" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/1d16b60f-d1dd-4793-a4d0-8846fe547bf4">

Finenyaco:
**Environment**
- Walrus version: main 8048eb2af52e636e42e656ee507d875cb83999d9
- UI version: dev-33d2ecd

**Test Result:** pass

PR #1734, opened 2023-12-13
fix: failed delete resources with dependencies

alexcodelf:
**Problem:**

Failed to batch delete resource with dependecies with no components.  Walrus will delete resource directly if it has no components, dependency check will be skipped.

**Solution:**

Reverse topology sort before delete resources, when resources does not have components. 

**Related Issue:**
#1636

### DIFF
--- a/pkg/resource/topology.go
+++ b/pkg/resource/topology.go
@@ -238,3 +238,19 @@ func TopologicalSortResources(resources model.Resources) (model.Resources, error
 
 	return sortedResources, nil
 }
+
+// ReverseTopologicalSortResources sorts the resource by dependencies in reverse order.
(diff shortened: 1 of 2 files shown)


gitlawr (dismissed):
LGTM

pkg/resource/topology.go
```diff
@@ -238,3 +238,19 @@ func TopologicalSortResources(resources model.Resources) (model.Resources, error
 
 	return sortedResources, nil
 }
+
+// TopologicalReverseSortResources sorts the resource by dependencies in reverse order.
+func TopologicalReverseSortResources(resources model.Resources) (model.Resources, error) {
```
aiwantaozi, line 243: ReverseTopologicalSortResources is better?

aiwantaozi (approved):

Issue #1500, opened 2023-11-21
Classify inactive status in the API


gitlawr:
In resource API, we can tell if the status is in error or transitioning, but we cannot differentiate inactive status like stopped from active status like ready.

Finenyaco:
**Environment**
- Walrus version: main 8fc1ee8f23f152bbeaf1aec685c03d15ffbccae1
- UI version: dev-199faf2

**Test Result:** pass

Inactive states include `Stopped` and `Undeployed`.

gitlawr:
This issue is for clarifying inactive in the status of resource API like transitioning and error. Not for the statistics in environment list.

gitlawr:
For UI: use a proper color for inactive resource/component status.
![image](https://github.com/seal-io/walrus/assets/5697937/0c8109cd-b8e7-4464-8fe1-1c1c52f9116e)

Finenyaco:
**Environment**
- Walrus version: main 278965251ef397a0f67ffab3c9348f4243d92c56
- UI version: dev-bb09fa3

**Test Result:** Failed

<img width="2450" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/7a34f6ea-141d-4f06-8efa-650dea141037">

Finenyaco:
**Environment**
- Walrus version: main 38bbd2af03b5e73b87a803f864ab2045c79b9fe0

The inactive flag is missing when the docker container's state is "Exited".

<img width="2460" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/39bb468b-0af3-4fe0-9c7b-e79b1f403920">

<img width="976" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/7252a73f-b76c-4aac-878b-5e779c8026ea">

<img width="2560" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/ef04d46a-8b70-4ef4-80bd-bf394a6070fb">

Finenyaco:
**Environment**
- Walrus version: main 3d6c9b2deb4970b456b6eff301c158e7543645db
- UI version: dev-6581d77

**Test Result:**
- Environment list page: failed 
<img width="2394" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/3018b79d-d51f-45c2-962c-e3ac6e01d6dd">

- Resource details page: pass
[...]

PR #1737, opened 2023-12-14
feat: Classify inactive status in resource API

shanewxy:
**Problem:**

Resource API cannot differentiate inactive status from active status.

**Solution:**

Add an `InActive` field to status summary and update it in WalkResource.

**Related Issue:**
#1500

### DIFF
--- a/pkg/dao/types/status/status_resource.go
+++ b/pkg/dao/types/status/status_resource.go
@@ -44,14 +44,52 @@ var resourceStatusPaths = NewWalker(
 	},
 	func(d Decision[ConditionType]) {
 		d.Make(ResourceStatusDeleted,
-			func(st ConditionStatus, reason string) (display string, isError, isTransitioning bool) {
+			func(st ConditionStatus, reason string) *Summary {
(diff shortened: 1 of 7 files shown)


pkg/dao/types/status/status_resource.go
```diff
@@ -58,5 +58,11 @@ var resourceStatusPaths = NewWalker(
 
 // WalkResource walks the given status by resource flow.
 func WalkResource(st *Status) *Summary {
-	return resourceStatusPaths.Walk(st)
+	summary := resourceStatusPaths.Walk(st)
```
gitlawr, line 61: It's not limited to resource summary. How do we declare components like completed jobs/pods are inactive?
image

shanewxy: Updated. Changed the Decide function to return a `Summary` directly, which encapsulates inactive status.

thxCode (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1528, opened 2023-11-23
Be able to remove context variable in schema


gitlawr:

Finenyaco:
**Environment**
- Walrus version: main 9735afb411c8b8bef83e11d048bb6a93494ac3ff
- UI version: dev-f0b97d5

**Test Result:**
- UI
1. Add a template with source `https://github.com/unitong/test-variable?ref=main`.
2. Go to the template details page and check the UI schema.

Still shows `context`.
<img width="1252" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/e4593f11-3a6c-42cb-b13c-249bb2ae1eba">

- CLI: block by https://github.com/seal-io/walrus/issues/1788

Finenyaco:
**Environment**
- Walrus version: main 7428401fef1d1a45e433c2e7722597a5e12830ef
- UI version: dev-6762e0b

**Test Result:**
Import a template without schema.yaml will still generate context variable for UI schema.

https://github.com/Finenyaco/terraform-kubernetes-containerservice?ref=main

<img width="1419" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f430389f-6eea-4315-a6ff-97dfc23933b2">

Finenyaco:
**Environment**
- Walrus version: main 98e5445080de1e1017db16951dd12d8e1e52b27f
- UI version: dev-3fcbabe

**Test Result:** pass

PR #1739, opened 2023-12-15
fix: hidden context variable in UI schema

alexcodelf:
**Problem:**

UI schema should not show context. Walrus will auto inject context when deploying, user does not care about this field.

**Solution:**

Since we need keep it in schema of the template to determine whether inject walrus context to the template. **The proposed solution involves excluding the context field when generating the UI schema file for the template. If users wish to modify the UI schema in Walrus, the context field will not be hidden, allowing for customization.**

**Related Issue:**
#1528

### DIFF
--- a/pkg/templates/git.go
+++ b/pkg/templates/git.go
@@ -31,7 +31,7 @@ func CreateTemplateVersionsFromRepo(
 	mc model.ClientSet,
 	entity *model.Template,
 	versions []*version.Version,
-	versionSchema map[*version.Version]types.TemplateVersionSchema,
+	versionSchema map[*version.Version]*schemaGroup,
 ) error {
 	logger := log.WithName("template")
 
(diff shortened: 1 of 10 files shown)


gitlawr:
What happen when users edit and save a custom UI schema

alexcodelf:
> What happen when users edit and save a custom UI schema

Users will not see the context field in `UIshcema` and rendered form，the deployment will run normally as we using `schema` to determine whether to inject context

pkg/templates/git.go
```diff
@@ -158,7 +163,8 @@ func syncTemplateFromRef(
 			Name:       entity.Name,
 			Version:    ref,
 			Source:     source + "?ref=" + repo.Reference,
-			Schema:     *schema,
+			Schema:     *originSchema,
+			UiSchema:   fileSchema.Expose(),
```
gitlawr, line 167: What happen when we revert uiSchema to the default?

alexcodelf: The reset logic before will reset all template version and UI schema. Using template version bus to revert to default.

aiwantaozi (approved):

gitlawr (approved):
LGTM

Issue #1466, opened 2023-11-20
Support Walrus file


gitlawr:
Support deploying Walrus services/resources via a yaml format file

Finenyaco:
Verified in c7fc65ae3ea5546452dd515722f9f0c0712c3e15.

PR #1742, opened 2023-12-17
Support walrus file apply and export

aiwantaozi:
**Problem:**
Need walrus file apply, delete and export
**Solution:**
1. CLI support walrus file apply and delete
2. server side support walrus file apply and export 

**Related Issue:**
#1466


pkg/apis/environment/extension.go
```diff
@@ -353,3 +358,131 @@ func (h Handler) RouteStop(req RouteStopRequest) error {
...
+	return name
+}
+
+func toResourceSpec(req RouteExportRequest, res *model.Resource) types.ResourceSpec {
+	spec := types.ResourceSpec{
+		Project: &types.MetadataName{
+			Name: req.Project.Name,
+		},
+		Environment: &types.MetadataName{
+			Name: req.EnvironmentQueryInput.Name,
```
gitlawr, line 471: What happen if we export a resource from env A then apply to env B?

aiwantaozi: While the exported file(A) includes the scope(project/environment), and the user applying with this file(Apply to B) will use the scope in the file. 
While the exported file doesn't include scope, apply will use the scope from API path params. To make the exported file apply/create to another env, prefer not to include the scope in the file.

gitlawr (approved):
LGTM

pkg/cli/formatter/yaml.go
```diff
@@ -37,5 +38,10 @@ func (f *YamlFormatter) Format(resp *http.Response) ([]byte, error) {
 		return nil, err
 	}
 
+	// Response status is not 200.
```
alexcodelf, line 41: How about compare status code before Unmarshal  body?

pkg/cli/formatter/json.go
```diff
@@ -41,5 +42,10 @@ func (f *JsonFormatter) Format(resp *http.Response) ([]byte, error) {
 		return nil, err
 	}
 
+	// Response status is not 200.
```
alexcodelf, line 45: How about compare status code before Unmarshal body?

pkg/cli/manifest/request.go
```diff
@@ -0,0 +1,603 @@
...
+	if patchOpt == nil {
+		return success, failed, fmt.Errorf("not found patch operation for %s", group)
+	}
+
+	var (
+		results = make(chan result, 1)
+		all     = objs.All()
+	)
+
+	for _, obj := range all {
```
alexcodelf, line 64: As our server has rate limit when object number is large will rate limit cause some patch failed?

pkg/apis/runtime/request_stream.go
```diff
@@ -34,7 +34,10 @@ func (r RequestUnidiStream) Write(p []byte) (n int, err error) {
 
 // SendMsg sends the given data to client.
 func (r RequestUnidiStream) SendMsg(data []byte) error {
+	data = append(data, "\n\n"...)
```
alexcodelf, line 37: Why we need two line breaks here?

staging/utils/strs/rune.go
```diff
@@ -52,3 +52,14 @@ func FillLeft(s string, w int) string {
 func FillRight(s string, w int) string {
 	return runewidth.FillRight(s, w)
 }
+
+// HasSuffix returns true if s ends with suffix.
+func HasSuffix(s string, suffixes ...string) bool {
+	for _, suffix := range suffixes {
+		if len(s) >= len(suffix) && s[len(s)-len(suffix):] == suffix {
```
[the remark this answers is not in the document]
aiwantaozi: The input suffixes mean could compare multiple suffixes and return true when matching one, just update the function comment.

pkg/apis/environment/extension.go
```diff
@@ -353,3 +358,125 @@ func (h Handler) RouteStop(req RouteStopRequest) error {
...
 	})
 }
+
+func (h Handler) RouteApply(req RouteApplyRequest) error {
+	sj := session.MustGetSubject(req.Context)
+
+	token, err := auths.GetAccessToken(
+		req.Context, h.modelClient,
+		sj.ID, types.TokenKindAPI, types.WalrusOperationTokenName)
+	if err != nil {
```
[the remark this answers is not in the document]
aiwantaozi: updated

pkg/cli/formatter/yaml.go
aiwantaozi: Unmarshal body before comparing the status is used for output the body while status is not 200.

pkg/cli/manifest/request.go
aiwantaozi: Now it will retry three times.

pkg/apis/runtime/request_stream.go
aiwantaozi: This is used for separating each message, then could decode the event object from stream.

alexcodelf (approved):
LGTM

Issue #1463, opened 2023-11-18
Support service/resource redeploy button & batch redeploy & batch stop/start


linyinli:
**Is your enhancement related to a problem? Please describe.**
Batch redeploy/stop/start is not supported and the need for each service/resource to select edit-save/stop/start to trigger redeployment/stop/start is cumbersome.

<img width="1281" alt="image" src="https://github.com/seal-io/walrus/assets/42955482/9b7257ac-711d-48be-8639-ca0dca802abb">

**Describe the solution you'd like**
Support service/resource redeploy button & batch redeploy & batch stop/start

Finenyaco:
**Environment**
- Walrus version: main a32da81f4a08410f854d9897f88885b7a5dc63fd
- UI version: dev-04f141b

**Test Result:**

When deploying in batches, please add a confirmation window. Otherwise, accidental triggers may lead to the interruption of services in the production environment.

<img width="2459" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/2768997e-744e-429e-8692-9a5cca122516">

Finenyaco:
**Environment**
- Walrus version: main a32da81f4a08410f854d9897f88885b7a5dc63fd
- UI version: dev-04f141b

**Test Result:**

Resources with interdependencies cannot be stopped.

<img width="684" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/aaae3714-a74c-440d-9e67-ba461760baf1">

<img width="2464" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/d69d5427-da26-46b2-ae3d-36f2f32f1d0c">

Finenyaco:
**Environment**
- Walrus version: main 645047fbb578508e9b8ab46353195be7de52f22e
- UI version: dev-b511a81

**Test Result:**
- Resources with interdependencies: pass
- Confirmation window when redeploy in batch: pas
- Other issue: https://github.com/seal-io/walrus/issues/1759

PR #1743, opened 2023-12-18
feat: support batch resource operations

alexcodelf:
**Problem:**

At this stage, there is no redeployment of operations. Each deployment requires editing and then upgrading, which is very inconvenient for users. Moreover, walrus does not support batch start and stop resources.

**Solution:**

- add reapply/batch reapply APIs
- add batch stop/start APIs
- set resource label stoppable in dev and staging by default.

**Related Issue:**
#1463 #1765

### DIFF
--- a/pkg/apis/resource/extension.go
+++ b/pkg/apis/resource/extension.go
@@ -33,7 +33,22 @@ import (
 )
 
 func (h Handler) RouteUpgrade(req RouteUpgradeRequest) error {
-	entity := req.Model()
+	var (
+		entity *model.Resource
+		err    error
+	)
+
+	if req.ReuseAttributes {
+		entity, err = h.modelClient.Resources().Query().
+			Where(resource.ID(req.ID)).
+			Only(req.Context)
+		if err != nil {
+			return err
+		}
+	} else {
+		entity = req.Model()
+	}
+
 	return h.upgrade(req.Context, entity, req.Draft)
 }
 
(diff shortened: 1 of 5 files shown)


pkg/apis/resource/extension.go
```diff
@@ -561,3 +566,70 @@ func (h Handler) RouteGetGraph(req RouteGetGraphRequest) (*RouteGetGraphResponse
 		Edges:    edges,
 	}, nil
 }
+
+func (h Handler) RouteReApply(req RouteReApplyRequest) error {
```
gitlawr, line 570: How do we show in the UI, should we use redeploy?

alexcodelf: Deploy may be better, it is weird to show re-deploy in UI.

aiwantaozi:
We have create/upgrade and deploy API quite confusing, how about adding an option reuse-attribute-values to the upgrade API？

pkg/resource/resource.go
```diff
@@ -528,3 +528,24 @@ func IsInactive(r *model.Resource) bool {
 	return r.Status.SummaryStatus == status.ResourceStatusUnDeployed.String() ||
 		r.Status.SummaryStatus == status.ResourceStatusStopped.String()
 }
+
+func SetDefaultLabels(r *model.Resource, env *model.Environment) error {
+	if r == nil || env == nil {
+		return errorx.Errorf("resource or environment is nil")
+	}
+
+	// Only set default labels if labels are not set.
```
gitlawr, line 537: Should we set labels when the key is not present?

gitlawr (approved):
LGTM

pkg/apis/resource/extension_view.go
```diff
@@ -36,14 +37,19 @@ type RouteUpgradeRequest struct {
...
+	Draft           bool `json:"draft,default=false"`
+	ReuseAttributes bool `json:"reuseAttributes,default=false"`
 }
 
 func (r *RouteUpgradeRequest) Validate() error {
 	if err := r.ResourceUpdateInput.Validate(); err != nil {
 		return err
 	}
 
+	if r.ReuseAttributes {
```
aiwantaozi, line 49: prefer still validate while reusing attributes, since template and definition may be changed

aiwantaozi (approved):
LGTM

Issue #1740, opened 2023-12-15
Resource is not deployed imediately on clone/batch creation


gitlawr:
**Is your enhancement related to a problem? Please describe.**
When we clone resources or call batch creation via API, they are in progressing for some period before the deployment start.

**Describe the solution you'd like**
Imediately start the deployment if there's no pending dependencies.

Finenyaco:
**Environment**
- Walrus version: main 8048eb2af52e636e42e656ee507d875cb83999d9
- UI version: dev-33d2ecd

**Test Result:** pass

PR #1747, opened 2023-12-19
feat: instant deploy resource without dependencies

alexcodelf:
**Problem:**

Resource is not deployed imediately on clone/batch creation.

**Solution:**

Support imediately deploy resource when resource does not have dependencies.

**Related Issue:**
#1740


gitlawr (approved):
LGTM

aiwantaozi (approved):

Issue #1736, opened 2023-12-14
Support catalog from Gitee and provide built-in catalog mirror in Gitee


gitlawr:
**Is your enhancement related to a problem? Please describe.**
It helps when it's inconvenient to access github for the built-in templates

Finenyaco:
**Environment**
- Walrus version: main a34a2926d722730c5cbecda30238b778ba0329f0
- UI version: dev-98d0e3e

**Test Result:** 
setup walrus with `--builtin-catalog-provider=gitee`.

1. Check built-in catalog source.
<img width="806" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/7377cfb3-7524-4d93-ad59-1201375aacd9">

2. Check deployment logs.
<img width="915" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/67d75dcc-32f1-4dde-b7c0-9cf1032428b4">

3. But cannot get the template icon.
<img width="2437" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/7598be19-c04a-4ca6-9b81-7249be0e57ad">

Finenyaco:
**Environment**
- Walrus version: main 278965251ef397a0f67ffab3c9348f4243d92c56
- UI version: dev-bb09fa3

**Test Result:** pass

<img width="2422" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/41035ed7-519d-49dd-9b4a-fc0953550482">

PR #1756, opened 2023-12-25
feat: support gitee catalog

alexcodelf:
**Problem:**

Template Catalog is not supported.

**Solution:**

Add gitee client to support gitee catalog. This is partial Pull Request to address #1736 we need support catalog server mirror [walrus-catalog](https://github.com/walrus-catalog) to gitee.

**Additional**
cc @hibig UI need a new catalog type `Gitee`, one thing to note is that template icons cannot be displayed from gitee, as gitee has implemented measures to prevent hot-linking of images, all image link will return 403. The possible measure is to set `<meta name="referrer" content="no-referrer">`

**Related Issue:**
#1736

### DIFF
--- a/pkg/server/runner.go
+++ b/pkg/server/runner.go
@@ -31,6 +31,7 @@ import (
 	"github.com/seal-io/walrus/pkg/cron"
 	"github.com/seal-io/walrus/pkg/dao/model"
 	_ "github.com/seal-io/walrus/pkg/dao/model/runtime" // Default = ent.
+	"github.com/seal-io/walrus/pkg/dao/types"
 	"github.com/seal-io/walrus/pkg/dao/types/crypto"
 	"github.com/seal-io/walrus/pkg/database"
 	"github.com/seal-io/walrus/pkg/datalisten"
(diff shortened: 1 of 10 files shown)


gitlawr (dismissed):
LGTM

pkg/server/runner.go
```diff
@@ -136,6 +139,20 @@ func (r *Server) Flags(cmd *cli.Command) {
...
 			Value:       r.EnableTls,
 		},
+		&cli.StringFlag{
+			Name:        "builtin-catalog-provider",
+			Usage:       "Builtin catalog provider, Github or Gitee.",
+			Destination: &r.BuiltinCatalogProvider,
+			Value:       r.BuiltinCatalogProvider,
+			Action: func(c *cli.Context, s string) error {
+				switch s {
+				case types.GitDriverGitee, types.GitDriverGithub:
```
gitlawr, line 149: Should we make it case insensitive in cli args?

alexcodelf: Updated, it's value is case-insensitive now.

gitlawr (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #1203, opened 2023-08-25
Support adding local cluster to make k8s deployment work out of the box


aiwantaozi:
**Is your enhancement related to a problem? Please describe.**

**Describe the solution you'd like**
Add a flag to support adding local k3s to the default project, which makes it easier to try out.

thxCode:
After a discussion with the internal team, we are not going to support this at present. The point of view is as follows,

1. Walrus is a platform to shorten the path of application deployment, the Kubernetes cluster is not the only target that Walrus focuses on. Walrus leverages a deployer, at present, a Terraform CLI, to finish the application deployment. Running a deployment is typically an elastic, asynchronous, and inherently stateless process, so a Kubernetes cluster matches this perfectly. In short, how to finish a deployment is what Walrus cares about, but where to deploy is what users care about.
2. Walrus provides an embedding Kubernetes cluster, it's like when you buy a cell phone you always have a battery. If users really want to deploy something to the embedding Kubernetes cluster, or the Kubernetes cluster specified by `--kube-config`, please try [add a Kubernetes Connector](https://seal-io.github.io/docs/concepts#connectors).
3. For a quick start, there are many ways to have a Kubernetes cluster now, [Kind](https://kind.sigs.k8s.io/docs/user/quick-start/), [K3d](https://k3d.io/), [Docker Desktop](https://docs.docker.com/desktop/kubernetes/), [Rancher Desktop](https://rancherdesktop.io/) and cloud Kuberentes services. The only concern for the user is to ensure that the target cluster is reachable from the Walrus's "battery" cluster.

orangedeng:
Thanks for taking this seriously. My point is that the user will get distracted very easily when deploying a new k8s. The public cloud is not the first choice in many case(mostly because of the cost). And like you said, people can deploy their own cluster by them selves, but also they will lost focus when finding out how to do it.

In quick start docs, you can add k8s cluster install link to guide user to do so.
Or as the suggestion above, you already have the embed k3s cluster, why not expose it to let people get to the core of Walrus?

rainfd:
The current version is not yet roubust enough to allow users to arbitrarily modify walrus internal k8s resources. And it is not certain that future versions will still use k8s as the basis for our orchestration. @orangedeng

orangedeng:
Yeah, i get that. I also don't think the embed k3s is a good idea to run terraform job. So adding the k3s/k3s quick start link to the docs should be enough.

gitlawr:
[...]

PR #1772, opened 2023-12-27
refactor: use nodeport to expose services in embedded cluster

gitlawr:
**Problem:**
Current proxy implementation does not work if the exposed service requires accessing the root path.

**Solution:**
Use node port to expose embedded services like normal clusters.
Some adaptations for the embeded cluster:
- Shrink node port range of embedded k3s to avoid long bootstrap time.
- Use hostname of walrus server instead of the discovered node IP because it's not accessbile externally.

To make it work, users need to start walrus with additional port mappings:

$ docker run -d -p 80:80 -p 443:443 **-p 30000-30100:30000-30100** --privileged sealio/walrus

Reasons not using host network:
1. Occupied ports are not clear.
2. It can easily conflicts with ports from other services. For example when developing walrus with a local k8s.

**Related Issue:**
https://github.com/seal-io/walrus/issues/1203

### DIFF
--- a/pkg/dao/types/built_in_labels.go
+++ b/pkg/dao/types/built_in_labels.go
@@ -26,6 +26,6 @@ const (
 	// LabelResourceStoppable indicates if the resource is stoppable.
 	LabelResourceStoppable string = "walrus.seal.io/stoppable"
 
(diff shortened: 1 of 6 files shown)


thxCode (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #1445, opened 2023-11-17
Add more details when rolling back a service or resource


linyinli:
**Is your enhancement related to a problem? Please describe.**
When rolling back a service or resource, only the timestamp and diff are showed when selecting a revision. We cannot get more information such as revision details or who generated the revision. I suggest it should  include more details when rolling back.

**Describe the solution you'd like**
1. Add an optional comment when deploying a service or resource, and record both the comment and the operator's information in the revision.
2. Show the comment and the operator's information when rolling back a service or resource.

Finenyaco:
**Environment**
- Walrus version: main 98e5445080de1e1017db16951dd12d8e1e52b27f
- UI version: dev-3fcbabe

**Test Result:** pass

<img width="766" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f2b71915-dada-46ba-ad37-bd6e6f29dc36">

PR #1774, opened 2023-12-27
feat: record optional comment and create user in resource revision

shanewxy:
**Problem:**

When rolling back a service or resource, only the timestamp and diff are showed when selecting a revision. Need record an optional comment and the user generated the revision. 

**Solution:**

- Add `create_user` and `comment` to resource revision schema. Add `comment` to resource schema for scheduled resource.  
- Add `comment` to resource start and rollback api.

**Related Issue:**
#1445


pkg/dao/schema/resource.go
```diff
@@ -64,6 +64,12 @@ func (Resource) Fields() []ent.Field {
 		property.ValuesField("attributes").
 			Comment("Attributes to configure the template.").
 			Optional(),
+		field.String("comment").
+			Comment("Comment of the revision.").
```
gitlawr, line 68: Why do we store revision comment in resources?

shanewxy: Resources pending on relationships are not applied immediately when API called. Using this comment when they are applied in scheduled tasks.

pkg/dao/schema/resource_revision.go
```diff
@@ -79,6 +79,12 @@ func (ResourceRevision) Fields() []ent.Field {
 		field.Text("record").
 			Comment("Record of the revision.").
 			Optional(),
+		field.String("comment").
+			Comment("Comment of the revision.").
+			Optional(),
+		field.String("create_user").
```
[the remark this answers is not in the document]
shanewxy: updated

pkg/dao/schema/resource.go
gitlawr: What about `"change_comment": "Change comment of the resource."`

gitlawr: It may also apply to revison to make it consistent and clear.

pkg/dao/schema/resource_revision.go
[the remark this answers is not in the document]
shanewxy: Having checked schema, annotation is skipped in output. Considered creator is used to display on UI, better using a field?

alexcodelf:
Please consider batch operations.

pkg/dao/schema/resource.go
shanewxy: updated

alexcodelf (approved):
LGTM

pkg/apis/resource/extension.go
```diff
@@ -161,10 +162,12 @@ func (h Handler) RouteRollback(req RouteRollbackRequest) error {
 }
 
 func (h Handler) RouteStart(req RouteStartRequest) error {
-	return h.start(req.Context, req.resource)
+	return h.start(req.Context, req.resource, req.ChangeComment)
 }
 
-func (h Handler) start(ctx context.Context, entity *model.Resource) error {
+func (h Handler) start(ctx context.Context, entity *model.Resource, comment string) error {
```
gitlawr, line 168: Do we need the `comment` param of the function? It's `entity`'s property.

pkg/apis/resource/extension_view.go
```diff
@@ -179,7 +179,8 @@ type RouteRollbackRequest struct {
 
 	model.ResourceQueryInput `path:",inline"`
 
-	RevisionID object.ID `query:"revisionID"`
+	RevisionID    object.ID `query:"revisionID"`
+	ChangeComment string    `query:"change_comment"`
```
gitlawr, line 183: Use camel case in API fields

pkg/dao/schema/resource.go
```diff
@@ -64,6 +64,12 @@ func (Resource) Fields() []ent.Field {
 		property.ValuesField("attributes").
 			Comment("Attributes to configure the template.").
 			Optional(),
+		field.String("change_comment").
+			Comment("Change comment of the revision.").
```
gitlawr, line 68: s/revision/resource/

gitlawr (approved):
LGTM

Issue #736, opened 2023-06-05
Cache terraform providers


thxCode:
provide the approach to speed up terraform plugin initialization, the following are two ways to be PoC.

1. leverage terraform configuration https://developer.hashicorp.com/terraform/cli/config/config-file, provide pre-installed caching terraform providers to https://github.com/seal-io/terraform-deployer.
2. pretend seal server as terraform registry https://developer.hashicorp.com/terraform/cli/config/config-file#network_mirror, provide terraform provider management from seal.

need to hijack the version of the providers that declare in the module.

<img width="506" alt="image" src="https://github.com/seal-io/seal/assets/3394724/d160dea1-cf2f-4c0d-9712-929c01894339">

Finenyaco:
Verified in 278965251ef397a0f67ffab3c9348f4243d92c56.

PR #1791, opened 2023-12-29
feat: support configuring deployer network mirror

thxCode:
- feat: support configuring deployer network mirror
- refactor: argo-workflows chart installer

#736 

after this merge, @hibig we should expose a new setting from UI.


alexcodelf (approved):
LGTM

PR #1799, opened 2024-01-02
fix: cli may crash when generate schema in ci

aiwantaozi:
**Problem:**
CLI may crash when generate schema in CI

**Solution:**
Should not return nil when file is not existed

**Issue:**
https://github.com/seal-io/walrus/issues/1800


thxCode (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1803, opened 2024-01-02
Failed to deploy resources: namespaces "default-test" not found


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create a resource in local environment.
2. Check the deployment result.

**Result**

1. Failed to deploy resources: `Error: namespaces "default-test" not found`. But the namespace should be `default-local`.
<img width="1544" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/2d56749c-dc1d-4e9a-bc8a-8dc781fe6745">

**Expected behavior**

**Environment**
- Walrus version: main c7df6180422f672786d170721dc2afd0d8ad30f6
- UI version: dev-04f141b

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main a32da81f4a08410f854d9897f88885b7a5dc63fd
- UI version: dev-98d0e3e

**Test Result:** pass

PR #1804, opened 2024-01-02
fix: get wrong schema for template version

alexcodelf:
**Problem:**

Deploy resource to wrong namespace.

**Solution:**

The reason is that fetching origin schema with wrong method.

**Related Issue:**
#1803


thxCode (approved):
LGTM

aiwantaozi (approved):

Issue #1797, opened 2024-01-02
fail to apply walrus file


gitlawr:
**Steps to reproduce**
1. Import the following yaml
```
resources:
- name: nginx
  template:
    name: kubernetes-containerservice
    version: v0.1.3
  attributes:
    containers:
    - image: nginx
      ports:
      - internal: 80
        external: 80
```

**Result**
<img width="802" alt="image" src="https://github.com/seal-io/walrus/assets/5697937/1dccfec9-03aa-4cb2-844b-992c076bfa1a">

**Environment**
- Walrus version: dev-babd256

PR #1805, opened 2024-01-02
fix: should not lock operation name since we may add new batch operation

aiwantaozi:
**Problem:**
Used to lock collection post request to batch-create and batch-delete, it causes problem since this PR https://github.com/seal-io/walrus/pull/1743 add more batch post operation

**Solution:**
Generate operation name based on go func

**Related Issue:**
#1797


alexcodelf (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1806, opened 2024-01-02
Failed to add connectors to an environment without connector: panic as assignment to entry in nil map


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create an environment without connectors.
2. Edit the environment and add a connector.

**Result**

1. Failed to update the environment: ` panic as assignment to entry in nil map`
<img width="2431" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/8c346eff-fe4b-45ef-b1e5-8a7addd2c51e">

```
2024-01-02T10:04:33.277Z	E	api	error requesting PUT /v1/projects/494348894252100866/environments/494357889356459266: 1 error occurred:
	* panic as assignment to entry in nil map


github.com/seal-io/walrus/pkg/apis/runtime.erroring
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_error.go:45
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.recovering
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_recovery.go:39
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.observing
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_observation.go:89
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/gin-gonic/gin.(*Engine).handleHTTPRequest
	github.com/gin-gonic/gin@v1.9.1/gin.go:620
github.com/gin-gonic/gin.(*Engine).ServeHTTP
	github.com/gin-gonic/gin@v1.9.1/gin.go:576
github.com/seal-io/walrus/pkg/apis/runtime.(*Router).ServeHTTP
	github.com/seal-io/walrus/pkg/apis/runtime/router.go:193
net/http.serverHandler.ServeHTTP
	net/http/server.go:2947
net/http.initALPNRequest.ServeHTTP
	net/http/server.go:3556
net/http.(*http2serverConn).runHandler
	net/http/h2_bundle.go:5910
[...]

PR #1807, opened 2024-01-02
fix: enviornment update failed

alexcodelf:
**Problem:**

Can not update environment when labels is empty of the environment.

**Solution:**

Init labels when environment labels is nil

**Related Issue:**
#1806


shanewxy (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1810, opened 2024-01-03
imported resources may stuck in deploying


gitlawr:
**Steps to reproduce**
Apply the following yaml:
```
resources:
- name: nginx-1
  template:
    name: kubernetes-containerservice
    version: v0.1.3
  attributes:
    containers:
    - image: nginx
      ports:
      - internal: 80
        external: 80
- name: nginx-2
  template:
    name: kubernetes-containerservice
    version: v0.1.3
  attributes:
    containers:
    - image: nginx
      ports:
      - internal: 80
        external: 80
      envs:
      - name: nginx-address
        value: ${res.nginx-1.address}
```

**Result**
nginx-2 stuck in `Deploying`

**Expected behavior**
Deployment succeeds.

**Environment**
- Walrus version: main

**Additional context**
Logs:
```
2024-01-03T06:01:34.521Z	E	task	panic in job: github.com/seal-io/walrus/utils/cron.timeoutTask.Process-fm, recover data: failed to load subject: failed to get context: not gin context
github.com/seal-io/walrus/utils/cron.init.0.func1
	github.com/seal-io/walrus/utils@v0.0.0-00010101000000-000000000000/cron/cron.go:39
github.com/go-co-op/gocron.(*executor).run.func1.1
	github.com/go-co-op/gocron@v1.29.0/executor.go:216
runtime.gopanic
	runtime/panic.go:884
github.com/seal-io/walrus/pkg/auths/session.MustGetSubject
	github.com/seal-io/walrus/pkg/auths/session/subject.go:214
github.com/seal-io/walrus/pkg/deployer/terraform.Deployer.createRevision
	github.com/seal-io/walrus/pkg/deployer/terraform/deployer.go:505
github.com/seal-io/walrus/pkg/deployer/terraform.Deployer.Apply
	github.com/seal-io/walrus/pkg/deployer/terraform/deployer.go:113
github.com/seal-io/walrus/pkg/resource.Apply
	github.com/seal-io/walrus/pkg/resource/resource.go:99
github.com/seal-io/walrus/pkg/scheduler/resource.(*RelationshipCheckTask).deployResource
	github.com/seal-io/walrus/pkg/scheduler/resource/relationship_check_task.go:267
github.com/seal-io/walrus/pkg/scheduler/resource.(*RelationshipCheckTask).applyResources
	github.com/seal-io/walrus/pkg/scheduler/resource/relationship_check_task.go:127
[...]

Issue #1765, opened 2023-12-26
Add the walrus.seal.io/stoppable label to resource automatically after resource created


linyinli:
Add the walrus.seal.io/stoppable label to resource automatically after resource created:

- For development/staging environments, `walrus.seal.io/stoppable=true` by default

- For production environments, `walrus.seal.io/stoppable=false` by default

linyinli:
A few details need to be added: remove walrus.seal.io/stoppable label when cloning and make it fit to environment type.

thxCode:
Following the description, @linyinli 's first requirement is to add stoppable to all resources, but the main branch only supports adding the stoppable label to the resources below the non-production env. Can @linyinli clarify what the real requirement?

linyinli:
Real requirement:
1. walrus.seal.io/stoppable label need to be added automatically after resources created, so users don't have to look at the documentation when they want to change the stop behavior, whether it's production env or non-production env;
2. For the stability of production, set walrus.seal.io/stoppable=false by default.
3. It only supports adding the stoppable label to the resources below the non-production env, I think it's because in the 0.4 design, the stoppable label of resources was stopped=false by default, this misses the production environment.

Finenyaco:
**Environment**
- Walrus version: main 918a699c606f64575b1ac86f58c93ae899c141c0
- UI version: dev-3f29d01

**Test Result:** pass

- Add walrus.seal.io/stoppable=false to production environment: pass
- Remove the walrus.seal.io/stoppable label when cloning and exporting resource: pass
<img width="1001" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/fa16ac03-1233-46e1-b8e7-24ba6d16b7ae">

PR #1813, opened 2024-01-03
fix: batch operation and job deployment failed, cloned resources labels wrong

alexcodelf:
**Problem:**

- Batch stop selected dependency resources but can not stop. E.G.: A &rarr; B, A and B are both selected to stop will failed.
- Task job deployment job failed.
- Production cloned resources should not contains stoppable labels。
**Solution:**

- Topological resource before all batch operations for resource.
- Get user info from resource annotaion.
- Remove default stoppables label for production resource when cloning.

**Related Issue:**
https://github.com/seal-io/walrus/issues/1463#issuecomment-1873919538 #1810 https://github.com/seal-io/walrus/issues/1765#issuecomment-1874729995


pkg/resource/resource.go
```diff
@@ -552,3 +552,41 @@ func SetDefaultLabels(r *model.Resource, env *model.Environment) error {
 
 	return nil
 }
+
+// SetCloneDefaultLabels sets default labels for cloned resource.
+func SetCloneDefaultLabels(r *model.Resource, env *model.Environment) error {
```
gitlawr, line 557: Why don't we make SetDefaultLabels() handle both cases?

alexcodelf: For user create case in UI, if user intend to fill the `stop` label, we will not remove the `stop` label for user. This is only for clone case.

gitlawr: What if user intend to fill the stop label on env clone? Should we remove the labels in the frontend when clone?

alexcodelf: It maybe better handler this in UI. @hibig

gitlawr: Per discussion we keep this. Users need to explicitly update the label to make it stoppable in production.

gitlawr (approved):
LGTM

gitlawr (approved):

thxCode (approved):
LGTM

Issue #1347, opened 2023-11-01
Display specific error messages when adding Kubernetes connectors fails 


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

**Describe the solution you'd like**

Now adding a connector that fails only returns `unreachable connector: context deadline exceeded`. It doesn't help the problem.

<img width="1327" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/132ed55e-f91a-4c89-8206-2c71859b0d19">

**Describe alternatives you've considered**

Display specific error messages when adding Kubernetes connectors fails as in the past.  For example,
```
Unable to connect to the server: tls: failed to verify certificate: x509: certificate is valid for 10.96.0.1, 172.19.0.2, 0.0.0.0, 192.168.2.5, not x.x.x.x
```

**Environment**
- Walrus version: main 9c5ea546b1918a53aaf1a3173a5f43b10dd5fe3e
- UI version: dev-19c51ab

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 937ca0fd19efd0e7137164f5b81c511f160426b2
- UI version: dev-437b352

**Test Result:** pass

<img width="1115" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/3c843390-31d8-4ad4-9307-0bf08e37615e">

PR #1824, opened 2024-01-04
feat: return the actual error when connectors fails

shanewxy:
**Problem:**

The error returned by the rest client is sometimes context deadline exceeded itself. 

**Solution:**

Record the error to lastErr only if it's not context deadline exceeded.

**Related Issue:**
#1347


thxCode:
1. We should unify the error processing of `pkg/k8s/helper.go#IsConnected` and `pkg/operator/k8s/operator.go#IsConnected`, as almost the same as they are.
2. The root cause is the error reported from the ConditionFunc `context.Context` affects the error reported by the RESTClient. Can we fix it as below and verify the issue again?
```
var lastErr error

err := wait.PollUntilContextCancel(ctx, time.Second, true,
    func(_ context.Context) (bool, error) {
        _, lastErr = op.CoreCli.RESTClient().
            Get().
            AbsPath("/version").
            Do(context.TODO()). // NB(thxCode): Keep the real error from request.
            Raw()

        return lastErr == nil, ctx.Err()
    },
)

if err != nil && lastErr != nil {
    err = lastErr // Use last error to overwrite context error while existed.
}

return err
```

shanewxy:
> 1. We should unify the error processing of `pkg/k8s/helper.go#IsConnected` and `pkg/operator/k8s/operator.go#IsConnected`, as almost the same as they are.
> 2. The root cause is the error reported from the ConditionFunc `context.Context` affects the error reported by the RESTClient. Can we fix it as below and verify the issue again?
> 
> ```
> var lastErr error
> 
> err := wait.PollUntilContextCancel(ctx, time.Second, true,
>     func(_ context.Context) (bool, error) {
>         _, lastErr = op.CoreCli.RESTClient().
>             Get().
>             AbsPath("/version").
>             Do(context.TODO()). // NB(thxCode): Keep the real error from request.
>             Raw()
> 
>         return lastErr == nil, ctx.Err()
>     },
> )
> 
> if lastErr != nil {
>     err = lastErr // Use last error to overwrite context error while existed.
> }
> 
> return err
> ```

updated

thxCode (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #1688, opened 2023-12-05
Default values of object type variables not returned


Finenyaco:
**Describe the bug**

Default value of object type variable not returned

**Steps to reproduce**
1. Add a template with source `https://github.com/unitong/test-variable`.
2. Enable the template for service.
3. Create a service using the template added above.
4. Check the default of `test_object`. 

Definition of `test_object`.
```
variable "test_object" {
  type          = object({
    name        = string
    age         = number
    email       = optional(object({
        address = string
        domain  = optional(string, "test.com")
    }))
  })
  default       = {
    name        = "Bob"
    age         = 23
    email       = {
        address = "bob"
        domain  = "example.com"
    }
  }
}
```

**Result**

1. Default values are not shown.
<img width="1566" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/649093ed-9734-4496-80e5-7b35d6e58406">

2. Checked the ui schema of GET template version api response, no default values.
<img width="1110" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/5e017fb5-e2b2-4c6c-a490-dd8641512bfd">

**Expected behavior**

Display the default values.

**Environment**
- Walrus version: main ab30b62917ad1ab35c43d37fb7a646efbd9a1572
- UI version: dev-2d14c4a

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main a12b753f2e823aa7a7456110215dceed4da46ba1
- UI version: dev-07fa8db

**Test Result:** pass

<img width="2390" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/aa9acfa4-94af-4242-ad0a-82f7820e85af">

Issue #1356, opened 2023-11-06
Incorrect values are displayed when the variable data type is not specified


Finenyaco:
**Describe the bug**

Incorrect values are displayed when the variable data type is not specified

**Steps to reproduce**
1. Add a template with source `https://github.com/unitong/test-variable`. The variable `test_map` is defined as follows.
```
variable "test_map" {
  default     = {
    name      = "John"
    age       = "30"
    email     = "john@example.com"
  }
}
```
2. Check the template form.

**Result**

1. The default value is incorrect.
<img width="1208" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/34932b1d-a81b-49f2-85de-f7d6421c1199">

2. Checked the GET template-versions api response. No `type` for this variable.
```
...
"test_map": {
    "default": {
        "age": "30",
        "email": "john@example.com",
        "name": "John"
    },
    "title": "test_map"
},
...
```

**Expected behavior**

Data types can be automatically converted based on default value

**Environment**
- Walrus version: main 6d4903951fc8b6284473bfbae8fe8f0fea34ecd4
- UI version: dev-d45e562

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 66f1bd1681cc926d6b35a6caed7ec5f793ae103a
- UI version: dev-21d34af

**Test Result:** pass

<img width="1515" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/bd9ce369-259d-4661-b5e6-f1220d23de9c">

PR #1830, opened 2024-01-05
refactor: nullable and default in schema generate

aiwantaozi:
**Problem:**
1. missing default value in some case while generate schema
2. nullable cause UI validate error

**Solution:**
1. Fix the default value generated
2. remove nullable

**Related Issue:**
#1356 
#1688

### DIFF
--- a/pkg/templates/translator/translator.go
+++ b/pkg/templates/translator/translator.go
@@ -18,13 +18,13 @@ type Translator interface {
 }
 
 type Options struct {
-	Name        string
-	Default     any
-	Description string
-	Sensitive   bool
-	Order       int
-	TypeExpress any
-	Nullable    bool
+	Name          string
+	DefaultValue  any
+	DefaultObject any
+	Description   string
+	Sensitive     bool
+	Order         int
+	TypeExpress   any
 }
 
 // SchemaOfType generates openAPI schema from original type.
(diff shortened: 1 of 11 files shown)


pkg/templates/loader/testdata/with_output/expected.yaml
```diff
@@ -13,6 +13,7 @@ openAPISchema:
             x-walrus-ui:
               colSpan: 12
               order: 1
+              widget: YamlEditor
```
gitlawr, line 16: What if we don't set widget in schema? Will UI fall back to use the yaml editor?

pkg/templates/translator/translator.go
```diff
@@ -18,13 +18,13 @@ type Translator interface {
...
-	Name        string
-	Default     any
-	Description string
-	Sensitive   bool
-	Order       int
-	TypeExpress any
-	Nullable    bool
+	Name          string
+	DefaultValue  any
+	DefaultObject any
```
gitlawr, line 23: Why do we need both fields?

aiwantaozi: DefaultObject is used for passing the default for nested object.

gitlawr (approved):
LGTM

shanewxy (approved):
LGTM

Issue #1829, opened 2024-01-05
Failed to rollback resource revision: Unsupported Media Type


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create a resource.
2. Rollback the resource.

**Result**

1. Returned an error: `Unsupported Media Type`.
<img width="1135" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/bec2fa99-172e-4c5f-86ba-373645a9b4ef">

2. Checked the request api.
<img width="1313" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/605e7fb3-c449-4fac-a731-0af9b3523e36">

<img width="283" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/339ae350-4e1a-4092-b370-bd1b12143559">

3. Checked the api definition.
<img width="1617" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/0065610a-7659-47a1-905c-5431f5b16cee">

**Expected behavior**

**Environment**
- Walrus version: main 38bbd2af03b5e73b87a803f864ab2045c79b9fe0
- UI version: dev-bb09fa3

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 937ca0fd19efd0e7137164f5b81c511f160426b2
- UI version: dev-437b352

**Test Result:** pass

PR #1844, opened 2024-01-09
fix: rollback revision comment param

shanewxy:
**Problem:**

Comment parameter should be submitted in payload instead of query.

**Related Issue:**
#1829


thxCode (approved):
LGTM

aiwantaozi (approved):

Issue #1833, opened 2024-01-08
Fail to deploy a containerservice


gitlawr:
**Describe the bug**

**Steps to reproduce**
1. Create a resource
2. Use template, choose kubernetes-containerservice v0.2.1
3. Input image=nginx
4. Save

**Result**
<img width="680" alt="image" src="https://github.com/seal-io/walrus/assets/5697937/5cb08544-696c-4487-8b42-42db92a2b13f">

**Environment**
- Walrus version: main

Finenyaco:
**Environment**
- Walrus version: main 1e59b4feaae335bfa20cbfa170908396806ca7b5
- UI version: dev-437b352

**Test Result:** pass

<img width="2383" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/bf3e2b21-c435-40c4-9b32-b22c74b0ef4c">

PR #1846, opened 2024-01-09
refactor: generate schema which sub schema include any type

aiwantaozi:
**Problem:**
Variable including any type will generate any type schema, this may cause loss of the original type info.

**Solution:**
Refactor generate schema of any type, add widget: YamlEditor for any type 

**Related Issue:**
#1833

### DIFF
--- a/pkg/templates/translator/translator.go
+++ b/pkg/templates/translator/translator.go
@@ -25,6 +25,8 @@ type Options struct {
 	Sensitive     bool
 	Order         int
 	TypeExpress   any
+	// IgnoreWidget used for skipping add widget for any type in sub schema.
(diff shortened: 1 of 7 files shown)


pkg/templates/translator/translator.go
```diff
@@ -25,6 +25,7 @@ type Options struct {
 	Sensitive     bool
 	Order         int
 	TypeExpress   any
+	IgnoreWidget  bool
```
gitlawr, line 28: Can you add comments to explain the option

gitlawr (approved):
LGTM

thxCode (approved):
LGTM

Issue #1708, opened 2023-12-11
Add built-in resource definitions


gitlawr:

gitlawr:
<img width="570" alt="image" src="https://github.com/seal-io/walrus/assets/5697937/6ccf5994-d484-4ceb-8f68-5e0c74e1eb88">
Check if we can use repo topics to manage builtin definitions from the builtin catalog.

Finenyaco:
**Environment**
- Walrus version: main 9f815bb4dc0d746fba0871b8085fd85182b2d41b
- UI version: dev-bd75b3b

**Test Result:**
- The order of matching rules defined by different built-in resource definitions is inconsistent. 
When both Docker and Kubernetes connectors are added to an environment simultaneously, the behavior becomes inconsistent when using different built-in resource definitions .

<img width="758" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/ad8e8366-5fd8-4890-a534-1383ba7df01c">

- For UI: Missing settings to opt out of creating the builtin definitions.

Finenyaco:
Verified in 645047fbb578508e9b8ab46353195be7de52f22e.

PR #1847, opened 2024-01-09
feat: add builtin field to mark builtin resource definitions

shanewxy:
**Problem:**

- Need builtin flag to disable edit and delete on UI.
- The order of matching rules defined by different built-in resource definitions is inconsistent.

**Solution:**

- Add `builtin` field to mark builtin resource definitions.
- Sort the connector type to ensure deterministic order.

**Related Issue:**
#1708


thxCode (approved):
LGTM

aiwantaozi (approved):

PR #1850, opened 2024-01-09
refactor: ui schema should use schema.yaml while it exsited

aiwantaozi:
**Problem:**
UI schema should generate from schema.yaml when exists but doesn't fully implement

**Solution:**
UI schema should use schema from schema.yaml when exists.

**Related Issue:**
#1833


alexcodelf (approved):
LGTM

thxCode (approved):
LGTM

PR #1855, opened 2024-01-10
fix: should use schema.yaml while exist in sync catalog

aiwantaozi:
**Problem:**
Sync from the catalog still shows widget YamlEditor for variable: containers in https://github.com/walrus-catalog/terraform-kubernetes-containerservice 

**Solution:**
Should use schema in schema.yaml while exist

**Related Issue:**
#1833


thxCode (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #1858, opened 2024-01-10
Support computed attributes


aiwantaozi:
**Is your enhancement related to a problem? Please describe.**
Support generates computed attributes base on the schema and user input attributes

Finenyaco:
Verified in 4eef11db3efc80ce93bb17814e5e8fb822d75d45.

PR #1859, opened 2024-01-10
feat: support computed attributes

aiwantaozi:
**Problem:**
Need computed attributes to merge the user input and schema default

**Solution:**
Implement computed attributes

**Related Issue:**
#1858


staging/utils/json/jsonpatch.go
```diff
@@ -27,3 +27,27 @@ func PatchObject(obj, patchObject any) (any, error) {
 
 	return obj, nil
 }
+
+func ApplyPatches(patches ...[]byte) ([]byte, error) {
```
thxCode, line 31: ApplyPatches(doc []byte, patches ...[]byte)

pkg/apis/resource/basic_view.go
```diff
@@ -205,6 +202,14 @@ func (r *PatchRequest) Validate() error {
 		return err
 	}
 
+	// Set computedAttributes.
+	patched.EnvironmentID = r.Environment.Refer.ID()
```
thxCode, line 206: 1. `r.Environment.Refer` cannot guarantee that it always receives an ID from the request, I think you can use `r.Environemnt.ID` or query with `patched` directly.
2. I saw we used `patched.EnvironmentID` previously, why do we need to change it here?

pkg/templates/openapi/traverse.go
```diff
@@ -0,0 +1,25 @@
...
+func PostorderTraversal(root Node) []Node {
+	if root == nil {
+		return nil
+	}
+
+	var result []Node
+	stack := &Stack{}
+	stack.Push(root)
+
+	for len(stack.data) > 0 {
```
thxCode, line 13: make `for len(stack.data) >0` to `for stack.Len() >0 `

pkg/templates/openapi/traverse.go
```diff
@@ -0,0 +1,25 @@
...
+		node := stack.Pop()
+
+		result = append([]Node{node}, result...)
```
thxCode, line 16: reduce slice extending.

pkg/apis/resource/helper.go
```diff
@@ -0,0 +1,243 @@
...
+		Where(environment.ID(entity.EnvironmentID)).
+		WithProject(func(pq *model.ProjectQuery) {
+			pq.Select(
+				project.FieldID,
+				project.FieldName)
+		}).
+		Only(ctx)
+	if err != nil {
+		return nil, err
+	}
```
thxCode, line 39: can we reuse the input validated result?

aiwantaozi: Need query project name here.

thxCode (approved):
LGTM

Issue #1755, opened 2023-12-25
Return endpoints data in the resource list API.


hibig:
**Is your enhancement related to a problem? Please describe.**

1. In the resource list can view the item endpoints.

**Describe the solution you'd like**

1. Return endpoints data in the resource list API.

**Describe alternatives you've considered**

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 8ff16bfe4326036642753b1ba895fa5b0e3b1791
- UI version: dev-567aeb4

**Test Result:** Failed to display the endpoints when there are more than one endpoint.

<img width="2391" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/1ab030ac-78d7-4ba8-9241-480d04e56b0a">
<img width="833" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/5f52a773-768d-49a2-878d-3ff38b541e83">

Finenyaco:
**Environment**
- Walrus version: main 8ff16bfe4326036642753b1ba895fa5b0e3b1791
- UI version: dev-7590f46

**Test Result:** pass

<img width="1162" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/8158a396-bd97-4942-aea1-60746068a9fe">

PR #1861, opened 2024-01-11
refactor: endpoints fetching and display

thxCode:
**Problem:**
We have struggled with endpoint for a long time: https://github.com/seal-io/walrus/issues?q=is%3Aissue+access+endpoints+is%3Aclosed, let's figure out this today.

Walrus relies on the [Operator's GetEndpiont](https://github.com/seal-io/walrus/blob/937ca0fd19efd0e7137164f5b81c511f160426b2/pkg/operator/types/types.go#L51-L55) to discover endpoints from a resource, but the discovered endpoints fail to convey their correct intention.

**Solution:**
For Walrus, a resource endpoint should be functional and accessible. To describe the purpose and access way of an endpoint, a good choice is URL, e.g. `https://host/path/subpath`, or `s3://host/bucket`.

The endpoint is an attribute of a resource, so this PR removes the endpoint record from the resource component's status and records to the resource itself directly.

Since the active endpoint discovery cannot convey the correct intention, so this PR removes all implementations for `GetEndpoints`, and respects the returning result(output) from the resource.

Walrus built-in ContainerService collection supports describing the endpoint's purpose with the schema on ports.
- https://github.com/walrus-catalog/terraform-kubernetes-containerservice/commit/8893c34c0ff2a8e06341f21f154bf54e0fceb2f9
- https://github.com/walrus-catalog/terraform-docker-containerservice/commit/6258147d2eec227175796ed71ab968e3432c6e43

This PR parses the endpoint from an output named `walrus_endpoints` or `endpoints`, and only accepts the `map(string)` type.
- `walrus_endpoints` gives the capability to be compatible with original `endpoints`.

**BREAKING CHANGES**
- route change from `/v1/projects/{project}/environments/{environment}/resources/{resource}/access-endpoints` to `/v1/projects/{project}/environments/{environment}/resources/{resource}/endpoints`, and response change `[{"name":"","endpoints":["",""]}]` to `[{"name":"", "url":""}]`.

**CHORES**
- function style: receiver, comment, and purpose. 

**Related Issue:**
#1755


pkg/apis/resource/extension.go
```diff
@@ -224,9 +217,9 @@ func (h Handler) RouteStop(req RouteStopRequest) error {
 	return pkgresource.Stop(req.Context, req.Client, entity, opts)
 }
 
-func (h Handler) RouteGetAccessEndpoints(req RouteGetAccessEndpointsRequest) (RouteGetAccessEndpointsResponse, error) {
+func (h Handler) RouteGetEndpoints(req RouteGetEndpointsRequest) (RouteGetEndpointsResponse, error) {
```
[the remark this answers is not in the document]
thxCode: this is no single resource watch api.

pkg/deployer/terraform/parse.go
```diff
@@ -354,20 +354,27 @@ func getServiceDependencyOutputs(
...
 		}
 
-		for n, o := range revisionOutput {
+		for n, o := range osm {
 			if _, ok := dependOutputs[n]; !ok {
 				continue
 			}
 
-			outputs[n] = o
+			// FIXME(thxCode): mitigate parser.OutputState to types.OutputValue.
```
gitlawr, line 372: Mitigate or migrate? Is it hard to include this refactoring in this pr?

thxCode: updated

alexcodelf (approved):
LGTM

gitlawr (approved):

Issue #1782, opened 2023-12-28
 Enhance the command for switching environments


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

Currently, using `walrus config setup -e <env-name>` to switch environments by CLI.   For users unfamiliar with the product, this may not be easily discovered.

**Describe the solution you'd like**

Make the command for switching environments more intuitive and understandable.

**Describe alternatives you've considered**

**Additional context**

Finenyaco:
**Environment**
```
clientVersion:
  version: main
  gitCommit: 7428401fef1d1a45e433c2e7722597a5e12830ef
  isDevVersion: true
serverVersion:
  version: main
  gitCommit: 7428401fef1d1a45e433c2e7722597a5e12830ef
  isDevVersion: true
```

**Test Result:**

I think it would be better to automatically print the current context after switching the project/environment. e.g.

```
 ~/seal/ walrus context switch -p default -e local
Current Project: default
Current Environment: local
```

Finenyaco:
**Environment**
clientVersion:
  version: main
  gitCommit: 942e758b93603e9ad40f9da90260c30af5ea350b
  isDevVersion: true
serverVersion:
  version: main
  gitCommit: 942e758b93603e9ad40f9da90260c30af5ea350b
  isDevVersion: true

**Test Result:** pass

```
 ~/ w context switch -p default -e orb-prod
Switched context successfully.
Current Project: default
Current Environment: orb-prod
```

PR #1871, opened 2024-01-11
refactor: print current context after switch

aiwantaozi:
**Problem:**
Inconvenient while switching context, better print the current context

**Solution:**
Print the current context after switch

**Related Issue:**
#1782


shanewxy (approved):
LGTM

Issue #1870, opened 2024-01-11
 When executing scheduled tasks, the operator for the revision is incorrect


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create two resources using the walrus fie as below:
```
resources:
- name: nginx-1
  template:
    name: kubernetes-containerservice
    version: v0.2.3
  attributes:
    containers:
    - image: nginx
      ports:
      - internal: 80
        external: 80
- name: nginx-2
  template:
    name: kubernetes-containerservice
    version: v0.2.3
  attributes:
    containers:
    - image: nginx
      ports:
      - internal: 80
        external: 80
      envs:
      - name: nginx-address
        value: ${res.nginx-1.address}
```
2. Create user called `test` and add it as default project member with 'member' role.
3. Login walrus as `test` user.
4. Stop these two resources in batch and check the revision.
5. Start the two resources in batch and check the revision.

**Result**

1. The operator is wrong for nginx-1 when stop.
<img width="2546" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f9b4930b-0a1a-4c32-b8b4-de4564fca5d8">

2. The operator is wrong for nginx-2 when start.
<img width="2560" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/d95b4159-b2e8-44c4-8037-f8518477eeed">

**Expected behavior**

**Environment**
- Walrus version: main 645047fbb578508e9b8ab46353195be7de52f22e
- UI version: dev-6762e0b

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 8ff16bfe4326036642753b1ba895fa5b0e3b1791
- UI version: dev-7590f46

**Test Result:** 

The operator is incorrect when start/stop an environment in the above case.

Finenyaco:
**Environment**
- Walrus version: main 8c258fa00a3e8c3f2895a5cfb7159ca2b5d7fc1f
- UI version: dev-88faa21

**Test Result:** pass

Issue #1869, opened 2024-01-11
walrus.seal.io/stoppable label is not automatically added after cloning resources


Finenyaco:
**Describe the bug**

walrus.seal.io/stoppable label is not automatically added after cloning resources.

**Steps to reproduce**
1. Clone resources to a dev environment.
2. Check the resource label after the deployment.

**Result**

1. Didn't add `walrus.seal.io/stoppable=true` automatically after deploying.
<img width="1130" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/8693775d-280f-45ef-8683-f81d8f0bebf4">

**Expected behavior**

**Environment**
- Walrus version: main 645047fbb578508e9b8ab46353195be7de52f22e
- UI version: dev-6762e0b

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 942e758b93603e9ad40f9da90260c30af5ea350b
- UI version: dev-b511a81

**Test Result:** pass

PR #1872, opened 2024-01-11
fix wrong subject id for resource operations

alexcodelf:
**Problem:**

- wrong subject id for resource operation like upgrade, start, stop.
- clone resource missing default stop labels

**Solution:**

- update subject id for resources when user do resource operations.
- add misssing default labels for batch create API

**Related Issue:**
#1869 #1870


gitlawr (approved):
LGTM

shanewxy (approved):
LGTM

Issue #1781, opened 2023-12-28
Optimize the order of fields in the exported Walrus file


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

The fields in the exported Walrus file currently do not conform to common reading habits.

```
version: v1
resources:
- attributes:
    containers:
    - image: nginx
      ports:
      - external: 80
        internal: 80
  name: test-nginx
  template:
    name: kubernetes-containerservice
    version: v0.1.3
```

**Describe the solution you'd like**

```
version: v1
resources:
- name: test-nginx
  template:
    name: kubernetes-containerservice
    version: v0.1.3
  attributes:
    containers:
    - image: nginx
      ports:
      - external: 80
        internal: 80
  
```

**Describe alternatives you've considered**

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main f41cb2758ccafbd4e458c1b171345b1987aa3247
- UI version: dev-2c4d8f3


**Test Result:** pass

```
version: v1
resources:
- name: test-nginx
  description: add description
  template:
    name: kubernetes-containerservice
    version: v0.2.3
  labels:
    mytest: abc123
  attributes:
    containers:
    - image: ngnix
      ports:
      - schema: http
        external: 80
        internal: 80
        protocol: tcp
      profile: run
      resources:
        cpu: 0.1
        gpu: null
        memory: 10
    deployment:
      rolling:
        max_surge: 0.25
        max_unavailable: 0.25
      timeout: 300
      replicas: 1

```

PR #1878, opened 2024-01-12
feat: optimize fields order in Walrus file

shanewxy:
**Problem:**

The fields of resource in the exported Walrus file currently do not conform to common reading habits.

**Solution:**

Reorder the fields in the struct and unmarshall the bytes into MapSlice directly to keep the order from struct.

**Related Issue:**
#1781


gitlawr (approved):
LGTM

aiwantaozi (approved):

Issue #1319, opened 2023-10-23
CPU 使用率过高


bean-du:
根据日志信息看不出是什么原因导致 CPU 使用率高达 100 %，本地电脑：CPU intel i58代，4核8线程 2.3Ghz。其他容器正常都在 10%以下

Finenyaco:
感谢反馈，目前提供的信息有些缺失，能否提供如下更多的信息方便我们更好地复现排查问题呢：

1. 系统版本是 win10 还是 win11？
2. docker是基于 hyper-v 还是 wsl 2运行的？
3. 在 CPU 占用率为100%的时候，可否简单描述walrus的运行情况，如部署了多少服务，目标集群是否也在本机运行等？
4. 如若方便请提供发生问题时的日志文件，谢谢

bean-du:
操作系统是 mac os 12.6.6，walrus 只是配置了三个环境，同时指向一个 k8s 集群，并没有在里面启动任何服务 。它是只要启动就会有 100%的 CPU 使用率，日志并没打印什么。
能看到的就是以下信息：
2023-10-20 09:13:37 2023-10-20T01:13:37.760Z    W       task.locker     previous locking is not finished        {"createdBy": "ResourceComponentsDiscoverCronExpr"}
2023-10-20 09:13:37 2023-10-20T01:13:37.762Z    W       task.locker     previous locking is not finished        {"createdBy": "ResourceStatusSyncCronExpr"}
2023-10-20 09:13:37 2023-10-20T01:13:37.765Z    W       task.locker     previous locking is not finished        {"createdBy": "ResourceStatusSyncCronExpr"}
2023-10-20 09:13:37 2023-10-20T01:13:37.765Z    W       task.locker     previous locking is not finished        {"createdBy": "ResourceComponentsDiscoverCronExpr"}
2023-10-20 09:13:37 2023-10-20T01:13:37.767Z    W       task.locker     previous locking is not finished        {"createdBy": "ResourceStatusSyncCronExpr"}
2023-10-20 09:13:37 2023-10-20T01:13:37.769Z    W       task.locker     previous locking is not finished        {"createdBy": "ResourceStatusSyncCronExpr"}
2023-10-20 09:13:37 2023-10-20T01:13:37.769Z    W       task.locker     previous locking is not finished        {"createdBy": "ResourceComponentsDiscoverCronExpr"}
[...]

PR #1884, opened 2024-01-12
fix(cron): configure locker after start scheduler

thxCode:
when analyzing #1319, a mistake in cron scheduler use was found. 

this PR configures the distribution lock before starting the scheduler.

it's noted that this PR doesn't address https://github.com/seal-io/walrus/issues/1319 as fewer clues were provided as I mentioned: https://github.com/seal-io/walrus/issues/1319#issuecomment-1888600304.


aiwantaozi (approved):

gitlawr (approved):
LGTM

Issue #1890, opened 2024-01-15
Upgrading the resource failed with an http status code 400: sql: converting argument $1 type: strconv.ParseInt: parsing \"\": invalid syntax


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create a resource and then upgrade.
2. Check the upgrade result.

**Result**

1. Failed to upgrade: `sql: converting argument $1 type: strconv.ParseInt: parsing \"\": invalid syntax`
<img width="2554" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/4fda9712-d382-42d7-b427-5d442bd42f71">

2. Check the api response.
<img width="670" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/72a3fb79-68c6-488a-8b39-6cc507a4ec49">

**Expected behavior**

**Environment**
- Walrus version: main f41cb2758ccafbd4e458c1b171345b1987aa3247
- UI version: dev-2c4d8f3
**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 8ff16bfe4326036642753b1ba895fa5b0e3b1791
- UI version: dev-567aeb4

**Test Result:** pass

Issue #1889, opened 2024-01-15
Failed to deploy resource, context missing: namespaces "default-test" not found


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Go to `default` project and `local` environment.
2. Create a Kubernetest containerservice resource.
3. Check the deployment result.

**Result**

1. Failed to deploy: `Error: namespaces "default-test" not found`.
<img width="1756" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/7af4dd64-125e-43e5-b08b-b926179898a5">

2. Check the schema of template `kubernetes-containerservice`  version 0.2.3 in the template_versions table.
```
...
"x-walrus-original": {
"sequence": [
    "context",
    "infrastructure",
    "deployment",
    "containers"
]
}
...
```

**Expected behavior**

**Environment**
- Walrus version: main f41cb2758ccafbd4e458c1b171345b1987aa3247
- UI version: dev-2c4d8f3

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 8ff16bfe4326036642753b1ba895fa5b0e3b1791
- UI version: dev-567aeb4

**Test Result:** pass

<img width="2197" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/ff8ac185-874f-4cf0-96ee-32f761b51ca1">

PR #1892, opened 2024-01-15
fix: origin schema context had been removed while exposing ui schema

aiwantaozi:
**Problem:**
1. context in the original schema has been removed while exposing ui schema which caused the failure to deploy
2. upgrade failed

**Solution:**
1. exposing ui schema should not affect the original schema
3. set environment ID while computing attributes in the upgrade

**Related Issue:**
#1889 
#1890


thxCode (approved):
LGTM

Issue #1710, opened 2023-12-11
Local deployment


gitlawr:
Support deploying on local docker

Finenyaco:
**Environment**
- Walrus version: main 66f1bd1681cc926d6b35a6caed7ec5f793ae103a
- UI version: dev-88faa21

**Test Result:** 
- Ubuntu docker: pass
- MacOs docker desktop: Failed to deploy resources.
<img width="1216" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/d8b169a0-7300-45b9-9650-29877ce46989">

Server address:
<img width="556" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/1bdac569-387f-44ea-b66f-87425b9ec591">

gitlawr:
fixed in https://github.com/seal-io/walrus-docker-extension/commit/73f7231d08aec5c0ea4f463da0d2cfcea7da2aea

Finenyaco:
**Environment**
- Walrus version: main 8c258fa00a3e8c3f2895a5cfb7159ca2b5d7fc1f
- UI version: dev-88faa21

**Test Result:** pass

PR #1896, opened 2024-01-15
fix: update ServeUrl in docker mode

gitlawr:
**Problem:**
The host.docker.internal domain only works in particular docker descktop versions, making the deployment fails.

**Solution:**
Set ServeUrl setting by network interface IP in docker mode.

**Related Issue:**
https://github.com/seal-io/walrus/issues/1710


thxCode (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #1900, opened 2024-01-15
Failed to deploy resource using kubernetes-postgresql or kubernetes-mysql: Invalid value for input variable


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create a resource using `kubernetes-postgresql` or `kubernetes-mysql` template.
2. Check the deployment result.

**Result**

1. Failed to deploy: `Invalid value for input variable`

```
Error: Invalid value for input variable

  on main.tf line 74, in module "postgresql":
  74:   seeding = {
  75:     type = "none"
  76:     url = {
  77:       storage = {
  78:         size = 10240
  79:       }
  80:     }
  81:   }

The given value is not suitable for module.postgresql.var.seeding declared at
.terraform/modules/postgresql/variables.tf:172,1-19: attribute "url":
attribute "location" is required.
```

**Expected behavior**

**Environment**
- Walrus version: main 8ff16bfe4326036642753b1ba895fa5b0e3b1791
- UI version: dev-567aeb4

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 42f6b1e6cbcd105e9e99ea8badacace1a799c043
- UI version: dev-2fca30f

**Test Result:** Deploy rds successfully but databases seeding through text failed https://github.com/seal-io/walrus/issues/1919.

PR #1908, opened 2024-01-15
fix: update generate default for schema

aiwantaozi:
**Problem:**
 Now will generate below computed attributes cause terraform validate failed.
```json
{
   "seeding": {
       "type": "none",
        "url": {
            "storage": {
                "size": 10240
             }
        }
   }
}
```
**Solution:**
Update generate computed attributes strategy, only generate optional(fieldname, "default") while the object default existed.
After updated:
```json
{
   "seeding": {
       "type": "none"
   }
}
```

**Related Issue:**
#1900

### DIFF
--- a/pkg/templates/openapi/default.go
+++ b/pkg/templates/openapi/default.go
@@ -7,6 +7,8 @@ import (
 
 	jsonpatch "github.com/evanphx/json-patch"
 	"github.com/getkin/kin-openapi/openapi3"
+	"github.com/tidwall/gjson"
+	"github.com/tidwall/sjson"
 
 	"github.com/seal-io/walrus/utils/gopool"
 )
@@ -17,20 +19,24 @@ type patchResult struct {
 	group string
 }
 
-// GenSchemaDefaultPatch generates the default patch for the schema.
+// GenSchemaDefaultPatch generates the default patch for the variable schema.
+// The input root schema type should be object.
 func GenSchemaDefaultPatch(ctx context.Context, schema *openapi3.Schema) ([]byte, error) {
 	if schema == nil || schema.IsEmpty() {
 		return nil, nil
 	}
 
+	// 1. Sort the nodes under the schema in a pre-order traversal.
 	sortedNodes := PreorderTraversal(&DefaultPatchNode{
 		Schema: schema,
 	})
 
-	// Group nodes by the first part of the id to speed up.
+	// 2. Group nodes by the first part of the id to speed up,
+	// after grouped, each group include the nodes from root to sub nodes in sequence.
 	gns := groupNodes(sortedNodes)
 	resultChan := make(chan patchResult, len(gns))
 
+	// 3. Use goroutine to speed up apply patches from root to sub nodes.
 	for gn, v := range gns {
 		nodes := v
 		name := gn
@@ -38,12 +44,14 @@ func GenSchemaDefaultPatch(ctx context.Context, schema *openapi3.Schema) ([]byte
 		gopool.Go(func() {
 			switch {
 			default:
+				// 3.1 No node found.
 				resultChan <- patchResult{
 					group: name,
 				}
 			case len(nodes) == 1:
-				np := nodes[0].GetPatch()
-				if len(np) == 0 {
+				// 3.1 Root node without sub nodes will set default and return.
+				def := nodes[0].GetDefault()
+				if def == nil {
 					resultChan <- patchResult{
 						group: name,
 					}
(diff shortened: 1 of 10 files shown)


pkg/templates/openapi/default.go
```diff
@@ -42,41 +44,80 @@ func GenSchemaDefaultPatch(ctx context.Context, schema *openapi3.Schema) ([]byte
...
+				var b []byte
+
+				p, err := sjson.SetBytes(b, nodes[0].GetID(), def)
+				if err != nil {
+					resultChan <- patchResult{
+						group: name,
+						err:   fmt.Errorf("set patch node error for %s %s: %w", nodes[0].GetID(), string(p), err),
+					}
+				}
+
```
thxCode, line 66: can we construct a path bytes without using sjson? if no, here also need to return inside the `err != nil` block.

pkg/templates/openapi/default.go
```diff
@@ -42,41 +44,80 @@ func GenSchemaDefaultPatch(ctx context.Context, schema *openapi3.Schema) ([]byte
...
+				// Generate root patch.
+				var b []byte
+
+				p, err := sjson.SetBytes(b, nodes[0].GetID(), nodes[0].GetDefault())
+				if err != nil {
+					resultChan <- patchResult{
+						group: name,
+						err:   fmt.Errorf("set patch node error for %s %s: %w", nodes[0].GetID(), string(p), err),
+					}
+				}
```
thxCode, line 90: ditto

pkg/templates/openapi/stack.go
```diff
@@ -86,35 +89,29 @@ func (n *DefaultPatchNode) Children() []Node {
...
-				log.Errorf("failed to set default value for %s: %v", p, err)
-			} else {
-				sdn.patch = patch
+		var (
+			p   = n.GetID() + ".0"
+			sdn = &DefaultPatchNode{
+				Schema: n.Schema.Items.Value,
+				id:     p,
 			}
+		)
```
thxCode, line 102: sdn := &DefaultPatchNode{
  Schema: n.Schema.Items.Value

  id : n.GetID() + ".0",
  def: n.Default
}

return []Node{sdn}

pkg/templates/openapi/stack.go
```diff
@@ -70,14 +80,7 @@ func (n *DefaultPatchNode) Children() []Node {
...
-				patch, err := sjson.SetBytes(b, p, prop.Value.Default)
-				if err != nil {
-					log.Errorf("failed to set default value for %s: %v", p, err)
-				} else {
-					sdn.patch = patch
-				}
+				sdn.def = prop.Value.Default
 			}
 
 			children = append(
```
thxCode, line 86: preallocate `children` slice.

pkg/templates/openapi/default.go
```diff
@@ -42,41 +44,80 @@ func GenSchemaDefaultPatch(ctx context.Context, schema *openapi3.Schema) ([]byte
...
+						}
+					}
+
+					var (
+						pid = np.GetParentID()
+						pr  = gjson.ParseBytes(p)
+					)
+
+					if pe := pr.Get(pid); pe.Exists() {
+						p, err = jsonpatch.MergePatch(npb, p)
```
thxCode, line 115: if an error occurs, the p will be reset, we should allocate a new local var to receive the merged result, and then assign it to p if no error occurs.

pkg/templates/openapi/default.go
```diff
@@ -42,41 +44,80 @@ func GenSchemaDefaultPatch(ctx context.Context, schema *openapi3.Schema) ([]byte
...
+				if def == nil {
 					resultChan <- patchResult{
 						group: name,
 					}
 
 					return
 				}
 
+				// Generate patch.
+				var b []byte
```
thxCode, line 57: remove this declaration

pkg/templates/openapi/default.go
```diff
@@ -42,41 +44,80 @@ func GenSchemaDefaultPatch(ctx context.Context, schema *openapi3.Schema) ([]byte
...
+						patch: nil,
 					}
 
-					if len(p) == 0 {
-						p = np
+					return
+				}
+
+				// Generate root patch.
+				var b []byte
```
thxCode, line 82: ditto

pkg/templates/openapi/default.go
```diff
@@ -42,41 +44,80 @@ func GenSchemaDefaultPatch(ctx context.Context, schema *openapi3.Schema) ([]byte
...
+
+				for i := 0; i < len(nodes); i++ {
+					np := nodes[i]
+					if np.GetDefault() == nil {
 						continue
 					}
 
-					p, err = jsonpatch.MergePatch(p, np)
+					// Generate sub node patch.
+					var nb []byte
```
thxCode, line 99: remove this declaration.

pkg/templates/openapi/default.go
aiwantaozi: Update to return error while error occurs

thxCode (approved):
LGTM

gitlawr (approved):
LGTM

PR #1914, opened 2024-01-16
fix:  environment start/stop subjectID wrong

alexcodelf:
**Problem:**

Environment resource operation subject not updated. It will cause resource schedule resource creator wrong.

**Solution:**

Save subjectID to resources before create schedule resources tasks.

**Related Issue:**
#1870


shanewxy (approved):
LGTM

pkg/apis/environment/common.go
```diff
@@ -270,3 +269,24 @@ func validateConnectors(ctx context.Context, mc model.ClientSet, ids []object.ID
...
+		return err
+	}
+
+	for i := range resources {
+		res := resources[i]
+		if err := mc.Resources().UpdateOne(res).
+			SetAnnotations(res.Annotations).
+			Exec(ctx); err != nil {
+			return err
+		}
```
thxCode, line 288: split `err:= ...` from the if-block condition.

pkg/apis/environment/common.go
```diff
@@ -270,3 +270,26 @@ func validateConnectors(ctx context.Context, mc model.ClientSet, ids []object.ID
 
 	return nil
 }
+
+func setResourceSubjectID(ctx context.Context, mc model.ClientSet, resources ...*model.Resource) error {
```
gitlawr, line 274: The function name looks identical to pkgresource.SetSubjectID. But the behavior is different. Should we call this updateResourceSubjectID to differentiate?

alexcodelf: Updated, move it to pkg/resource.

gitlawr (approved):
LGTM

Issue #1909, opened 2024-01-16
Resource runtime configuration is null after cloning an environment


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create a resource and then clone the environment.
2. Check the deployment result.

**Result**

1. Failed to deploy.
<img width="1008" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/cde478c2-3689-4425-8f8c-392e1936dbc4">

2. Checked the runtime configuration.
<img width="982" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/2297ae8d-c9dc-4549-87af-796e81372bce">

3. Checked the custom configuration.
<img width="871" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/731aa185-ae42-4fec-97ba-1c8c15683ed4">

**Expected behavior**

**Environment**
- Walrus version: main 8ff16bfe4326036642753b1ba895fa5b0e3b1791
- UI version: dev-7590f46

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 3c468543a2cacb6c277c03f7d36bf2839e850cee
- UI version: v0.5.0-rc1

**Test Result:** pass

PR #1915, opened 2024-01-16
fix: missing computed attributes while clone environment

aiwantaozi:
**Problem:**
Missing computed attributes while cloning environment

**Solution:**
Generate computed attributes for resource while cloning environment

**Related Issue:**
#1909


shanewxy (dismissed):
LGTM

pkg/resource/computed_attributes.go
```diff
@@ -0,0 +1,183 @@
...
+	entity *model.Resource,
+	modelClient *model.Client,
+) (property.Values, error) {
+	// Check.
+	if !entity.TemplateID.Valid() && !entity.ResourceDefinitionMatchingRuleID.Valid() {
+		return nil, fmt.Errorf("failed to generate computed attributes, " +
+			"both template and resource definition machine rule id are empty")
+	}
+
+	if !entity.EnvironmentID.Valid() {
```
gitlawr, line 33: It panics if any ID is nil

aiwantaozi: updated.

pkg/resource/computed_attributes.go
```diff
@@ -0,0 +1,184 @@
...
+)
+
+// GenComputedAttributes generate computed attributes for resource.
+// Required:
+// entity.EnvironmentID, entity.Attributes
+// entity.TemplateID/entity.ResourceDefinitionMatchingRuleID.
+func GenComputedAttributes(
+	ctx context.Context,
+	entity *model.Resource,
+	modelClient *model.Client,
```
alexcodelf, line 25: we follow the pattern `func name(context.Context, model.ClientSet`, should we change the order of the params? The modelClient is `*model.Client`, will the function call pass a transaction model client?

alexcodelf (approved):
LGTM

Issue #1911, opened 2024-01-16
 Different users are not allowed to use the same name for API keys


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Log in walrus as admin user.
2. Create a API key.
3. Create a general user.
4. Log in walrus as general user and create a API key with the same name.

**Result**

<img width="802" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/7b7c9f09-a449-411d-81d3-fbe1209e4dd3">

**Expected behavior**

**Environment**
- Walrus version: main 8ff16bfe4326036642753b1ba895fa5b0e3b1791
- UI version: dev-7590f46

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 66f1bd1681cc926d6b35a6caed7ec5f793ae103a
- UI version: dev-21d34af

**Test Result:** pass

PR #1917, opened 2024-01-16
fix: different users cannot use the same name for api keys

shanewxy:
**Problem:**

Tokens name was set to be unique.

**Solution:**

Use subject_id and name unique index instead.

**Related Issue:**
#1911


aiwantaozi (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1910, opened 2024-01-16
Walrus file fail to deploy


shanewxy:
**Describe the bug**

**Steps to reproduce**
1. Export a previously successful deployed container service resource.
2. Import the exported yaml
```
version: v1
resources:
- name: nginx
  type: containerservice
  attributes:
    containers:
    - image: nginx
      ports:
      - external: 80
        internal: 80
        protocol: tcp
      profile: run
```
**Result**

<img width="679" alt="image" src="https://github.com/seal-io/walrus/assets/45061862/f7d747c0-bb73-4bff-b299-9cbd51dc1ef2">

**Expected behavior**

Successfully deployed.

**Environment**
- Walrus version: 1ba50028b1f2c30ea3155a195e856e78a3f0aab0

**Additional context**

gitlawr:
Deployment in the UI also fails

Finenyaco:
**Environment**

```
clientVersion:
  version: main
  gitCommit: 0f1cc5f1584deeec4cd890ef1a18f529294006d2
  isDevVersion: true
serverVersion:
  version: main
  gitCommit: 0f1cc5f1584deeec4cd890ef1a18f529294006d2
  isDevVersion: true
```

**Test Result:** 

- CLI:
-- with template: pass
-- with type: pass

```
 ~/seal/cli_yaml/ w apply -f test-deploy.yml --wait
resource default/local/nginx is creating
resource default/local/nginx is in status: Deploying
resource default/local/nginx is in status: Preparing
resource default/local/nginx is in status: Ready
 ~/seal/cli_yaml/ w delete -f test-deploy.yml --wait
resource default/local/nginx is deleting
resource default/local/nginx is in status: Deleting
resource default/local/nginx is deleted
```

- UI: pass

PR #1924, opened 2024-01-16
fix: resource id check cause panic

aiwantaozi:
**Problem:**
apply walrus file cause panic

**Solution:**
fix wrong precheck

**Related Issue:**
#1910

### DIFF
--- a/pkg/resource/computed_attributes.go
+++ b/pkg/resource/computed_attributes.go
@@ -26,9 +26,9 @@ func GenComputedAttributes(
 ) (property.Values, error) {
 	// Check.
 	if (entity.TemplateID == nil || !entity.TemplateID.Valid()) &&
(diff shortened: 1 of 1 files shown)


pkg/resource/computed_attributes.go
```diff
@@ -26,9 +26,9 @@ func GenComputedAttributes(
 ) (property.Values, error) {
 	// Check.
 	if (entity.TemplateID == nil || !entity.TemplateID.Valid()) &&
-		(entity.ResourceDefinitionID == nil || !entity.ResourceDefinitionMatchingRuleID.Valid()) {
```
alexcodelf, line 29: Is it acceptable that ResourceDefinitionID is nil?

aiwantaozi: ResourceDefinitionID isn't required for this function

thxCode (approved):
LGTM

gitlawr (approved):

Issue #1898, opened 2024-01-15
No default values for built-in resource definitions


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Go to resource definitions page.
2. View the built-in resource definition.

**Result**

1. No default values for built-in resource definition.
<img width="2553" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/de753a30-71b8-49d8-9aa1-41ebf9e4d80b">

**Expected behavior**

**Environment**
- Walrus version: main 8ff16bfe4326036642753b1ba895fa5b0e3b1791
- UI version: dev-567aeb4

**Additional context**

1. Create the same definition manually and default value are shown.
<img width="2553" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/19d47947-76ed-4a72-8fb1-9f202d113594">

aiwantaozi:
Can't reproduce in 8ff16bf

Finenyaco:
**Environment**
- Walrus version: main 0f1cc5f1584deeec4cd890ef1a18f529294006d2
- UI version: v0.5.0-rc1

**Test Result:** pass

PR #1927, opened 2024-01-16
fix: remove nested default values

thxCode:
**Problem:**

the default value of the nested structure has been cleaned, for example, the default of `a.b.c` has been reset, however, `a.b` doesn't have a default.

**Solution:**

make sure not to reset the child's values if the parent is empty.

**Related Issue:**
#1898


gitlawr (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #1922, opened 2024-01-16
Resource creation fails when the resource type corresponds to multiple resource definitions


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create two resource definitions with the same type.

<img width="2433" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/501e9ea6-3dac-4190-a8ed-5c7ec10ba17f">

2. Using the above resource type to create a resource and ensure that the conditions meet any one of the definitions.

**Result**

1. Failed to deploy: `find no matching resource definition`

<img width="2536" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/0aebe97b-85bf-4422-90a0-c79f631446fe">

**Expected behavior**

Create successfully.

**Environment**
- Walrus version: main 42f6b1e6cbcd105e9e99ea8badacace1a799c043
- UI version: dev-2fca30f

**Additional context**

Finenyaco:
Case 1: Create a definition with one rule and one selector with two items. Failed to match.
<img width="1179" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/c6e0f3cd-e54c-4cd3-a625-f05457661ffb">

Case 2: Create two definitions with the same type. 
- rule test-1: failed to match.
- rule test-2: created resource successfully.
<img width="2311" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/d6eeb0f2-9236-44c2-8bf9-9919c5e14b1c">
<img width="1101" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/ac6c2bab-55c5-42da-80e0-9e89ed37be92">
<img width="1111" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/20b565f4-1171-43f6-98d4-8fbf0412acba">

gitlawr:
The environment type selector is broken.
1. Project name selector with multiple items works.
2. Combination of other selectors works.

Finenyaco:
**Environment**
- Walrus version: main 66f1bd1681cc926d6b35a6caed7ec5f793ae103a
- UI version: dev-21d34af

**Test Result:** 
Environment type selector: pass

PR #1933, opened 2024-01-17
fix: remove env field select so that type is available for matches

gitlawr:
**Problem:**
env type selector is broken due to missing type in the query.

**Solution:**
Remove select.

**Related Issue:**
https://github.com/seal-io/walrus/issues/1922


alexcodelf (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #1564, opened 2023-11-24
Can't  create a environment when the name length is 63.


hibig:
**Describe the bug**

**Steps to reproduce**
1. Create a environment, enter a name with length is 63.
2. Save

**Result**

1. Response 500 error.

<img width="1530" alt="截屏2023-11-24 17 04 48" src="https://github.com/seal-io/walrus/assets/24664190/b1d7bb6d-ebd4-4651-9dc4-0dfb628c4ca2">

**Expected behavior**

1. Create a environment successfully.

**Environment**
- Walrus version: main

**Additional context**

shanewxy:
The error is from Kubernetes namespace creation, including project name + environment name. Can we dynamically limit the environment name length on UI?

thxCode:
@shanewxy , should we use https://developer.hashicorp.com/terraform/language/functions/substr to avoid this?

Finenyaco:
**Environment**
- Walrus version: main 66a2214c8ea5dfb7fd8542fecb5e87a3778fc07e
- UI version: dev-7449733

**Test Result:** 

Project, environment, and resource names are limited to 30 characters in length.

- API: pass
- UI: pass

PR #1934, opened 2024-01-17
fix: can't create environment when the name length is 63

shanewxy:
**Problem:**

Walrus creates a Kubernetes namespace `{projectName}-{environmentName}` when environment created. The namespace  should not be over 63 characters. 

The existing solution limits the environment name length on UI only, but the project name length is not taken into consideration.

**Solution:**

Trim the environment name to keep namespace name within 63 characters.

**Related Issue:**
#1564


thxCode:
how about if we limit the input of the project/environment name to 30 chars?

Issue #1628, opened 2023-11-27
Two services with dependencies are stuck.


hibig:
**Describe the bug**

**Steps to reproduce**
1. Create two services: service-a, service-b.
2. "service-b" depends on  "service-a".
3. Deploy them, then stop "service-b".
4. Back to environment list, toggle between the stop and start buttons multiple times, "service-a" and "service-b" status would be stuck.
<img width="1180" alt="截屏2023-11-26 20 04 31" src="https://github.com/seal-io/walrus/assets/24664190/2d69a0a7-6f01-4a89-88d9-d0edbebcc4b4">
<img width="1130" alt="截屏2023-11-26 20 04 51" src="https://github.com/seal-io/walrus/assets/24664190/c51637f9-4fa2-4444-80ab-bb21e3097dd0">

**Result**

**Expected behavior**

"service-a" and "service-b" status are normal.
**Environment**
- Walrus version: v0.4.0

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 66f1bd1681cc926d6b35a6caed7ec5f793ae103a
- UI version: dev-21d34af

**Test Result:**

It will turn to progressing failed status now. And need to redeploy it.

<img width="1566" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f631f00d-e60b-4aff-bcd0-d0fbc6c45cad">

PR #1937, opened 2024-01-17
fix: resource stuck in progressing

alexcodelf:
**Problem:**

Stopping and starting the environment can result in resources being perpetually in the Progressing running status. 

B &rarr; A

In the event where Resource B is waiting for Resource A, both will be stuck in the progressing status — Resource A is in the process of stopping, while Resource B is in deployment. Consequently, they will end up waiting for each other indefinitely, causing a lock in their statuses.

**Solution:**

Check dependecies before progressingly apply resource, dependencies to be delete will make resource deploy failed.
Check dependant before stop/delete resource, dependants to be deployed will make stop/delete failed.

**Related Issue:**
#1628

### DIFF
--- a/pkg/scheduler/resource/relationship_check_task.go
+++ b/pkg/scheduler/resource/relationship_check_task.go
@@ -155,6 +155,15 @@ func (in *RelationshipCheckTask) destroyResources(ctx context.Context) error {
 			continue
 		}
 
+		ok, err := in.checkDependants(ctx, res)
+		if err != nil {
+			return err
+		}
+
+		if !ok {
+			continue
+		}
+
 		err = pkgresource.Destroy(ctx, in.modelClient, res, pkgresource.Options{
 			Deployer: in.deployer,
 		})
@@ -185,6 +194,15 @@ func (in *RelationshipCheckTask) stopResources(ctx context.Context) error {
 
 	for _, res := range resources {
 		if !status.ResourceStatusStopped.IsUnknown(res) {
+			continue
+		}
+
+		ok, err := in.checkDependants(ctx, res)
+		if err != nil {
+			return err
+		}
+
+		if !ok {
 			continue
 		}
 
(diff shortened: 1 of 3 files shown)


gitlawr (dismissed):
LGTM

pkg/scheduler/resource/relationship_check_task.go
```diff
@@ -299,5 +341,56 @@ func (in *RelationshipCheckTask) setResourceStatusFalse(
...
+
 	return nil
 }
+
+// checkDependantResourceStatus sets a resource status to false if dependant resource status is false or deployed.
+func (in *RelationshipCheckTask) checkDependantResourceStatus(
+	ctx context.Context,
+	res, dependantResource *model.Resource,
+) (bool, error) {
+	if pkgresource.IsStatusFalse(dependantResource) {
```
aiwantaozi, line 366: IsStatusFalse is confusing, could you update to IsStatusFailed or IsStatusError

pkg/scheduler/resource/relationship_check_task.go
```diff
@@ -299,5 +341,56 @@ func (in *RelationshipCheckTask) setResourceStatusFalse(
...
+		status.ResourceStatusProgressing.False(
+			res,
+			fmt.Sprintf("Dependant resource status is false, name: %s", dependantResource.Name),
```
aiwantaozi, line 369: Status message may display in the UI, this message is confusing

pkg/scheduler/resource/relationship_check_task.go
alexcodelf: updated

pkg/scheduler/resource/relationship_check_task.go
alexcodelf: updated

aiwantaozi (approved):
LGTM

Issue #1932, opened 2024-01-17
Cannot access published endpoint in the local environment


gitlawr:
**Steps to reproduce**
1. Install Walrus: `docker run -d -p 80:80 -p 443:443 -p 30000-30100:30000-30100 --privileged sealio/walrus:v0.5.0-rc1`
2. Create a containerservice with nginx image, 80:80 port
3. Wait until the deployment succeeds and access the endpoint.

**Result**
The endpoint is inaccessible.

**Environment**
- Walrus version: v0.5.0-rc1

**Additional context**
It's using the internal IP of the walrus server container. It works when changing the IP to the Walrus server address.

Finenyaco:
**Environment**
- Walrus version: main 9a4516ef8771ed0bdce4acb990f084aa4e9a2a11
- UI version: dev-88faa21

**Test Result:** pass

<img width="1371" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f9963d0d-ae92-4c49-8f26-76b4e14516df">

<img width="2361" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/92da6319-3cd9-4a9c-9f89-49a7dfde5211">

PR #1941, opened 2024-01-17
feat: replace local ip endpoint with server's host

thxCode:
as #1932 mentioned, we cannot access an endpoint with the internal k3s node IP(a.k.a. the walrus container's IP).

- get non-loopback IPs at the beginning
- replace the endpoint's host/IP with the server's host if it matches the above IP set.

#1932

### DIFF
--- a/pkg/apis/resourcerevision/extension.go
+++ b/pkg/apis/resourcerevision/extension.go
@@ -6,6 +6,8 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
+	"net/url"
 	"time"
 
 	"k8s.io/apimachinery/pkg/util/sets"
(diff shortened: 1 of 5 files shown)


pkg/server/init_local_environment.go
```diff
@@ -138,7 +139,7 @@ func (r *Server) applyKubernetesConnector(
 			},
 		}
 
-		if os.Getenv("KUBERNETES_SERVICE_HOST") == "" && os.Getenv("_RUNNING_INSIDE_CONTAINER_") != "" {
+		if servervars.ServingEmbeddedKubernetes.Get() {
```
gitlawr, line 142: Previously, operator used the LabelEmbeddedKubernetes to determine the address of the discovered access endpoints. Looks like we can remove the label and related fields in operator since we no longer do endpoint discovery.

thxCode: updated

gitlawr (approved):
LGTM

pkg/apis/resourcerevision/extension.go
```diff
@@ -240,6 +244,26 @@ func manageResourceComponentsAndEndpoints(
 			}
 		}
 
+		// NB(thxCode): in order to access some endpoints pointed to local IP,
```
alexcodelf, line 247: The function is too long, can we extract endpoints logic to a new function?

Issue #1940, opened 2024-01-17
Incorrect  UI schema after resetting the template schema twice


Finenyaco:
**Describe the bug**

Reset template UI schema to default

**Steps to reproduce**
1. Add a template with source `https://github.com/unitong/test-variable`.
2. Go template details page -> UI Schema -> Reset to default.

https://github.com/seal-io/walrus/assets/26084756/b7e45b91-e962-47e8-a251-6dff4ed378cf

**Result**

1. After two resets, the UI schema is incorrect.

- Before
<img width="572" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/bc8ba6e6-f61e-4143-ac16-5e392e853a8b">

- After
<img width="546" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/46306341-bd21-45fc-8f31-d80ba559ea4a">

**Expected behavior**

**Environment**
- Walrus version: main a12b753f2e823aa7a7456110215dceed4da46ba1
- UI version: dev-07fa8db

**Additional context**

<img width="1370" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/c918ff44-678b-4f17-a191-fc4805e7c418">

<img width="1459" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/1511017f-ccb0-4aad-80a1-ddf1d9095838">

Finenyaco:
**Environment**
- Walrus version: main 66f1bd1681cc926d6b35a6caed7ec5f793ae103a
- UI version: dev-21d34af

**Test Result:** pass

PR #1944, opened 2024-01-17
fix: template version schema reset wrong

alexcodelf:
**Problem:**

Reset template version schema wrong. The reason is that git repo not reset to target version tag.

**Solution:**

Reset git repo to right version.

**Related Issue:**
#1940

### DIFF
--- a/pkg/templates/git.go
+++ b/pkg/templates/git.go
@@ -451,6 +451,10 @@ func syncTemplateVersion(ctx context.Context, mc model.ClientSet, tv *model.Temp
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


aiwantaozi (approved):
LGTM

pkg/templates/git.go
```diff
@@ -442,6 +442,8 @@ func syncTemplateVersion(ctx context.Context, mc model.ClientSet, tv *model.Temp
 		return err
 	}
 
+	repo.Reference = tv.Version
```
gitlawr, line 445: Why do we need this? Is ref included in `tv.Source` and get parsed to `repo.Reference`?

gitlawr (approved):
LGTM

Issue #1945, opened 2024-01-17
Resource labels selector does not take effect


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create a resource definition and add a resource label selector.

<img width="1362" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/ce59732e-0e59-402c-8b4d-ae5dae7da542">

2. Using the above definition to create a resource.

**Result**

1. Unable to select the definition with resource labels selector.

<img width="977" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/b20d6d3b-64f5-48b7-b796-95be79225c4b">

**Expected behavior**

**Environment**
- Walrus version: main a12b753f2e823aa7a7456110215dceed4da46ba1
- UI version: dev-07fa8db

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 50709d6d68b3f34858a63684418af0bfb19e7585
- UI version: dev-88faa21

**Test Result:** 

Failed to deploy during batch deployment.

<img width="2467" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f7e4bce7-0de9-4f4a-909c-5236bbc5de4d">


https://github.com/seal-io/walrus/assets/26084756/76996b17-40ed-44b5-a680-874ba0660c29

Finenyaco:
**Environment**
- Walrus version: main 66a2214c8ea5dfb7fd8542fecb5e87a3778fc07e
- UI version: dev-7449733

**Test Result:** pass

PR #1947, opened 2024-01-17
fix: env/res-defs missing defs with resource label selector

gitlawr:
**Problem:**
Available defs of the environment exclude all definitions with resource label selectors.

**Solution:**
Strip resource level selectors when matching at environment level.

**Related Issue:**
https://github.com/seal-io/walrus/issues/1945


shanewxy (approved):
LGTM

thxCode (approved):
LGTM

Issue #1923, opened 2024-01-16
Import yaml fails if the attributes field is not present


gitlawr:
Import the following yaml fails:
```
resources:
  - name: test
    type: postgresql
```

Import the following yaml succeeds:
```
resources:
  - name: test
    type: postgresql
    attributes: {}
```

**Environment**
- Walrus version: main

Finenyaco:
**Environment**
- Walrus version: main 8c258fa00a3e8c3f2895a5cfb7159ca2b5d7fc1f
- UI version: dev-88faa21

**Test Result:** pass

PR #1951, opened 2024-01-18
fix: empty attributes causes apply walrus file failed

aiwantaozi:
**Problem:**
lack of attributes causes apply walrus file failed

**Solution:**
check attributes empty before generate computed attributes

**Related Issue:**
#1923


gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #1935, opened 2024-01-17
Deploying resources with the same name and different types through walrus file will not return an error


Finenyaco:
**Describe the bug**

Deploying resources with the same name and different types through walrus file will not return an error

**Steps to reproduce**
1. Create a resource by walrus file as below:

```
resources:
  - name: db
    type: mysql
    attributes: {}
```

2. Create another resource with the same name, but the type is different.
```
version: v1
resources:
- name: db
  type: postgresql
  attributes:
    resources:
      cpu: 0.2
      memory: 1024
```

**Result**

1. The created resource has been upgraded, and the parameters have been modified. The default value for cpu is 0.25.
<img width="2388" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/a24c5dad-9b48-4db4-b800-bc5330a4b5e8">

**Expected behavior**

1. Prompt that type/template is different and no upgrade will be performed.

**Environment**
- Walrus version: main a12b753f2e823aa7a7456110215dceed4da46ba1
- UI version: dev-07fa8db

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 50709d6d68b3f34858a63684418af0bfb19e7585
- UI version: dev-88faa21

**Test Result:**

Upgrade the template version through walrus file returns an error: `invalid template: immutabl`.
<img width="1139" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/64d73366-f550-4bab-88d0-6170df5d2322">

Finenyaco:
**Environment**
- Walrus version: main 66a2214c8ea5dfb7fd8542fecb5e87a3778fc07e
- UI version: dev-7449733

**Test Result:** 

- Change type/template: pass

<img width="1133" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/814151b3-3c4b-4a36-bece-55b0b8e22715">

```
 ~/Downloads/ w apply -f default-k18-dev-test-web.yaml --wait
resource default/k18-dev/test-web apply failed
unexpected status code 400, invalid template name: immutable
```

- Upgrade template version: pass

PR #1962, opened 2024-01-18
fix: check immutable field to prevent update while patch

aiwantaozi:
**Problem:**
Deploying resources with the same name and different types through walrus file will not return an error

**Solution:**
Check immutable field while patch

**Related Issue:**
#1935


thxCode (approved):
LGTM

Issue #1946, opened 2024-01-17
Default values vary among different rules, but the definition schema still set the default values


Finenyaco:
**Describe the bug**

Default values vary among different rules, but the definition schema still set the default values

**Steps to reproduce**
1. New a resource definition.
2. Add different rules using template `kubernetes-containerservice` with different env values.

<img width="2435" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/49c3c668-67be-4ec5-8d97-0d36ccb1acf4">

3. Preview the resource definition UI schema.

**Result**

1. Set the value of 'Envs' in the last rule to the default value of definition's 'Envs'
<img width="2414" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/3f109ef1-1e80-4eb4-8ba4-366f26afb4fd">

**Expected behavior**

If the default values in the rules are different, the merged schema will not have a default value.

**Environment**
- Walrus version: main 92b986390ec6ebc79d72efde4762453084c236a1
- UI version: v0.5.0-rc1

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 474911154e649e2811e39f28c6eb2573ab0b9f92
- UI version: dev-7449733

**Test Result:** pass

PR #1963, opened 2024-01-18
fix: invalid defaults in res def

thxCode:
**problem**

this happens when the rules are configured with the same template but have different defaults.
since we take the defaults from a map indexed by the template ID before, we only get one default value at processing.

**solution**

getting defaults from the original matching rules.

**related**
#1946


aiwantaozi (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1854, opened 2024-01-10
Templates with duplicate names cause versions of different templates to be merged


linyinli:
**Describe the bug**
Templates with duplicate names cause versions of different templates to be merged.

**Steps to reproduce**
1. Add a catalog: https://github.com/alibabacloud-automation;
2. View the template: alicloud-rds-mysql, due to the same name: terraform-alicloud-rds-mysql, these versions of the template are merged into the existing template.

**Result**
<img width="1838" alt="image" src="https://github.com/seal-io/walrus/assets/42955482/39eae430-f0a0-46f2-9d60-702345bcb3e8">

**Expected behavior**
Split into two templates.

**Environment**
- Walrus version: 
  - UI: dev-437b352
  - Server: dev-71aeab4

Finenyaco:
Duplicate of https://github.com/seal-io/walrus/issues/1363

Finenyaco:
**Environment**
- Walrus version: main 66a2214c8ea5dfb7fd8542fecb5e87a3778fc07e
- UI version: dev-7449733

**Test Result:** pass

Added the catalog name as the template prefix.

<img width="1873" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/005882b1-c580-4133-ba04-aaa5b4bb70b1">

Issue #1363, opened 2023-11-07
Adding a catalog containing a template with the same name will overwrite the existing template


Finenyaco:
**Describe the bug**

Adding a catalog containing a template with the same name will overwrite the existing template.

**Steps to reproduce**
1. Add a catalog `https://github.com/unitong`. There is a template with the same name as the built-in template `webservice`.
2. Check the built-in template `webservice`.

**Result**

1. It overrode the built-in templates.
<img width="2499" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/db8e62f0-faa1-465a-8ee6-44499576eeb8">
<img width="2539" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/df1b3705-80ee-423d-be20-5e2cd4dff005">

2. After refreshing the built-in catalog.
<img width="2490" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/272f6dca-6845-4032-8d7a-9c63acbc421d">

**Expected behavior**

Does not affect the existing template.

**Environment**
- Walrus version: main 065ee92e55e14044fa0777be4c631f7082fadc0e
- UI version: dev-ef51150

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 66a2214c8ea5dfb7fd8542fecb5e87a3778fc07e
- UI version: dev-7449733

**Test Result:** pass

Added the catalog name as the template prefix.

<img width="1873" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/005882b1-c580-4133-ba04-aaa5b4bb70b1">

Issue #1340, opened 2023-10-31
Update naming of templates imported from a catalog


gitlawr:
**Is your enhancement related to a problem? Please describe.**
Improve readability of imported templates.

**Describe the solution you'd like**

For templates imported from a catalog, generate template name as `<catalog-name>/<template-name>`. Trim "terraform-" prefix if it is present in the repository name.

For example, 
`github.com/terraform-aws-modules/terraform-aws-rds` from catalog `aws-modules` results in `aws-modules/aws-rds`

Finenyaco:
**Environment**
- Walrus version: main 66a2214c8ea5dfb7fd8542fecb5e87a3778fc07e
- UI version: dev-7449733

**Test Result:** pass


<img width="2436" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/b9ca3d66-ed50-4127-9ca7-fa51fd6bed1a">
<img width="2412" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/e64c9607-6a28-475d-a495-0daf911b86bc">

PR #1964, opened 2024-01-18
fix: add catalog name to template

alexcodelf:
**Problem:**

Catalog with same template name can  not be imported.

**Solution:**

Add catalog name as prefix to its template.
image
image

**Related Issue:**
#1854 #1363 #1340

### DIFF
--- a/pkg/catalog/catalog.go
+++ b/pkg/catalog/catalog.go
@@ -22,6 +22,7 @@ import (
 	"github.com/seal-io/walrus/pkg/vcs/options"
 	"github.com/seal-io/walrus/utils/gopool"
 	"github.com/seal-io/walrus/utils/log"
+	"github.com/seal-io/walrus/utils/strs"
 	"github.com/seal-io/walrus/utils/version"
 )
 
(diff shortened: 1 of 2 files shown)


pkg/apis/catalog/basic_view.go
```diff
@@ -24,6 +25,14 @@ func (r *CreateRequest) Validate() error {
 		return err
 	}
 
+	if err := validation.IsDNSLabel(r.Name); err != nil {
+		return fmt.Errorf("invalid name: %w", err)
+	}
+
+	if len(r.Name) > 30 {
+		return fmt.Errorf("name %q is too long, must be less than 30 characters", r.Name)
+	}
```
gitlawr, line 34: What happen when it exceeds the length?

alexcodelf: When catalog name is  too long, it is not easy for users to choose template to create resource or check info of templates. The name is not related to resource deployment, we don't need to worry about it.

alexcodelf: Let's handle the name limit in other Pull Request.

gitlawr (approved):
LGTM

thxCode (approved):
LGTM

PR #1968, opened 2024-01-19
fix: can't create environment when the name length is 63

shanewxy:
**Problem:**

Walrus creates a Kubernetes namespace {projectName}-{environmentName} when environment created. The namespace should not be over 63 characters.

**Solution:**

Limit all object names within 30 characters, except for templates(considering lengthy ones like [terraform-aws-managed-service-prometheus](https://github.com/terraform-aws-modules/terraform-aws-managed-service-prometheus).

**Related Issue:**
#1564


gitlawr (approved):
LGTM

thxCode (approved):
LGTM

Issue #1931, opened 2024-01-16
Sometimes upgrade resource fails: invalid variables: violate ui schema: invalid password: minimum string length is 16


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create a resource using built-in `redis` type.
2. Upgrade the resource.

**Result**

1. Sometimes upgrade fails and prompt: `invalid variables: violate ui schema: invalid password: minimum string length is 16`.

<img width="2469" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/ea932fcf-dd11-4063-8c01-785c6d9a2111">

<img width="2471" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/07a19bae-3215-48cb-a654-2a82c94f2027">

**Expected behavior**

**Environment**
- Walrus version: main 0f1cc5f1584deeec4cd890ef1a18f529294006d2
- UI version: v0.5.0-rc1

**Additional context**

aiwantaozi:
Could not reproduce with 192fc6469767ae105ce2662048495585d12e5114

Finenyaco:
**Environment**
- Walrus version: main 66f1bd1681cc926d6b35a6caed7ec5f793ae103a
- UI version: dev-88faa21

**Test Result:** This issue can be reproduced when the password is set as a referenced variable.
<img width="1238" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/c730278a-39dc-4bd5-a5e8-7baceb323304">

<img width="2560" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/abcdf000-555f-4289-a4ea-086299d03849">

Finenyaco:
**Environment**
- Walrus version: main 73e6ce88f4205e030d30ff4276e82d687c0ccdd0
- UI version: dev-860ae96

**Test Result:** pass

PR #1969, opened 2024-01-19
refactor: inject var.xx and res.xx.output attribute before validate

aiwantaozi:
**Problem:**
attributes include ${var.xx} or ${res.xx.output} may failed to validate

**Solution:**
inject attributes before validating with schema

**Related Issue:**
#1931

### DIFF
--- a/pkg/apis/resource/helper.go
+++ b/pkg/apis/resource/helper.go
@@ -1,10 +1,181 @@
 package resource
 
 import (
+	"bytes"
+	"context"
+	"fmt"
+	"strings"
+
+	"k8s.io/apimachinery/pkg/util/sets"
+
 	"github.com/seal-io/walrus/pkg/dao/model"
+	"github.com/seal-io/walrus/pkg/dao/model/resource"
 	"github.com/seal-io/walrus/pkg/dao/types/object"
 	"github.com/seal-io/walrus/pkg/dao/types/property"
+	pkgresource "github.com/seal-io/walrus/pkg/resource"
+	pkgvariable "github.com/seal-io/walrus/pkg/variable"
+	"github.com/seal-io/walrus/utils/json"
 )
+
+var (
+	// VariableFormat the format of variable.
+	variableFormat = `"${var.%s}"`
+	// ResourceFormat the format of resource output.
+	resourceFormat = `"${res.%s.%s}"`
+)
+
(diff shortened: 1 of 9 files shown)


pkg/apis/resource/helper.go
```diff
@@ -1,11 +1,187 @@
...
+
+		n := fmt.Sprintf(resourceFormat, ss[0], ss[1])
+		values[n] = v.Value
+	}
+
+	return values, nil
+}
+
+func getShouldInject(
+	bs []byte,
```
gitlawr, line 154: use readable param name

pkg/apis/resource/helper.go
```diff
@@ -1,11 +1,187 @@
...
+			resource.NameIn(resNames...),
+			resource.ProjectID(projectID),
+			resource.EnvironmentID(environmentID)).
+		IDs(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Get should inject resource outputs.
+	outputs, err := terraform.GetServiceDependencyOutputs(ctx, client, depIDs, resOutputs)
```
gitlawr, line 128: Calling deployer/terraform pkg in apis/resource seems to be odd. Should we move the function to pkg/resource(and change its name)?

gitlawr (dismissed):
LGTM

pkg/apis/resource/helper.go
```diff
@@ -1,11 +1,188 @@
...
 	"github.com/seal-io/walrus/pkg/dao/types/property"
+	pkgresource "github.com/seal-io/walrus/pkg/resource"
+	pkgvariable "github.com/seal-io/walrus/pkg/variable"
+	"github.com/seal-io/walrus/utils/json"
 )
 
+var (
+	// _variableReg the regexp to match the variable.
+	variableReg = regexp.MustCompile(`\${var\.([a-zA-Z0-9_-]+)}`)
+	// _resourceReg the regexp to match the resource output.
```
alexcodelf, line 24: There are two regexp `\${res\.([^.}]+)\.([^.}]+)` and `\${var\.([a-zA-Z0-9_-]+)}` in this package and `pkg/deployer/terraform` can we unify them? Please update the comment of the variable.

alexcodelf (approved):
LGTM

Issue #1957, opened 2024-01-18
Incorrect workflow status when a step status changes to failed


Finenyaco:
**Describe the bug**

When there are multiple steps within a stage, and one of the steps fails, the workflow status does not change to 'Failed'.

**Steps to reproduce**
1. New a workflow and add two steps within a stage.
2. Execute the workflow.
3. Make one of the steps fail.
4. Check the workflow status.

**Result**

1. The workflow remains in the "running" state.
<img width="753" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/e55ad4c4-2d76-4f0d-9657-eecb1838ed00">

<img width="2405" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/4072d483-145a-4e75-b13e-44f1c962d91e">

**Expected behavior**

The workflow status should be changed to "failed."

**Environment**
- Walrus version: main 66f1bd1681cc926d6b35a6caed7ec5f793ae103a
- UI version: dev-88faa21

**Additional context**

Issue #1956, opened 2024-01-18
Failed to upgrade resources by workflow


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. New a workflow and add a 'Deploy Resource' step.
2. Execute the workflow.
3. Execute the workflow again.

**Result**

1. In step 2, resource was deployed successfully.

<img width="807" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/9911359f-aabe-454c-8ea8-005cea2217f9">

2. In step 3, failed to upgrade the resource.

<img width="1029" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/ee26c009-a8b6-461e-8118-f8d59f231534">

**Expected behavior**

Upgrade successfully.

**Environment**
- Walrus version: main 66f1bd1681cc926d6b35a6caed7ec5f793ae103a
- UI version: dev-88faa21

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 918a699c606f64575b1ac86f58c93ae899c141c0
- UI version: dev-8062a50

**Test Result:** pass

Issue #1955, opened 2024-01-18
Incorrect workflow step status when stopping the step in the pending state


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. New a workflow and a manual approval.
2. Execute the workflow and then stop it immediately when the  'manual approval' step in pending state.

**Result**

1. The workflow status changed to canceled but the step status did not change to canceled.
<img width="2459" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/10ea79bf-f383-44b2-998c-52480477aa99">

<img width="892" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/38261098-1ab3-4abb-a817-55a127a13f2b">

2. Approve the step. The workflow status changed to failed from canceled.
<img width="1407" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/da283363-435b-4c78-b21e-ff28cfa1d904">

<img width="1216" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/06701ceb-58e2-4048-9298-a2d0ef1fcf65">

**Expected behavior**

**Environment**
- Walrus version: main 66f1bd1681cc926d6b35a6caed7ec5f793ae103a
- UI version: dev-88faa21

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 918a699c606f64575b1ac86f58c93ae899c141c0
- UI version: dev-8062a50

**Test Result:** Failed


https://github.com/seal-io/walrus/assets/26084756/eb2cf501-a735-4490-bdfa-e54dcedb2dbd

alexcodelf:
The reason is that argo won't stop pending task. It may caused with the same reason https://github.com/argoproj/argo-workflows/pull/12196.

A workflow is termniated, but pending pods will not be deleted. After a while, the pending pod is running.


![Image](https://github.com/seal-io/walrus/assets/85595852/b0e02085-e5ad-4d55-b37f-c240f39ec014)


Let's figure a better solution in v0.6.0

alexcodelf:
As we plan to import Kubernetes CRDs, we may hold off on addressing this issue until later. Remove from 0.6.0

PR #1973, opened 2024-01-19
fix: workflow status update and resource step can not upgrade 

alexcodelf:
**Problem:**

- can not upgrade resource in workflow
- a pending step will not change status when cancelling workflow.
- workflow execution status will not changed to error when sub step failed.

**Solution:**

- add resource definition matching rules for resource step.
- omitted step will update step status
- fail workflow fast when step failed

**Related Issue:**
#1955 #1956 #1957

### DIFF
--- a/pkg/apis/resource/extension_view.go
+++ b/pkg/apis/resource/extension_view.go
@@ -77,26 +77,9 @@ func (r *RouteUpgradeRequest) Validate() error {
 		}
 	}
 
-	if r.ReuseAttributes {
-		r.Attributes = entity.Attributes
-		r.ComputedAttributes = entity.ComputedAttributes
-
-		if r.Labels == nil {
-			r.Labels = entity.Labels
-		}
-	} else {
-		en := r.Model()
-
-		// Set environment ID since Model will not set the environment ID.
-		en.EnvironmentID = r.Environment.ID
-
-		computedAttr, err := pkgresource.GenComputedAttributes(r.Context, r.Client, en)
-		if err != nil {
-			return err
-		}
-
-		r.ComputedAttributes = computedAttr
-	}
+	en := r.Model()
+	// Set environment ID since Model will not set the environment ID.
+	en.EnvironmentID = r.Environment.ID
 
 	switch {
 	case r.Template != nil:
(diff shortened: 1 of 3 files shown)


pkg/workflow/step/service_step.go
```diff
@@ -199,3 +201,35 @@ func (s *ServiceStepManager) GenerateTemplates(
...
+		Where(resource.Name(name)).
+		WithResourceDefinitionMatchingRule().
+		Only(ctx)
+	if err != nil && !model.IsNotFound(err) {
+		return nil, err
+	}
+
+	if res != nil && res.Edges.ResourceDefinitionMatchingRule != nil {
+		stepAttrs["resourceDefinitionMatchingRule"] = map[string]any{
+			"id": res.Edges.ResourceDefinitionMatchingRule.ID.String(),
```
gitlawr, line 230: How do we use this?

alexcodelf: Resource validation need `ResourceDefinitionMatchingRuleID` or `templateID` to generate computed attributes, will need add this for resource upgrade API.

pkg/apis/resource/extension_view.go
```diff
@@ -530,15 +533,10 @@ func (r *CollectionRouteUpgradeRequest) Validate() error {
...
-		} else {
-			en := updateInputsItemToResource(entity.Type, input, r.Project, r.Environment)
-
-			input.ComputedAttributes, err = pkgresource.GenComputedAttributes(r.Context, r.Client, en)
-			if err != nil {
-				return err
-			}
 		}
 
+		en := updateInputsItemToResource(entity.Type, input, r.Project, r.Environment)
```
gitlawr, line 538: We don't call this if not ReuseAttributes before. Is this expected change?

alexcodelf: This variable Is only used in computed resources, the func will new `*model.Resource`, the input data will not be effected.

gitlawr (approved):
LGTM

pkg/apis/resource/extension_view.go
```diff
@@ -176,10 +168,21 @@ func (r *RouteUpgradeRequest) Validate() error {
...
 
+	if !r.ReuseAttributes {
+		computedAttr, err := pkgresource.GenComputedAttributes(r.Context, r.Client, en)
+		if err != nil {
+			return err
+		}
+
+		r.ComputedAttributes = computedAttr
+	}
+
```
thxCode, line 185: we should make `if-else` look like a whole logic, please put the above `if` block here too.

```
if r.ReuseAttributes {
		r.Attributes = entity.Attributes
		r.ComputedAttributes = entity.ComputedAttributes
		if r.Labels == nil {
			r.Labels = entity.Labels
		}
} else {
                computedAttr, err := pkgresource.GenComputedAttributes(r.Context, r.Client, en)
		if err != nil {
			return err
		}

		r.ComputedAttributes = computedAttr
}
```

thxCode (approved):
LGTM

Issue #1928, opened 2024-01-16
Provide a Walrus file for Walrus


gitlawr:
Add a walrus file in https://github.com/seal-io/walrus/tree/main/deploy/manifests

Finenyaco:
**Environment**
- Walrus version: main 98e5445080de1e1017db16951dd12d8e1e52b27f
- UI version: dev-3fcbabe

**Test Result:**

In the Walrus file, the `profile` must be configured. Otherwise, the configured ports cannot be viewed on the UI.

<img width="2560" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/e48b3836-a55d-42a2-afac-6fb8c8f2494e">

```
resources:
  - name: walrus-casdoor
    type: containerservice
    attributes:
      containers:
        - image: sealio/casdoor:v1.344.0-seal.1
          profile: run             <<<<
          ports:
            - internal: 8000
              external: 8000
 - name: walrus
    type: containerservice
    attributes:
      containers:
        - image: sealio/walrus:main
          profile: run             <<<<
          ports:
            - internal: 80
              external: 80
            - internal: 443
              external: 443
```

Finenyaco:
Verified

PR #1974, opened 2024-01-19
fix: walrus file profile attribute

shanewxy:
**Problem:**

Without profile configured, the configured ports cannot be viewed on the UI. And the ports will be removed when upgrade on UI.

**Solution:**

Fill in the default profile `run`.

**Related Issue:**
https://github.com/seal-io/walrus/issues/1928#issuecomment-1900065799

### DIFF
--- a/deploy/manifests/walrus-file.yaml
+++ b/deploy/manifests/walrus-file.yaml
@@ -13,6 +13,7 @@ resources:
     attributes:
       containers:
         - image: sealio/casdoor:v1.344.0-seal.1
+          profile: run
           ports:
             - internal: 8000
               external: 8000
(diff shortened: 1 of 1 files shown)


deploy/manifests/walrus-file.yaml
```diff
@@ -13,6 +13,7 @@ resources:
     attributes:
       containers:
         - image: sealio/casdoor:v1.344.0-seal.1
+          profile: run
```
gitlawr, line 16: Isn't it the default? Should we fix on template side?

shanewxy: It is the default in template, but looks like frontend only displays attributes filled in during resource creation. Can @hibig check if we could display schema default value on UI besides the filled in attributes?

thxCode (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1960, opened 2024-01-18
cli decrease the timeout for server reachable check


aiwantaozi:
**Is your enhancement related to a problem? Please describe.**
Now CLI configured with an unreachable server takes a long time to timeout. 
Need to decrease the timeout for server reachable check.

Finenyaco:
**Environment**
- CLI version: 73e6ce88f4205e030d30ff4276e82d687c0ccdd0

**Test Result:**
When unable to connect to the server, about 20-second waiting period is required.
```
 ~/.walrus/ time walrus version
access https://walrus.tsfxwx.com failed
walrus version  0.05s user 0.03s system 0% cpu 20.202 total
```

Finenyaco:
**Environment**
- Walrus version: main 4eef11db3efc80ce93bb17814e5e8fb822d75d45
- UI version: dev-8ceaed8

**Test Result:** 

The current response takes 10 seconds. The timeout duration can be further reduced.

<img width="544" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/e321142a-29ba-4de3-8909-7d31581b6dce">

aiwantaozi:
This is caused by the office network(proxy).

Under the office network, could access the unreachable network.
![Image](https://github.com/seal-io/walrus/assets/5936271/90f49920-dc05-42b1-96d2-513f4c2b81ab)

After switching to the mobile network, couldn't access it, and the walrus cli response time is around 3 second
![Image](https://github.com/seal-io/walrus/assets/5936271/0cf205b0-b14d-441c-b108-6f634ec81df3)

Finenyaco:
**Environment**
- Walrus version: main 4eef11db3efc80ce93bb17814e5e8fb822d75d45
- UI version: dev-94aee09

**Test Result:** 

- When the domain name cannot be resolved
```
 ~/seal/ time w version
clientVersion:
  version: main
  gitCommit: 4c1b4a42d2281ef308519aed71c0dac46cff6c87
  isDevVersion: true
walrus version  0.04s user 0.01s system 50% cpu 0.106 total
```

- When the domain name can be resolved but cannot be connected
```
 ~/seal/ time w version
clientVersion:
  version: main
  gitCommit: 4c1b4a42d2281ef308519aed71c0dac46cff6c87
  isDevVersion: true
walrus version  0.04s user 0.01s system 1% cpu 3.058 total
```

- When the address can be connected, but the Walrus service is inaccessible.
```
[...]

PR #1975, opened 2024-01-19
refactor: decrease the cli request timeout

aiwantaozi:
**Problem:**
while the server isn't reachable, cli takes a long time to timeout  

**Solution:**
decrease the timeout for server reachable check

**Related Issue:**
#1960


shanewxy (dismissed):
LGTM

pkg/cli/config/config.go
```diff
@@ -196,9 +202,9 @@ func (c *Config) ServerVersion() (*Version, error) {
 		return nil, err
 	}
 
-	resp, err := c.DoRequest(req)
+	resp, err := c.DoRequestWithTimeout(req, fastTimeout)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("access %s failed", c.Server)
```
gitlawr, line 207: Do we need to print original error

pkg/cli/config/config.go
```diff
@@ -18,7 +18,8 @@ import (
 )
 
 const (
-	timeout     = 15 * time.Second
+	fastTimeout = 1 * time.Second
```
thxCode, line 21: what means `fastTimeout`, can we use shortTimeout?

pkg/cli/config/config.go
aiwantaozi: No need, while use --debug flag, will log the response, and can check the failed reason by response body.

pkg/cli/config/config.go
```diff
@@ -18,10 +18,11 @@ import (
 )
 
 const (
-	timeout     = 15 * time.Second
-	openAPIPath = "/openapi"
-	apiVersion  = "v1"
-	versionPath = "/debug/version"
+	shortTimeout = 1 * time.Second
+	timeout      = 5 * time.Second
```
thxCode, line 22: @aiwantaozi  the `client.Timeout` controls the request from dial to response header reading, I think 1s is too short for some bad networking connection cases.

we should configure more granular timeouts, here is an example.
```
&http.Client{
    Transport: &http.Transport{
        Proxy: http.ProxyFromEnvironment,
        DialContext: (&net.Dialer{
            Timeout:   1 * time.Second,           // shorten this to make sure if we can connect to server
            KeepAlive: 30 * time.Second,  
            DualStack: true,
        }).DialContext,
        TLSHandshakeTimeout:   10 * time.Second,  // server must finish cert generation if using self-signed CA
        ResponseHeaderTimeout: 30 * time.Second,  // server must complete the request and start to transit the response 
    }
}
```

aiwantaozi: updated

thxCode (approved):
LGTM

PR #1977, opened 2024-01-19
fix: show stop label in prod

alexcodelf:
**Problem:**

Production env resources do not show stop label. As https://github.com/seal-io/walrus/issues/1765#issuecomment-1899547756 introduced, production env should show stop label to users.

**Solution:**

Add stop label false in prod env.

**Related Issue:**
https://github.com/seal-io/walrus/issues/1765#issuecomment-1899547756

### DIFF
--- a/pkg/resource/resource.go
+++ b/pkg/resource/resource.go
@@ -601,6 +601,8 @@ func SetEnvResourceDefaultLabels(env *model.Environment, r *model.Resource) erro
 	// Dev and staging environments resources are stoppable by default.
 	case types.EnvironmentDevelopment, types.EnvironmentStaging:
 		r.Labels[types.LabelResourceStoppable] = "true"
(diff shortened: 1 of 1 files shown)


pkg/resource/resource.go
```diff
@@ -602,6 +602,7 @@ func SetEnvResourceDefaultLabels(env *model.Environment, r *model.Resource) erro
 	case types.EnvironmentDevelopment, types.EnvironmentStaging:
 		r.Labels[types.LabelResourceStoppable] = "true"
 	default:
```
thxCode, line 604: i prefer to use `case types.EnvirontmnetProduction` rather than `default`, even if we know the last one is prd at present.

thxCode (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1972, opened 2024-01-19
bug: failed to deploy walrus on k3s using embedded cluster


thxCode:
**Describe the bug**

failed to launch walrus if the host ip conflicted with the embedded k3s cluster cidr.

**Steps to reproduce**
1. quick start a k3s cluster on VM, https://docs.k3s.io/quick-start.
2. launch walrus with `--kubeconfig=<point to the kubeconfig generated by step 1>`.
3. deploy walrus Walrus File from UI.

**Result**

the walrus deployed by Walrus File keeps in restarting.

**Expected behavior**

make it work.

**Environment**
- Walrus version: main

**Additional context**

set up a k3s cluster with multipass.

``` bash
# start vm
multipass launch --cpus 6 --memory 12G --disk 100G --name k3s

# install k3s
multipass shell k3s
curl -sfL https://get.k3s.io | sh -s - server --write-kubeconfig-mode="0644" --token=k3s:test123
exit

# export k3s kubeconfig
multipass exec k3s -- sudo cat /etc/rancher/k3s/k3s.yaml > ~/.kube/k3s.yaml
sed -i "s#127.0.0.1#<k3s vm IP, get by multipass list>#g" $(pwd)/.kube/k3s.yaml
```

Finenyaco:
**Environment**
- Walrus version: main 141312005b8adf7d63ed15b16b95271072d39577
- UI version: dev-8062a50

**Test Result:**
- deploy walrus on k3s: pass
- deploy walrus on embedded k3s: pass

PR #1978, opened 2024-01-19
fix: failed to start in k3s cluster

thxCode:
support Russian Dolls

**problem**

1. the embedded k3s IP conflicts with the on-host cluster
2. the nesting embedded k3s keeps using root cgroup, which causes launching to fail. 

**solution**

1. when it starts k3s, walrus tries to pick a proper subnet according to the host's subnet.
2. when it starts k3, walrus picks a proper cgroup.

**related**

#1972

### DIFF
--- a/pkg/server/runner.go
+++ b/pkg/server/runner.go
@@ -19,6 +19,8 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/urfave/cli/v2"
 	"k8s.io/apimachinery/pkg/runtime"
+	utilnet "k8s.io/apimachinery/pkg/util/net"
+	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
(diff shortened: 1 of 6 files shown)


aiwantaozi (approved):
LGTM

pkg/server/runner.go
```diff
@@ -731,6 +734,49 @@ func (r *Server) configure(_ context.Context) error {
...
+
+			addrs, err := _if.Addrs()
+			if err != nil {
+				return fmt.Errorf("error querying addresses from interface %s(%d): %w",
+					_if.Name, _if.Index, err)
+			}
+
+			for _, addr := range addrs {
+				if ipnet, ok := addr.(*net.IPNet); ok && !ipnet.IP.IsLoopback() {
+					// NB(thxCode): There is no NAT for IPv6,
```
alexcodelf, line 760: Will it run in a pure IPv6 environment?

thxCode: there is no "private" ipv6 address.

alexcodelf (approved):
LGTM

Issue #1970, opened 2024-01-19
Failed to deploy docker postgresql  resource using the built-in resource definition


Finenyaco:
**Describe the bug**

Failed to deploy docker postgresql using the built-in resource definition `postgresql`.

**Steps to reproduce**
1. Setup Walrus with docker mode.
2. Create a posgresql resource using the built-in resource definition `postgresql`.
3. Check the deployment result.

**Result**

1. Failed to deploy.
<img width="1046" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/228922dc-2b6b-46a4-adfa-3556325dfecc">

2. The engine version is incorrect.
<img width="2404" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/e5a78381-53b4-4905-8a02-c690f12f3fdb">

**Expected behavior**

Deploy successfully.

**Environment**
- Walrus version: main 98e5445080de1e1017db16951dd12d8e1e52b27f
- UI version: dev-3fcbabe

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 98e5445080de1e1017db16951dd12d8e1e52b27f
- UI version: dev-3fcbabe

**Test Result:**

Similar issues exist in other fields. For example, in the `container` definition, the 'Accept Changed' attribute in Kubernetes was originally modifiable, but now it is not.

<img width="1262" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/e7b3d54b-2a16-4ea1-9dc1-f869aba3a218">

shanewxy:
> **Environment**
> 
> * Walrus version: main [98e5445](https://github.com/seal-io/walrus/commit/98e5445080de1e1017db16951dd12d8e1e52b27f)
> * UI version: dev-3fcbabe
> 
> **Test Result:**
> 
> Similar issues exist in other fields. For example, in the `container` definition, the 'Accept Changed' attribute in Kubernetes was originally modifiable, but now it is not.
> 
[...]

PR #1979, opened 2024-01-20
fix: hasn't merged readonly in res def

thxCode:
**Problem:**

the generated UI schema doesn't process the read-only attribute.

**Solution:**

merge read-only between multiple templates' schemas.

**Related Issue:**

#1970


pkg/resourcedefinitions/helper.go
```diff
@@ -261,6 +265,15 @@ func alignSchemaRef(nb map[string]any, key string, lsr, rsr *openapi3.SchemaRef)
 				if period+1 < len(d) {
 					lv.Description = string(d[:period+1])
 				}
+
+				switch {
+				case lv.ReadOnly && lv.WriteOnly:
+					lv.Description += " (read-only, write-only)"
```
gitlawr, line 271: What does read-only + write-only mean?

thxCode: that is what we should point out, if the res def gets this, the maintainer should watch out for this.

thxCode: replace with (restricted) or something else?

gitlawr (approved):
LGTM

pkg/resourcedefinitions/helper.go
```diff
@@ -230,6 +230,10 @@ func alignSchemaRef(nb map[string]any, key string, lsr, rsr *openapi3.SchemaRef)
 		lv.WriteOnly = rv.WriteOnly
 	}
 
+	if !lv.ReadOnly && rv.ReadOnly {
+		lv.ReadOnly = rv.ReadOnly
+	}
+
```
shanewxy, line 236: Do we need to merge `immutable` attributes as well?

thxCode: https://github.com/OAI/OpenAPI-Specification/issues/2720 looks like there is no `immutable` in OpenAPI at present

shanewxy (approved):
LGTM

Issue #1618, opened 2023-11-26
Update schema to invalid format shows success


shanewxy:
**Describe the bug**

**Steps to reproduce**
1. Create a template.
2. Edit the schema to invalid format.
3. Click save.

**Result**

Shows success, but actually not updated after refresh.

https://github.com/seal-io/walrus/assets/45061862/15d08fe9-bd6c-4c77-a627-96db735a33c8

**Expected behavior**

Show error when schema input is invalid.

**Environment**
- Walrus version: 501535b858f4bbfb681ab45af9ca332b96c776aa

**Additional context**

aiwantaozi:
1. remove the components part from the Yaml
2. find UI still send to the API
<img width="1427" alt="image" src="https://github.com/seal-io/walrus/assets/5936271/d7f67cc4-82d2-4fe1-9739-48ee4d0a5a9a">

Finenyaco:
**Environment**
- Walrus version: main 8c258fa00a3e8c3f2895a5cfb7159ca2b5d7fc1f
- UI version: dev-88faa21

**Test Result:**

- Template UI schema: delete the components will return an unexpected schema.

https://github.com/seal-io/walrus/assets/26084756/3fd8fea0-42a2-40ca-8f3d-e1be1cd05bb0

- Resource definition UI schema: deletion of components prompts successful, but in reality, it has not been removed.

https://github.com/seal-io/walrus/assets/26084756/cc332618-2762-4170-9f15-c3ef8cca2181

Finenyaco:
**Environment**
- Walrus version: main 0b2678c0df661cd20cafbdb716a641717f158e7f
- UI version: v0.5.0-rc2

**Test Result:** pass

PR #1988, opened 2024-01-22
fix: return ui schema even it without components/variables

aiwantaozi:
**Problem:**
Now will expose server-generated schema while ui schema doesn't include components/variables

**Solution:**
return ui schema even it without components/variables

**Related Issue:**
#1618


gitlawr (approved):
LGTM

thxCode (approved):
LGTM

Issue #1966, opened 2024-01-19
Failed to deploy resources through Walrus file in the server run by Kubernetes


Finenyaco:
**Describe the bug**

Deployment of resources through the Walrus file in the HA environment failed.

**Steps to reproduce**
1. Setup a HA Walrus servce.
2. Create a nginx resource by importing a walrus file.

**Result**

1. Failed to create: `failed to apply walrus file: Get \"https://localhost/v1/projects/default/environments/local/resources/nginx\": dial tcp [::1]:443: connect: connection refused`
<img width="2467" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/2a3dd555-78ad-4d75-b8c8-8ee77360a998">

2. View logs.
```
2024-01-19T02:51:16.431Z	E	api	error requesting POST /v1/projects/496778235128840202/environments/496778235162394634/apply: 1 error occurred:
	* failed to apply walrus file: Get "https://localhost/v1/projects/default/environments/local/resources/nginx": dial tcp [::1]:443: connect: connection refused


github.com/seal-io/walrus/pkg/apis/runtime.erroring
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_error.go:45
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.recovering
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_recovery.go:39
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.observing
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_observation.go:89
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/gin-gonic/gin.(*Engine).handleHTTPRequest
	github.com/gin-gonic/gin@v1.9.1/gin.go:620
github.com/gin-gonic/gin.(*Engine).ServeHTTP
	github.com/gin-gonic/gin@v1.9.1/gin.go:576
github.com/seal-io/walrus/pkg/apis/runtime.(*Router).ServeHTTP
	github.com/seal-io/walrus/pkg/apis/runtime/router.go:193
net/http.serverHandler.ServeHTTP
	net/http/server.go:2947
net/http.(*conn).serve
[...]

PR #1989, opened 2024-01-22
fix: use proper servercontext to apply by tls info

gitlawr:
**Problem:**
Fail to apply in HA

**Solution:**
Use http server context when tls is disabled.

**Related Issue:**
https://github.com/seal-io/walrus/issues/1966


aiwantaozi (approved):
LGTM

thxCode (approved):
LGTM

Issue #1984, opened 2024-01-22
Incorrect workflow step status after the retry


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. New a workflow and add a 'Deploy Resource' step.
2. Set `Retry Times` to 1.
<img width="1193" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/3596f936-e54e-4ef5-990b-04e0c30ef5db">

3.  Reference a non-existent resource output.
4. Run the workflow.
5. Check the step status.

**Result**

1. The workflow step stuck in "Running" state.

https://github.com/seal-io/walrus/assets/26084756/ac31d93c-4bd3-4e46-b07b-e0ab1a7022e5

**Expected behavior**

The workflow step status should change to "Failed".

**Environment**
- Walrus version: main 918a699c606f64575b1ac86f58c93ae899c141c0
- UI version: dev-8062a50:

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 3806cbeb980963a23e83bf42b8ac3b15c287fbba
- UI version: v0.5.0-rc2

**Test Result:** pass

<img width="1012" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/1bdda03f-2fcd-4849-b1e5-a2d9cf0ac6cc">

PR #1994, opened 2024-01-22
fix: retry failed status not updated

alexcodelf:
**Problem:**

Step with retry strategy status not update when failed.

**Solution:**

Should not fail fast for step template.

**Related Issue:**
#1984


shanewxy (approved):
LGTM

thxCode (approved):
LGTM

Issue #1938, opened 2024-01-17
Incorrect computed attribute for variable optional value


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Add a template with source `https://github.com/unitong/test-variable`.
The definition of test_object variable:
```
variable "test_object" {
  type          = object({
    name        = string
    age         = number
    email       = optional(object({
        address = string
        domain  = optional(string, "test.com")
    }))
  })
  default       = {
    name        = "Bob"
    age         = 23
    email       = {
        address = "bob"
        domain  = "example.com"
    }
  }
}
```
3. Create a resource using the above template version. 0.0.8.
4. Delete the email domain value for `Test Object`.
<img width="2408" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f06215a0-9008-4e4e-a1cf-71a73dee19c4">
5. Check the setting after deploying.

**Result**

1. The computed attribute of email domain is wrong. It should be 'test.com'.
<img width="945" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/9d816440-f953-4c8f-a282-76868869426c">

<img width="673" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/beb9a283-05c7-4380-8c08-2d066715dba8">

**Expected behavior**

**Environment**
- Walrus version: main a12b753f2e823aa7a7456110215dceed4da46ba1
- UI version: dev-07fa8db

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 3806cbeb980963a23e83bf42b8ac3b15c287fbba
- UI version: v0.5.0-rc2

**Test Result:** pass

PR #1995, opened 2024-01-22
fix: support compute attributes base on user input

aiwantaozi:
**Problem:**
While user input is different from default, computed attributes use default to calculate now.

**Solution:**
While user input is different from default, use user input attributes to generate computed attributes.

**Related Issue:**
#1938


alexcodelf (approved):
LGTM

shanewxy (approved):
LGTM

Issue #1991, opened 2024-01-22
Restarting Walrus through docker-compose, Walrus startup failure


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Run Walrus server by docker compose.
```
version: '3.6'
services:
  seal:
    image: "sealio/walrus:main"
    restart: always
    privileged: true
    container_name: seal
    command:
      - "walrus"
      - "--log-debug"
      - "--log-verbosity=5"
    environment:
      - K3S_RESOLV_CONF=
    ports:
      - "80:80"
      - "443:443"
    volumes:
      - 'walrus-app:/var/run/walrus'
volumes:
  walrus-app: {}
```
2. Execute `docker-compose down`.
3. Execute `docker-compose up`.
4. Check the running status of Walrus server.

**Result**

1. Walrus failed to startup.
```
2024-01-22T10:03:55.007Z	I	k3s	time="2024-01-22T10:03:55Z" level=fatal msg="flannel exited: failed to acquire lease: subnet \"192.168.109.0/24\" specified in the flannel net config doesn't contain \"192.168.169.0/24\" PodCIDR of the \"local\" node"
2024-01-22T10:03:55.026Z	I	signal.handler	received signal "child exited"
2024-01-22T10:03:55.026Z	I	signal.handler	reaped child process 47
2024-01-22T10:03:55.026Z	E	error running embedded kubernetes: exit status 1
github.com/seal-io/walrus/utils/log.DelegatedLogger.Errorf
	github.com/seal-io/walrus/utils@v0.0.0-00010101000000-000000000000/log/delegate.go:168
github.com/seal-io/walrus/utils/log.Errorf
	github.com/seal-io/walrus/utils@v0.0.0-00010101000000-000000000000/log/logger.go:174
github.com/seal-io/walrus/pkg/server.(*Server).Run.func1
	github.com/seal-io/walrus/pkg/server/runner.go:525
github.com/seal-io/walrus/utils/gopool.contextWaitGroup.Go.func1
	github.com/seal-io/walrus/utils@v0.0.0-00010101000000-000000000000/gopool/waitgroup.go:115
github.com/alitto/pond.(*TaskGroupWithContext).Submit.func1
	github.com/alitto/pond@v1.8.3/group.go:60
github.com/alitto/pond.(*WorkerPool).executeTask
	github.com/alitto/pond@v1.8.3/pond.go:454
github.com/alitto/pond.worker
[...]

PR #1999, opened 2024-01-23
fix: failed to restart after infra network changed

thxCode:
**problem**

failed to restart walrus with steps like `docker compose up -> docker compose down -> docker compose up`, the root cause is that the `compose down` has removed the docker network created by the first-time up. 

generally, if the outside networking changes, it affects the embedded Kubernetes cluster, and we cannot reset the joined node's PodCIDR at present.

**solution**

embedded Kubernetes cluster is for demonstration purposes, we never guarantee using it for production. so we can easily remove the Etcd data(similar to removing the node from a Kubernetes cluster) if the networking changes.

**related**

#1991  

@gitlawr, we need a WARNING at docs to tell users that it's dangerous to run workloads in the default-local environment.

### DIFF
--- a/pkg/k8s/local.go
+++ b/pkg/k8s/local.go
@@ -130,7 +130,7 @@ func (Embedded) Run(ctx context.Context) error {
 		return fmt.Errorf("error link server data: %w", err)
 	}
 
-	// Reset server data.
+	// Reset server database.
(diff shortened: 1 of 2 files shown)


gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #1959, opened 2024-01-18
cli prints a warning message on list commands


gitlawr:
<img width="1022" alt="image" src="https://github.com/seal-io/walrus/assets/5697937/98cd20c9-a4c1-43cb-999a-a99a2187e1b3">

**Environment**
```
clientVersion:
  version: dev
  gitCommit: 9a4516ef8771ed0bdce4acb990f084aa4e9a2a11
  isDevVersion: true
```

Finenyaco:
**Environment**
```
clientVersion:
  version: main
  gitCommit: 4eef11db3efc80ce93bb17814e5e8fb822d75d45
  isDevVersion: true
serverVersion:
  version: main
  gitCommit: 4eef11db3efc80ce93bb17814e5e8fb822d75d45
  isDevVersion: true
```

**Test Result:** pass
```
 ~/ w variable list
 ID                   NAME               VALUE         CREATED
-------------------- ------------------ ------------- -----------------------------
 497531122612204802   project-password   <sensitive>   2024-01-24T07:27:38.284776Z
 ~/ w project list
 ID                   NAME      CREATED
-------------------- --------- ----------------------------
 497517331371554050   default   2024-01-24T05:10:38.06008Z
 ~/ w environment list
 ID                   NAME    TYPE          CREATED
-------------------- ------- ------------- -----------------------------
 497517331371554050   local   development   2024-01-24T05:10:38.064002Z
```

PR #2000, opened 2024-01-23
refactor: enhance cli

aiwantaozi:
**Problem:**
1. cli wait for the tls handshake timeout while the remote server is unreachable
2. show version compare log while listing resource

**Solution:**
1. reduce the timeout by adding dial to check reachable, reduce the tls handshake timeout and remove duplicate validation multiple times
2. remove version compare while listing resource
3. remove env clone command

**Related Issue:**
#1960 
#1959

### DIFF
--- a/pkg/cli/config/config.go
+++ b/pkg/cli/config/config.go
@@ -70,19 +70,56 @@ func (c *Config) DoRequest(req *http.Request) (*http.Response, error) {
 	return c.DoRequestWithTimeout(req, timeout)
 }
 
-// ValidateAndSetup validate and setup the context.
-func (c *Config) ValidateAndSetup() error {
-	msg := `cli configuration is invalid: %s. You can configure cli by running "walrus login"`
-
-	if c.Server == "" {
-		return fmt.Errorf(msg, "server address is empty")
+func (c *Config) CheckReachable() (err error) {
+	u, err := url.Parse(c.Server)
+	if err != nil {
+		return err
 	}
 
-	if c.Token == "" {
+	if u.Port() == "" {
+		if u.Scheme == "http" {
+			u.Host += ":80"
+		} else if u.Scheme == "https" {
+			u.Host += ":443"
+		}
+	}
+
+	var conn net.Conn
+
+	conn, err = net.DialTimeout("tcp", u.Host, 3*time.Second)
+	if err != nil {
+		return fmt.Errorf("access %s failed %w", c.Server, err)
+	}
+
+	defer conn.Close()
+
+	return nil
+}
+
+// ValidateAndSetup validate and setup the context.
+func (c *Config) ValidateAndSetup() (err error) {
+	// Reachable set.
+	c.Reachable = true
+
+	defer func() {
+		if err != nil {
+			c.Reachable = false
+		}
+	}()
+
+	msg := `cli configuration is invalid: %v. You can configure cli by running "walrus login"`
+
+	switch {
+	case c.Server == "":
+		return fmt.Errorf(msg, "server address is empty")
+	case c.Token == "":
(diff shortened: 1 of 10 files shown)


pkg/cli/cmd/version.go
```diff
@@ -29,6 +29,11 @@ func Version(sc *config.Config) *cobra.Command {
 			}
 
 			fmt.Print(string(b))
+
+			err = api.CompareVersion(info)
+			if err != nil {
+				panic(err)
+			}
```
gitlawr, line 36: Can we skip this check? Just print the version info.

pkg/cli/config/config.go
```diff
@@ -71,18 +71,46 @@ func (c *Config) DoRequest(req *http.Request) (*http.Response, error) {
...
+
 	msg := `cli configuration is invalid: %s. You can configure cli by running "walrus login"`
 
-	if c.Server == "" {
-		return fmt.Errorf(msg, "server address is empty")
+	switch {
+	case c.Server == "":
+		err = fmt.Errorf(msg, "server address is empty")
+	case c.Token == "":
+		err = fmt.Errorf(msg, "token is empty")
```
gitlawr, line 90: In local docker mode, authn is disabled and empty token is used

pkg/cli/config/config.go
```diff
@@ -70,20 +70,57 @@ func (c *Config) DoRequest(req *http.Request) (*http.Response, error) {
...
 	}
 
-	if c.Token == "" {
-		return fmt.Errorf(msg, "token is empty")
+	err = c.CheckReachable()
```
alexcodelf, line 119: Is it intend to overwirte err?

pkg/cli/config/config.go
```diff
@@ -115,7 +153,7 @@ func (c *Config) HttpClient(timeout time.Duration) *http.Client {
 		DialContext: (&net.Dialer{
 			Timeout: 3 * time.Second,
 		}).DialContext,
-		TLSHandshakeTimeout: 10 * time.Second,
+		TLSHandshakeTimeout: 5 * time.Second,
```
thxCode, line 156: why do we need to change this?

pkg/cli/config/config.go
```diff
@@ -94,7 +131,8 @@ func (c *Config) ValidateAndSetup() error {
 	}
 
 	if err != nil {
-		return fmt.Errorf(msg, err.Error())
+		err = fmt.Errorf(msg, err.Error())
+		return
```
thxCode, line 135:
```
return fmt.Errorf(`cli configuration is invalid: %w. You can configure cli by running "walrus login"`, err)
```
or 
```
return fmt.Errorf(`cli configuration is invalid: %v. You can configure cli by running "walrus login"`, err)
```

pkg/cli/config/server_context.go
```diff
@@ -26,6 +26,9 @@ type ServerContext struct {
 	Server   string `json:"server" survey:"server"`
 	Token    string `json:"token" survey:"token"`
 	Insecure bool   `json:"insecure" survey:"insecure"`
+
+	// Reachable.
+	Reachable bool
```
thxCode, line 31: can we use `Unreachable bool` here?

pkg/cli/config/config.go
aiwantaozi: updated

pkg/cli/config/server_context.go
aiwantaozi: Prefer positive and affirmative variable names, Reachable might be more intuitive.

gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #2001, opened 2024-01-23
Resource creation fails when the template contains a variable of type map(any): invalid map_any: value must be an object


Finenyaco:
**Describe the bug**

Resource creation fails when the template contains a variable of type map(any).

**Steps to reproduce**
1. Add a template with source `https://github.com/unitong/test-upgrade`.
2. Using this template to create a resource.
3. Configure `Map Any` as below and save.

<img width="846" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/96253010-8ade-4de2-aa92-da4e74b87bb6">

**Result**

1. Failed to create: `invalid variables: violate ui schema: invalid map_any: value must be an object`.
<img width="928" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/59b9ac80-856b-4603-ac4f-3f0d10915ca2">

2. Check the post api request.
<img width="932" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/3c954c27-8344-4dab-a307-b67ebb670f96">

3. Check the ui schema.
<img width="1155" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/21a3fb38-e6ec-459c-bc22-dccefe3c8401">

**Expected behavior**

Create successfully.

**Environment**
- Walrus version: main 3806cbeb980963a23e83bf42b8ac3b15c287fbba
- UI version: v0.5.0-rc2

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 66a2214c8ea5dfb7fd8542fecb5e87a3778fc07e
- UI version: dev-4dd2e5a

**Test Result:** pass

PR #2004, opened 2024-01-23
refactor: skip validate for any type attributes

aiwantaozi:
**Problem:**
Validate any type with schema may causes deploy fail

**Solution:**
Skip validate any type with schema

**Related Issue:**
#2001

### DIFF
--- a/pkg/dao/types/property/field.go
+++ b/pkg/dao/types/property/field.go
@@ -95,7 +95,7 @@ func (i Values) ValidateWith(schema *openapi3.Schema) error {
 			err error
 		)
 
-		switch s.Type {
+		switch {
 		default:
 			_, ok, err := GetAny[any](v)
 			if !ok || err != nil {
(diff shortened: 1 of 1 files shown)


pkg/dao/types/property/field.go
```diff
@@ -143,3 +146,39 @@ func (i Values) ValidateWith(schema *openapi3.Schema) error {
...
+
+// HasAnyTypes check whether the schema include any type.
+// Object without properties and additional properties will be considered as any type.
+func HasAnyTypes(s *openapi3.Schema) bool {
+	switch s.Type {
+	case openapi3.TypeBoolean,
+		openapi3.TypeInteger,
+		openapi3.TypeNumber,
+		openapi3.TypeString:
+		return false
```
gitlawr, line 158: Looks like we can remove these

gitlawr (approved):
LGTM

Issue #1582, opened 2023-11-24
Cli support display more meaningful fields


shanewxy:
**Is your enhancement related to a problem? Please describe.**

Our cli command seems to support list only `id`, `name`, and `createTime`, but extension APIs like getVersions need more information like `versions` returned.
```
$ walrus template get-versions terraform-aws-customer-gateway
                 id                             name                    createTime
-------------------- -------------------------------- -----------------------------
 488228946027360846   terraform-aws-customer-gateway   2023-11-21T03:18:48.977076Z
 488228946027295310   terraform-aws-customer-gateway   2023-11-21T03:18:48.977076Z
 488228946027229774   terraform-aws-customer-gateway   2023-11-21T03:18:48.977076Z
 488228946027164238   terraform-aws-customer-gateway   2023-11-21T03:18:48.977076Z
 488228946027098702   terraform-aws-customer-gateway   2023-11-21T03:18:48.977076Z
 488228946027033166   terraform-aws-customer-gateway   2023-11-21T03:18:48.977076Z
 488228946026967630   terraform-aws-customer-gateway   2023-11-21T03:18:48.977075Z
```
**Describe the solution you'd like**

Support display meaningful fields for specific APIs.

**Describe alternatives you've considered**

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 387b458b939852736bdcc40f2f83fec184da6b3f

Project and environment information is not displayed when getting resources of resource definition.

```
 ~/ walrus resourcedefinition get-resources 490697291011777538
                 id         name                    createTime
-------------------- ------------ -----------------------------
 490728095540634626   test-ports    2023-12-08T09:05:38.16371Z
 490727888761447426   resource-2   2023-12-08T09:03:34.916244Z
 490727851818017794   resource-1   2023-12-08T09:03:12.892028Z
```

thxCode:
Can we also fix this https://github.com/seal-io/walrus/issues/1607? or already fixed?

Finenyaco:
**Environment**
- Walrus version: main 3806cbeb980963a23e83bf42b8ac3b15c287fbba
- UI version: v0.5.0-rc2

**Test Result:** 

```
 ~/ w template get-versions 497362217688147202  --project ""
 ID                   NAME                                  VERSION   CREATED
-------------------- ------------------------------------- --------- -----------------------------
[...]

PR #2005, opened 2024-01-23
refactor: expose more column while cli list/get

aiwantaozi:
**Problem:**
variables, settings and resource outputs list did not show the value

**Solution:**
support more columns

**Related Issue:**
#1582


pkg/dao/types/resource_revision.go
```diff
@@ -9,7 +9,7 @@ import (
 
 type OutputValue struct {
 	Name      string          `json:"name,omitempty"`
-	Value     property.Value  `json:"value,omitempty"`
+	Value     property.Value  `json:"value,omitempty,cli-table-column"`
 	Type      cty.Type        `json:"type,omitempty"`
 	Sensitive bool            `json:"sensitive,omitempty"`
```
shanewxy, line 14: Do we need to expose sensitive as well like in variable?

aiwantaozi: Now the outputs API doesn't return the sensitive field, it uses <sensitive> to replace the value. So we customize the variable list, and replace value with <sensitive> for sensitive value.
Variable list:
image

Output list:
image

shanewxy (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1997, opened 2024-01-23
Issues with walrus template commands


Finenyaco:
**Describe the bug**

**Issue 1:  Unable to get versions of global templates.**

```
 ~/ w template get-versions -h
Usage:
  walrus template get-versions <template> [flags]

Flags:
      --extract strings
  -o, --output string     Output format [table, json, yaml] (default "table")
      --page int           (default 1)
      --per-page int       (default 100)
      --project string
      --query string
      --sort strings

Global Flags:
  -d, --debug   Enable debug log
  -h, --help    Help for this command

 ~/ w  template get-versions 497350247396201986
unexpected status code 404: model: template not found
```

**Issue 2: Unable to get templates that imported through the catalog using their names**

```
 ~/ w template list
 ID                   NAME                                         STATUS   CREATED
-------------------- -------------------------------------------- -------- -----------------------------
 497353104690304514   test                                         Ready    2024-01-23T01:59:11.338665Z
 497352883952473602   walrus-catalog/kubernetes-redis              Ready    2024-01-23T01:56:59.760923Z
 497352883885364738   walrus-catalog/generic-sql-seeder            Ready    2024-01-23T01:56:59.726967Z
 497352883868587522   walrus-catalog/kubernetes-containertask      Ready    2024-01-23T01:56:59.714496Z
 497352882559964674   walrus-catalog/aws-rds-postgresql            Ready    2024-01-23T01:56:58.938148Z
 497352882442524162   walrus-catalog/aws-elasticache-redis         Ready    2024-01-23T01:56:58.866367Z
 497352877526799874   walrus-catalog/alicloud-rds-mysql            Ready    2024-01-23T01:56:55.933183Z
 497352876704716290   walrus-catalog/docker-redis                  Ready    2024-01-23T01:56:55.449542Z
 497352876453058050   walrus-catalog/static-mysql                  Ready    2024-01-23T01:56:55.299621Z
 497352875681306114   walrus-catalog/alicloud-rds-postgresql       Ready    2024-01-23T01:56:54.839415Z
 497352875513533954   walrus-catalog/deploy-source-code            Ready    2024-01-23T01:56:54.737685Z
 497352870614586882   walrus-catalog/kubernetes-containerservice   Ready    2024-01-23T01:56:51.811411Z
 497352870262265346   walrus-catalog/kubernetes-postgresql         Ready    2024-01-23T01:56:51.604491Z
[...]

PR #2007, opened 2024-01-24
fix: API get template by name doesn't work while template name include slash

aiwantaozi:
**Problem:**
API get template by name doesn't work while template name include slash like (${catalog_name}/${template_name}

**Solution:**
enable UseRawPath to support getting path parameters from rawPath

**Related Issue:**
#1997


alexcodelf (approved):
LGTM

gitlawr (approved):
LGTM

Issue #2008, opened 2024-01-24
Configuration in the matching rule does not take effect when the resource definition schema is empty


Finenyaco:
**Describe the bug**

Configuration in the matching rule does not take effect when the resource definition schema is empty.

**Steps to reproduce**
1. New a resource definition and add a matching rule using 'builtin/kubernertes-mysql' template as below:

<img width="2384" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/216c2057-9bff-44ec-82de-094ba3f11207">

2. Edit the resource definition and delete all components in UI schema.
<img width="1026" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/0b3772b6-a510-4ba1-8572-a537bcb21d4e">

3. Using the definition to create a resource.
4. Check the resource configuration on revision page.

**Result**

1. The configuration in the matching rule does not take effect.

**Expected behavior**

**Environment**
- Walrus version: main 4eef11db3efc80ce93bb17814e5e8fb822d75d45
- UI version: dev-4dd2e5a

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 4eef11db3efc80ce93bb17814e5e8fb822d75d45
- UI version: dev-94aee09

**Test Result:** pass

PR #2010, opened 2024-01-24
fix: matching rule does not take effect when the resource definition …

aiwantaozi:
**Problem:**
matching rule does not take effect when the resource definition is empty

**Solution:**
fix computed attributes generate function

**Related Issue:**
#2008


gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

PR #2015, opened 2024-01-24
fix: walrus local installed return access failed

aiwantaozi:
**Problem:**
Walrus local install return access failed

**Solution:**
Set reachable while check access


gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #2020, opened 2024-01-25
Importing terraform aws modules failed: does not satisfy server version constraint


Finenyaco:
**Describe the bug**

Importing terraform aws modules failed.

**Steps to reproduce**
1. Add a catalog with source `https://github.com/terraform-aws-modules`.

**Result**

1. Failed to import templates.

<img width="2447" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/505eafee-41f7-4ea3-8b44-0f4e64a1b896">

2. Check Walrus logs:
```
2024-01-25T03:34:10.685Z	D	template	get "terraform-aws-modules/aws-pricing:v1.0.0" of catalog "497652116857290754" schema
2024-01-25T03:34:10.685Z	W	template	failed to validate "terraform-aws-modules/aws-pricing:v1.0.0" of catalog "497652116857290754" schema: invalid schema: at least one of requiredProviders, variables, outputs must be specified
2024-01-25T03:34:10.691Z	D	template	get "terraform-aws-modules/aws-pricing:v1.1.0" of catalog "497652116857290754" schema
2024-01-25T03:34:10.691Z	W	template	failed to validate "terraform-aws-modules/aws-pricing:v1.1.0" of catalog "497652116857290754" schema: invalid schema: at least one of requiredProviders, variables, outputs must be specified
2024-01-25T03:34:10.695Z	D	template	get "terraform-aws-modules/aws-pricing:v1.2.0" of catalog "497652116857290754" schema
2024-01-25T03:34:10.695Z	W	template	failed to validate "terraform-aws-modules/aws-pricing:v1.2.0" of catalog "497652116857290754" schema: invalid schema: at least one of requiredProviders, variables, outputs must be specified
2024-01-25T03:34:10.701Z	D	template	get "terraform-aws-modules/aws-pricing:v1.3.0" of catalog "497652116857290754" schema
2024-01-25T03:34:10.701Z	W	template	failed to validate "terraform-aws-modules/aws-pricing:v1.3.0" of catalog "497652116857290754" schema: invalid schema: at least one of requiredProviders, variables, outputs must be specified
2024-01-25T03:34:10.706Z	D	template	get "terraform-aws-modules/aws-pricing:v1.3.1" of catalog "497652116857290754" schema
[...]

PR #2022, opened 2024-01-25
fix: check template constrain failed

alexcodelf:
**Problem:**

Can not import template with out file schema.yaml. Walrus version constrain failed.

**Solution:**

Skip version constrain if there is not file shcema.

**Related Issue:**
#2020


gitlawr (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #2023, opened 2024-01-25
UI label for kubernetes connector name


gitlawr:
Change `Cluster Name` to `Name` on creation

<img width="1132" alt="image" src="https://github.com/seal-io/walrus/assets/5697937/700a3f7c-e0be-4920-bb36-f975a877e07e">

**Environment**
- Walrus version: v0.5.0-rc3

Finenyaco:
**Environment**
- Walrus version: main a803edf588fc3dc8024ebdb4745b184b3b40bba7
- UI version: dev-186c2e0

**Test Result:** pass

<img width="1003" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/0642008f-cd7f-40f5-8057-b5be075da709">
<img width="908" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/b6b60558-828a-429f-9998-c9d0428a097e">

PR #2038, opened 2024-01-25
fix: crash after use template without components

aiwantaozi:
**Problem:**
Crash while creating definition with template with schema doesn't include component

**Solution:**
Check schema empty before use it

**Related Issue:**
#2023


alexcodelf (approved):
LGTM

gitlawr (approved):
LGTM

Issue #2018, opened 2024-01-25
local install fails in rc


gitlawr:
**Describe the bug**
![image](https://github.com/seal-io/walrus/assets/5697937/c54c6736-2859-4710-b098-a6552b25e0bd)

**Steps to reproduce**
1. run cli local install

**Environment**
- Walrus version: v0.5.0-rc2

Finenyaco:
**Environment**
- Walrus version: v0.5.0-rc4

**Test Result:**

- MacOS docker desktop: pass
<img width="1273" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/70f953e7-cd17-4621-b923-3714ecf48e7e">

```
 ~/Downloads/ docker ps
CONTAINER ID   IMAGE                      COMMAND                   CREATED              STATUS          PORTS                                                               NAMES
d08c28065590   sealio/walrus:v0.5.0-rc4   "walrus --log-debug …"   About a minute ago   Up 45 seconds   0.0.0.0:7080->80/tcp, 0.0.0.0:7443->443/tcp                         sealio_walrus-docker-extension-desktop-extension-service
```

- Ubuntu docker: pass
<img width="2235" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/713e2553-7513-4e31-bfa5-a8155217438b">

PR #2039, opened 2024-01-25
fix: ready check fail after local install

gitlawr:
**Problem:**
local install returns
```
runtime error: invalid memory address or nil pointer dereference
```

**Related Issue:**
https://github.com/seal-io/walrus/issues/2018


alexcodelf (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #2032, opened 2024-01-25
On the resource definition details page, if two resource names are the same, batch operations will return an error


Finenyaco:
**Describe the bug**

On the resource definition details page, if two resource names are the same, it is unable to deploy resources in batch or delete resources in batch.
**Steps to reproduce**

https://github.com/seal-io/walrus/assets/26084756/1ecaaf23-f839-4623-a68b-30b4835aef11

```
2024-01-25T09:06:35.516Z	E	api	error requesting DELETE /v1/resource-definitions/497684247608295442/resources?withoutCleanup=false: 1 error occurred:
	* duplicate resource name: env-type


github.com/seal-io/walrus/pkg/apis/runtime.erroring
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_error.go:45
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.recovering
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_recovery.go:39
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.observing
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_observation.go:89
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/gin-gonic/gin.(*Engine).handleHTTPRequest
	github.com/gin-gonic/gin@v1.9.1/gin.go:620
github.com/gin-gonic/gin.(*Engine).ServeHTTP
	github.com/gin-gonic/gin@v1.9.1/gin.go:576
github.com/seal-io/walrus/pkg/apis/runtime.(*Router).ServeHTTP
	github.com/seal-io/walrus/pkg/apis/runtime/router.go:194
net/http.serverHandler.ServeHTTP
	net/http/server.go:2938
net/http.(*conn).serve
	net/http/server.go:2009
2024-01-25T09:06:35.516Z	D	api	500 | HTTP/1.1 |       67 B |      116 B |     2.30308ms |  46.137.225.211 | DELETE  /v1/resource-definitions/497684247608295442/resources?withoutCleanup=false
[...]

PR #2040, opened 2024-01-25
fix: resource definition batch resource operations fail for duplicate names

shanewxy:
**Problem:**

Resource definition detail page manage resources across different environments, which results in duplicate resource names failing on name checks.

**Solution:**

Change name checking to ID checking.

**Related Issue:**
#2032

### DIFF
--- a/pkg/apis/resource/common.go
+++ b/pkg/apis/resource/common.go
@@ -8,6 +8,7 @@ import (
 	"github.com/seal-io/walrus/pkg/dao"
 	"github.com/seal-io/walrus/pkg/dao/model"
 	"github.com/seal-io/walrus/pkg/dao/model/resource"
+	"github.com/seal-io/walrus/pkg/dao/types/object"
 	"github.com/seal-io/walrus/pkg/dao/types/status"
 	"github.com/seal-io/walrus/pkg/deployer"
 	deployertf "github.com/seal-io/walrus/pkg/deployer/terraform"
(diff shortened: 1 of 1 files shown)


alexcodelf:
Changing names to IDs can result in failures when batch creating resources since we use names to establish dependencies. Can we conduct the name check on a project-by-project basis? In case there are multiple resources spread across different projects, can we check each one individually?

gitlawr:
What happen when we clone multiple resources with dependencies? Previously, we get dependencies from attributes before relationships are stored in db.

pkg/apis/resource/common.go
```diff
@@ -31,6 +31,7 @@ func DeleteResources(req CollectionDeleteRequest, mc model.ClientSet, kubeConfig
 
 		resources, err := tx.Resources().Query().
```
gitlawr, line 32: What about we group these resources by their environment and do TopologicalSort per group?

pkg/apis/resource/common.go
alexcodelf, line 1: Please also check collection upgrade.

pkg/apis/resource/common.go
```diff
@@ -55,44 +49,83 @@ func DeleteResources(req CollectionDeleteRequest, mc model.ClientSet, kubeConfig
...
+		environmentIDToResources[r.EnvironmentID] = append(environmentIDToResources[r.EnvironmentID], r)
+	}
+
+	return environmentIDToResources, nil
+}
 
-		if err = upgrade(req.Context, kubeConfig, mc, entity, req.Draft); err != nil {
+func UpgradeResources(req CollectionRouteUpgradeRequest, mc model.ClientSet, kubeConfig *rest.Config) error {
+	err := mc.WithTx(req.Context, func(tx *model.Tx) error {
+		environmentIDToResources, err := getResourcesGroupByEnvironment(tx, req.IDs(), req.Context)
```
alexcodelf, line 104: The resources will not receive request resources update data `req.Model()` if we query from DB.

gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #2033, opened 2024-01-25
Failed to redeploy the resource after updating the matching rule


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. New a resource definition as below:

<img width="972" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/fec24b9a-0de2-4e2c-bebb-102cee26390c">

2. Using the definition to create a resource.
3. Update the definition as below:
<img width="962" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/561ac730-3720-4187-92f0-468b1f2f95b1">

4. Update the project label, make it match the rule and redeploy the resource.

**Result**

1. Failed to redeploy the resource after updating the matching rule.

https://github.com/seal-io/walrus/assets/26084756/33e2592e-36e7-4100-9d02-2d61aa98fc1f

**Expected behavior**

**Environment**
- Walrus version: v0.5.0-rc3

**Additional context**

aiwantaozi:
Haven't reproduced it, need more info.

Finenyaco:
**Steps to reproduce**

1. Create two resource definitions with the same type and one identical rule.
<img width="1121" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/30cb1c06-e294-47aa-8378-9469d8c22325">
<img width="996" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/2ac8a51f-f78d-4e03-be12-4b614e49ff16">

2. Create a resource that meets the above conditions. The resource matches the first created resource definition (my-test).
3. Modify the rule of the resource definition created later  (my-test-2).
4. Modify the resource created in step 2 to match the rules of "my-test-2".
5. Redeploy the resource.

gitlawr:
Matched definition changes and the redeploy follows validation of the new schema. It should be by design.

PR #2041, opened 2024-01-26
fix: skip validate user input while isn't existed

aiwantaozi:
**Problem:**
While the user does not input the variables, we still validate it schema.

**Solution:**
Skip validate while the user doesn't input variables

**Related Issue:**
#2033

### DIFF
--- a/pkg/apis/resource/basic_view.go
+++ b/pkg/apis/resource/basic_view.go
@@ -655,6 +655,10 @@ func validateAttributesWithTemplate(
 			return fmt.Errorf("failed to inject attributes before valiate: %w", err)
 		}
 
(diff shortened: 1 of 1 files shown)


pkg/apis/resource/basic_view.go
```diff
@@ -655,6 +655,10 @@ func validateAttributesWithTemplate(
 			return fmt.Errorf("failed to inject attributes before valiate: %w", err)
 		}
 
+		if len(injectedAttrs) == 0 {
```
gitlawr, line 658: Should we check len(attrs) so as to return earlier?

gitlawr:
Why do we skip in this case? What happen if there's a required variable and users provide empty input?

PR #2044, opened 2024-01-26
refactor(workflow): remove aggregate roles of argo workflow

alexcodelf:
**Problem:**

No need to interact with argo CD crds right now.

**Solution:**

Remove aggregate roles when install argo workflow chats.

**Related Issue:**


thxCode (approved):
LGTM

gitlawr (approved):
LGTM

Issue #2049, opened 2024-01-26
 Sometimes fails to redeploy resource using docker-containerservices template: violates foreign key constraint 


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Run Walrus server by executing `walrus local install`.
2. New a resource using `contianerservice` definition and set 'Profile' to 'run'.
3. Edit the resource, modify 'Profile' to 'init' and deploy.

**Result**

1. Failed to deploy.
<img width="2316" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/40403fbd-5cc6-4fad-bc57-e97c8eceffbf">

```
model: constraint failed: insert nodes to table "resource_component_relationships": pq: insert or update on table "resource_component_relationships" violates foreign key constraint "resource_component_relationshi_30de23d07752addc6670562911741a82"
```

2. The deployment logs are empty.

<img width="919" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/341dab56-90d4-40a8-bb15-4f95713b15cf">

**Expected behavior**

**Environment**
- Walrus version: v0.5.0-rc4

**Additional context**

This error may be triggered in other scenarios, but it only occurs in Docker mode.
<img width="2450" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/53117621-d6e5-4780-a60e-513eb0ded8c3">

Finenyaco:
**Environment**
- Walrus version: main eb8e29d68c15d16e0fe6b1bc95dca3bdab96a454
- UI version: dev-bb86dbc

**Test Result:** pass

PR #2056, opened 2024-01-27
fix: components relationship update failed

alexcodelf:
**Problem:**

Updating resource components relationships can lead to the violation of foreign key constraints. This issue arises during the update of resource components (see https://github.com/seal-io/walrus/blob/main/pkg/apis/resourcerevision/extension.go#L200). In the update process, old resources get deleted. However, if a relationship was created using a resource prior to its deletion, it can result in a query failure.

**Solution:**

Clear removed resource components before create components relationships.

**Related Issue:**
#2049

### DIFF
--- a/pkg/dao/resource_component.go
+++ b/pkg/dao/resource_component.go
@@ -4,10 +4,13 @@ import (
 	"context"
 	"errors"
 
+	"k8s.io/apimachinery/pkg/util/sets"
+
 	"github.com/seal-io/walrus/pkg/dao/model"
 	"github.com/seal-io/walrus/pkg/dao/model/resourcecomponent"
 	"github.com/seal-io/walrus/pkg/dao/model/resourcecomponentrelationship"
 	"github.com/seal-io/walrus/pkg/dao/types"
+	"github.com/seal-io/walrus/pkg/dao/types/object"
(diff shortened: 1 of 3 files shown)


pkg/dao/resource_component.go
```diff
@@ -116,3 +118,36 @@ func ResourceComponentGetUniqueKey(r *model.ResourceComponent) string {
 	// Align to schema definition.
 	return strs.Join("-", string(r.ConnectorID), r.Shape, r.Mode, r.Type, r.Name)
 }
+
+// clearComponents clears the components that not exist in the database.
+func clearComponents(
+	ctx context.Context,
+	mc model.ClientSet,
+	components model.ResourceComponents,
+) (model.ResourceComponents, error) {
```
thxCode, line 127: i didn't see any delete/remove action in this function, can we clarify the purpose of this function?

thxCode:
Can we try to get a clean component slice from outside?

pkg/dao/resource_component.go
alexcodelf: updated, we need refactor the resource components update logic in 0.6.0.

thxCode (approved):
LGTM

gitlawr (approved):
LGTM

PR #2059, opened 2024-01-29
fix: keep required after defaulting in res def

thxCode:
after creating a resource definition, we turn the required fields optional if we have different values.

**related**

https://github.com/seal-io/walrus/issues/2060


gitlawr:
Let's add some test cases to prevent unexpected behavioral changes in the future

thxCode:
> Let's add some test cases to prevent unexpected behavioral changes in the future

updated

pkg/resourcedefinitions/helper_test.go
```diff
@@ -41,24 +104,69 @@ func Test_alignSchemas(t *testing.T) {
...
+				expectedOpenAPISchema.Components.Schemas == nil {
+				t.Fatalf("expected file %q has no components.schemas", expectedPath)
+			}
+
+			expectedSchemas = expectedOpenAPISchema.Components.Schemas
+		}
+
+		var nb map[string]any
+		{
+			nbJSONPath := filepath.Join(dir, tc.Name(), "nb.json")
```
gitlawr, line 128: What does `nb.json` stand for?

thxCode: notebook, a shared data uses for the previous alignSchema func, here we mock that with this json file.

gitlawr: Can we improve the naming or add comments

thxCode: comment in the testdata and the function.

gitlawr (approved):
LGTM

aiwantaozi (approved):
LGTM

PR #2064, opened 2024-01-29
fix: cli should support parse catalog provider

orangedeng:
When doing local install, add SERVER_BUILTIN_CATALOG_PROVIDER environment to docker run command

**Problem:**

walrus local doesn't support setting `gitee` as catalog provider.

**Solution:**

**Related Issue:**


orangedeng:
Ok, then I will close this PR

Issue #2057, opened 2024-01-28
Add checksum for release assets


gitlawr:

gitlawr:
Verify in rc

Finenyaco:
**Environment**
- Walrus version: v0.5.1-rc2

**Test Result:** pass

<img width="656" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/990c80ad-9060-46e3-b765-06924ae105a3">


```
 ~/Downloads/ shasum -a 256 walrus-cli-darwin-amd64
583a598274d5c39ac5b365f687a758453be0b9a83dbc913de888b456aee72bec  walrus-cli-darwin-amd64

 ~/Downloads/ shasum -a 256 walrus-cli-darwin-arm64
c5374ba4b080e7bd78735c0e642946175a3f9e61c602a44bd698bdad1c7dcdde  walrus-cli-darwin-arm64

 ~/Downloads/ shasum -a 256 walrus-cli-linux-amd64
f074022553d565f730968e02a88b76c0e3fd67b3f6f4dc75119bc7e61f8c12ad  walrus-cli-linux-amd64

 ~/Downloads/ shasum -a 256 walrus-cli-linux-arm64
a64c60f5669fd621c6e8164f2384309650f1e855ac740085e2b7582338ab1942  walrus-cli-linux-arm64
```

PR #2068, opened 2024-01-30
ci: add checksums for release assets

shanewxy:
**Problem:**

Add checksum for release assets.

**Related Issue:**
#2057

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -208,3 +208,6 @@ jobs:
             hack/mirror/walrus-images.txt
             hack/mirror/walrus-load-images.sh
             hack/mirror/walrus-save-images.sh
+      - name: Generate checksum
+        uses: wangzuo/action-release-checksums@v1
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}


.github/workflows/ci.yml
```diff
@@ -196,6 +196,13 @@ jobs:
           remote_path: releases/${{ steps.metadata.outputs.version == 'main' && 'latest' || steps.metadata.outputs.version }}
           accelerate: true
           clean: false
+      - name: Generate SHA256 checksums
+        if: ${{ startsWith(github.ref, 'refs/tags/') }}
+        run: |
+          sha256sum .dist/package/${{ matrix.target }}/${{ matrix.task }}/build/cli/* > sha256sums.txt
+          sha256sum hack/mirror/walrus-images.txt >> sha256sums.txt
+          sha256sum hack/mirror/walrus-load-images.sh >> sha256sums.txt
+          sha256sum hack/mirror/walrus-save-images.sh >> sha256sums.txt
```
thxCode, line 205: `sed` always includes a useless dirname, e.g. https://github.com/seal-io/terraform-provider-courier/blob/20612f54d1335c3b206ef3e519b7273384f4d8ff/hack/release.sh#L17, we need to trim them.

thxCode: can we find a release action implementation with shasum generation?

shanewxy: Updated to use https://github.com/wangzuo/action-release-checksums. There are other implementations like https://github.com/marketplace/actions/generate-checksum, but turns out to include path as well.

thxCode (approved):
LGTM, @gitlawr we need a RC version to verify this later.

thxCode:
https://github.com/wangzuo/action-release-checksums/blob/8e38f587f6e4ea5d1b2e350be1c51c5d72c36934/index.js#L38-L54, it looks like the new action download the assets from remote which uploading recently by the previous step. I approve of this at present. I look forward to integrating into the `make package` step later.

gitlawr (approved):
LGTM

Issue #2058, opened 2024-01-29
Warlsu v0.5.0 image Trivy scan report and fix


pengjiang80:
Scan warlus:v0.5.0 image with Trivy, there are some critical vulnerabilities. We need to check if we can update and mitigate these in v0.6.0.

Finenyaco:
**Environment**
- Walrus version: main fa6e56aca10e61259243c938130c8e63dd19ba27
- UI version: dev-4218197

**Test Result:** No critical vulnerabilities.

```
go.mod (gomod)

Total: 2 (UNKNOWN: 0, LOW: 0, MEDIUM: 2, HIGH: 0, CRITICAL: 0)

┌───────────────────────────────┬────────────────┬──────────┬──────────┬───────────────────┬───────────────┬─────────────────────────────────────────────────────────┐
│            Library            │ Vulnerability  │ Severity │  Status  │ Installed Version │ Fixed Version │                          Title                          │
├───────────────────────────────┼────────────────┼──────────┼──────────┼───────────────────┼───────────────┼─────────────────────────────────────────────────────────┤
│ github.com/go-jose/go-jose/v3 │ CVE-2024-28180 │ MEDIUM   │ fixed    │ 3.0.1             │ 3.0.3         │ Package jose aims to provide an implementation of the   │
│                               │                │          │          │                   │               │ Javascript Objec ......                                 │
│                               │                │          │          │                   │               │ https://avd.aquasec.com/nvd/cve-2024-28180              │
├───────────────────────────────┼────────────────┤          ├──────────┼───────────────────┼───────────────┼─────────────────────────────────────────────────────────┤
│ helm.sh/helm/v3               │ CVE-2019-25210 │          │ affected │ 3.14.2            │               │ helm: shows secrets with --dry-run option in clear text │
│                               │                │          │          │                   │               │ https://avd.aquasec.com/nvd/cve-2019-25210              │
└───────────────────────────────┴────────────────┴──────────┴──────────┴───────────────────┴───────────────┴─────────────────────────────────────────────────────────┘
```

PR #2074, opened 2024-01-30
refactor: reduce false positives vulns reporting

thxCode:
**problem**

Lots of false positive vulns reporting at container image

**solution**

- bump base image to reduce false positive vulns reporting.
- bump dependent software, like Casdoor, K3s, PostgreSQL to fix vulns.

**related**
#2058

**BREAK CHANGES**
- bump PostgreSQL to 16.1, `postgres:16.1`
- bump Casdoor to v1.515.0, `sealio/casdoor:v1.515.0-seal.1`
- bump K3s to v1.27.9-k3s1
cc @shanewxy @hibig 

**NOTE**
this PR can reduce false positive vulns reporting, but we still get some unfixable vulns as shown.

[scan.txt](https://github.com/seal-io/walrus/files/14096342/scan.txt)

- casdoor relies on a lower version beego, although it looks terrible, it gets a lower EPSS: https://api.first.org/data/v1/epss?cve=CVE-2021-27116,CVE-2021-27117. honestly saying, for the using walrus suggests, we can ignore these risks. 
  + we will remove the dependency of casdoor later.
- crane is a tool for image building, there is no need for when the walrus is running.
  + we will remove this tool from the image later.
- containerd/k3s/runc/etcdctl are for embedding use, which is for demonstration purposes.
- `/etc/ssl/private/ssl-cert-snakeoil.key`, see https://askubuntu.com/questions/396120/what-is-the-purpose-of-the-ssl-cert-snakeoil-key. there is no need for when the walrus is running.


gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #2076, opened 2024-01-31
Merge the user-input parameters and the preset values in the resource definition for variables of type map and object


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

**Test Steps:**

1. New a resource definition.
2. Add two matching rules using template `aws-rds-mysql` and `alicloud-rds-mysql`.
3. Fill in different Vpc Ids in two templates and save.
4. Remove the field for `vpc_id` from the UI schema of the resource definition.
5. Using the definition to create a resource.

**Test Result:**
1. The user-input parameters are not merged with the preset values in the resource definition.
<img width="835" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f60c6014-ee9c-4d45-8fb7-bc42b2e6b357">

<img width="750" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/d924e9ab-0b47-4a61-9b40-21f91e824e6b">

**Describe the solution you'd like**

Merge the user-input parameters and the preset values in the resource definition for variables of type map and object.

**Describe alternatives you've considered**

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 57e7bb83ce7aacac99c5a1c723bb488ad5ef5a4e
- UI version: dev-a712ac5

**Test Result:** pass

PR #2083, opened 2024-01-31
refactor: update generate schema default with attributes

aiwantaozi:
**Problem:**
Use user input attributes while exist, compare with merge it with ops configuration, merge is more intuitive. 

**Solution:**
After discussion, it was decided that the merge operation would eventually be performed, and the mutually exclusive parts of the merge would need to be handled in the template.

**Related Issue:**
#2076

### DIFF
--- a/pkg/templates/openapi/default.go
+++ b/pkg/templates/openapi/default.go
@@ -12,7 +12,6 @@ import (
 
 	"github.com/seal-io/walrus/pkg/dao/types/property"
 	"github.com/seal-io/walrus/utils/gopool"
-	"github.com/seal-io/walrus/utils/json"
 )
 
 type patchResult struct {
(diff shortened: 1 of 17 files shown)


thxCode:
looks like all cases are preparing for `with_attributes_`, let's shorten the testdata's filename, remove `with_attributes_`.

if we want to emphasize testing with `GenSchemaDefaultWithAttribute`, we can put all cases under `gen_schema_default_with_attribute`.

aiwantaozi:
> looks like all cases are preparing for `with_attributes_`, let's shorten the testdata's filename, remove `with_attributes_`.
> 
> if we want to emphasize testing with `GenSchemaDefaultWithAttribute`, we can put all cases under `gen_schema_default_with_attribute`.

Each case dir includes some files shared with two cases (main.tf, schema.yaml) and other files used for specific tests. files with prefix: with_attributes are for `TestGenSchemaDefaultWithAttribute`, and without a prefix are for `TestGenSchemaDefaultPatch`.

thxCode (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #2062, opened 2024-01-29
Support configuring server options for local Walrus


gitlawr:
**Is your enhancement related to a problem? Please describe.**
Be able to configure [server options](https://seal-io.github.io/docs/deploy/options) when running `walrus local install`. e.g., set builtin-catalog-provider to gitee.

Finenyaco:
**Environment**
- Walrus version: main bd8b2f4560a522fcfb061a081817869f1a088f3a
- UI version: dev-95fef06

**Test Result:** pass

```
root@iZbp14mscl82upj4amm48eZ:~# walurs local install -e SERVER_BUILTIN_CATALOG_PROVIDER=gitee
Installing...
Checking readiness...
Walrus CLI is configured.
```

<img width="816" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/0b298e9f-e68a-40ee-98a0-1a8ef2fcd437">


Note: The Docker extension is not available when using the `-e` flag.

PR #2087, opened 2024-01-31
feat: support configuring local walrus env

gitlawr:
Refer to https://github.com/seal-io/docs/pull/191/files for the UX.

**Related Issue:**
https://github.com/seal-io/walrus/issues/2062

### DIFF
--- a/pkg/cli/local/install.go
+++ b/pkg/cli/local/install.go
@@ -16,7 +16,7 @@ const (
 	walrusDockerExtensionName = "sealio/walrus-docker-extension"
 )
 
-func InstallLocalWalrus() error {
+func InstallLocalWalrus(opts InstallOptions) error {
 	if !IsDockerInstalled() {
 		return errors.New("docker is not available")
 	}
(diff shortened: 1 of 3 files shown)


thxCode:
the `--env-file` is syntactic sugar that loads the env vars from a file without specifying `--env` one by one, which means modifying `local.env` also cannot update the existing container's running options after `docker restart`. we can verify this by `docker inspect`.

if so, I suggest supporting something like this: `SERVER_foo walrus local install/reinstall`, and also give a fixed docker network to avoid the IP change after reinstall. See https://github.com/seal-io/walrus/pull/1999.

to implement this, we can filter out the env vars from `os.Environ` with `SERVER_` prefix, and lock the special env vars we need, like `SERVER_ENABLE_AUTHN` and `SERVER_SETTING_LOCAL_ENVIRONMENT_MODE`, finally, inject the filter list into `cmd.Env`.

pkg/cli/local/install.go
```diff
@@ -77,12 +82,20 @@ func InstallLocalWalrusDockerContainer() error {
...
 		"/var/run/docker.sock:/var/run/docker.sock",
-		fmt.Sprintf("sealio/walrus:%s", getLocalWalrusTag()),
-		"walrus",
-		"--enable-authn=false",
-	)
+	}
+
+	for _, env := range opts.Env {
+		runArgs = append(runArgs, "-e", env)
+	}
```
thxCode, line 93: image

should we allow a user to overwrite `SERVER_SETTING_LOCAL_ENVIRONMENT_MODE` and `SERVER_ENABLE_AUTHN` ?

gitlawr: Updated.

thxCode (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #2080, opened 2024-01-31
Fix the walrus image tag in walrus-images.txt


linyinli:
**Describe the bug**
The walrus image tag in walrus-images.txt is still v0.4.0:
![image](https://github.com/seal-io/walrus/assets/42955482/4575ad53-ac53-4948-96b0-5269574c77ff)

**Expected behavior**
Fix the walrus image tag.

**Environment**
- Walrus version: v0.5.0

gitlawr:
Verify in rc

Finenyaco:
**Environment**
- Walrus version: v0.5.1-rc2

**Test Result:** pass

<img width="612" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/3389babc-5d9a-4adf-9039-4d3a045ececc">

PR #2091, opened 2024-02-01
fix: failed upload cli

thxCode:
## problem

for now, the package layout looks as shown, which causes two problems: 1. UI cannot download CLI as not found; 2. failed to upload CLI to Tencent COS.

image

## solution

make sure the package layout is as below:
image

- upload assets to COS  from `walrus/server/cos` dir.
- release assets to Github from `walrus/server/release` dir.

## related
#2080

### DIFF
--- a/hack/package.sh
+++ b/hack/package.sh
@@ -39,12 +39,12 @@ function download_ui() {
 }
 
 function download_walrus_file_hub() {
-    local path="${1}"
+  local path="${1}"
 
-    mkdir -p "${path}"
-    if ! git clone --depth 1 https://github.com/seal-io/walrus-file-hub "${path}"; then
-      seal::log::fatal "failed to download walrus-file-hub"
-    fi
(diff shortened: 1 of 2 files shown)


hack/package.sh
```diff
@@ -96,6 +139,9 @@ function package() {
   # shellcheck disable=SC2155
   local context="$(setup_image_package_context "${target}" "${task}" "${path}")"
 
+  setup_release "${context}"
+  setup_tencent_cos "${context}"
```
shanewxy, line 143: Do we need to check `PACKAGE_BUILD=false` before setup release?

thxCode: that's ok, we just do these steps before PACKAGE_BUILD.

shanewxy (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #2069, opened 2024-01-30
Support Walrus cli in Windows OS


pengjiang80:
Currently Walrus cli only has MacOS and Linux versions. Windows should also be supported.

thxCode:
<img width="1094" alt="image" src="https://github.com/seal-io/walrus/assets/3394724/e477b606-d390-4aa9-b091-02761d3c4340">

there are lots of paths util cannot recognize windows path, I will make another PR to modify this.

Finenyaco:
Windows OS: arm64 win11

Verified in 5d939ccf90935cc12b8b7cd7d28e5bd255c22df5.

PR #2092, opened 2024-02-01
refactor: support windows file path

thxCode:
the original path processing uses the `path` package, which is for URL purpose, let's change to `filepath` package.

## related
https://github.com/seal-io/walrus/issues/2069#issuecomment-1920667135


alexcodelf (approved):
LGTM

shanewxy (approved):
LGTM

Issue #2094, opened 2024-02-01
Returning incorrectly computed attributes causes the value to be unable to display on the resource details page


Finenyaco:
**Describe the bug**

The computed attributes returned by the resource interface are incorrectly named.

**Steps to reproduce**
1. Go to the resource details page and click on the `Settings` tab.
2. Check the attributes.

**Result**

1. Did not display play the computed attributes. 
2. Check the GET resource api response, `computed_attributes` should be `computedAttributes`.
<img width="2486" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/1719c3b4-421c-4a30-a8c3-63137a8a1de0">

**Expected behavior**

**Environment**
- Walrus version: main 57e7bb83ce7aacac99c5a1c723bb488ad5ef5a4e
- UI version: dev-a712ac5

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 5d939ccf90935cc12b8b7cd7d28e5bd255c22df5
- UI version: dev-a712ac5

**Test Result:** pass

<img width="885" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/88518a8e-3a99-4ec5-a416-87f9ee4c47d9">

PR #2095, opened 2024-02-01
fix: update compute_attributes to computeAttributes

aiwantaozi:
**Problem:**
Wrong compute_attributes field name format 

**Solution:**
Update compute_attributes to computeAttributes

**Related Issue:**
#2094


thxCode (approved):
LGTM

Issue #2097, opened 2024-02-01
Mark objects without required properties as optional after merging schema


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

1. New a resource definition.
2. Add two matching rule using template `bulitin/aws-rds-mysql`.
3. Fill in different “Vpc Id” in the two rules and save.
4. Check the resource definition UI Schema.

There were no required properties in `Infrastructure`. However, it is still marked as required.
<img width="2013" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/7f28c80d-54a8-49fb-a9c3-d3469f0bd02e">

**Describe the solution you'd like**

When an object doesn't have required properties, set the object as optional.

**Describe alternatives you've considered**

**Additional context**

- Walrus version: main 57e7bb83ce7aacac99c5a1c723bb488ad5ef5a4e
- UI version: dev-a712ac5

Finenyaco:
Verified in b926a09a84b4ea20551cc9ded7a0d1f57b9c4f68.

Issue #2060, opened 2024-01-29
 When there are preset values in the matching rule, the field is still marked as required


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. New a resource definition.
2. Add two matching rule using template `bulitin/aws-rds-mysql`.
3. Fill in different “Vpc Id” in the two rules and save.
4. Check the resource definition UI Schema.

**Result**

1. 'VPC ID' is marked as required.
<img width="2393" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/febd1a10-20b3-41a2-95be-f37e6022b8a4">

<img width="814" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/e3ba65f1-ce27-42d4-bb89-1fed6484ec2d">

**Expected behavior**

When there are preset values in the matching rule, it should remove the field from the required list in the schema automatically.

**Environment**
- Walrus version: v0.5.0

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 57e7bb83ce7aacac99c5a1c723bb488ad5ef5a4e
- UI version: dev-a712ac5

**Test Result:** 

1. Create a resource definition using a template that includes a required list-type variable.
2. Add two matching rules with different values.
3. Check the definition UI schema. The variable is still required.

<img width="974" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/d638719c-d3f8-4c77-ad11-dd33075650b9">

Finenyaco:
**Environment**
- Walrus version: main 5d939ccf90935cc12b8b7cd7d28e5bd255c22df5
- UI version: dev-a712ac5

**Test Result:** 

This issue still can be reproduced in variables of type list(map(string)).

Finenyaco:
**Environment**
- Walrus version: main bd8b2f4560a522fcfb061a081817869f1a088f3a
- UI version: dev-95fef06

**Test Result:** pass

Finenyaco:
**Environment**
- Walrus version: main b926a09a84b4ea20551cc9ded7a0d1f57b9c4f68
- UI version: dev-5eecc2b

**Test Result:** pass

- One required property.
<img width="1953" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/6fb8da96-984f-4b87-a1a3-f9c95ac39336">

- No required property.
[...]

PR #2099, opened 2024-02-02
fix: failed make optional on array property

thxCode:
make optional on the array property if all default values are not the same.

### related
https://github.com/seal-io/walrus/issues/2060#issuecomment-1921260126
https://github.com/seal-io/walrus/issues/2097


gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

PR #2101, opened 2024-02-02
fix: builtin resource definition computed

aiwantaozi:
**Problem:**
Builtin resource definition matching rule's default value is empty

**Solution:**
Add template default in merge sequence while generate resource definition matching rule's default

**Related Issue:**
#2094


thxCode (approved):
LGTM

alexcodelf (approved):
LGTM

PR #2104, opened 2024-02-02
fix: failed make optional on map property

thxCode:
make optional on the map property if all default values are not the same.

### related
https://github.com/seal-io/walrus/issues/2060#issuecomment-1923382504
https://github.com/seal-io/walrus/issues/2097


alexcodelf (approved):
LGTM

aiwantaozi (approved):
LGTM

PR #2114, opened 2024-02-04
fix: failed make optional on ancestor node

thxCode:
### background

It's hard to refill defaults to complex structures

The past five PRs have been trying to solve the problem of default refilling, and they solved a part of this issue but not completely.

### solution

To better clarify where the problem lies, this PR separates the cases that appeared in the past into a completed testdata, and describes the final expected in the given input.

Refilling has always followed this principle, finding the same parts in all rule default values (the input of each rule is regarded as the default of this rule), and filling it into the merge schema.

However the structural description of the schema is indeed complex, and this must be dealt with in conjunction with the expectations: when we should default an object, and when should not.

Hope the new `complex` case makes everyone understand the expectations of the refilling default value, including me.

### related
https://github.com/seal-io/walrus/issues/2060#issuecomment-1925740095


shanewxy (approved):
LGTM

gitlawr (approved):
LGTM

PR #2121, opened 2024-02-05
fix: value contains line breaks make unmarshal failed

alexcodelf:
**Problem:**

Can not save resoruce that attributes contains new line charactors.

**Solution:**

Escape special characters before unmarshal json.

**Related Issue:**
#2117

Issue #2117, opened 2024-02-05
Could not inject an ssh private key variable to resource


linyinli:
**Describe the bug**
Could not inject an ssh private key variable to resource.

**Steps to reproduce**
1. Create a variable by using an ssh private key;
2. Create a resource, reference the ssh private key variable in "Connection Private Key" attribute.

**Result**
400 Bad Request error:
<img width="1510" alt="image" src="https://github.com/seal-io/walrus/assets/42955482/10932853-0438-4e3a-a34a-2444e73ffc77">

**Expected behavior**
Can inject an ssh private key variable to resource.

**Environment**
- Walrus version: v0.5.0

Finenyaco:
**Environment**
- Walrus version: main 700d971e404e208e929bcbb088ec974f07c871c0
- UI version: dev-7fa18e0

**Test Result:** failed

<img width="1810" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/1c25a05d-466f-41ae-9896-053814700878">

Finenyaco:
**Environment**
- Walrus version: main e92cde6256262c9671c9d51475f0bee8669f6628
- UI version: dev-7fa18e0

**Test Result:** pass

PR #2127, opened 2024-02-06
fix: replace variables error

alexcodelf:
This is a further repair for the variable replacement, as missing the quotation marks for the var key in #2121.

The key should be `"${var.abc}"` rather than `${var.abc}`

**Related Issue:**
#2117


thxCode (approved):
LGTM

gitlawr (approved):
LGTM

Issue #2126, opened 2024-02-06
Failed to reference workflow variables


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. New a workflow.
2. Add a workflow variable.
3. Add a 'Deploy Resource' step and reference a workflow variable.
4. Run the workflow and check deployment result.

<img width="1187" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/0824e3e1-5f59-410a-9d82-b1f2fcc0e782">

**Result**

1. The values of workflow variables were not replaced during deployment.
<img width="1196" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/9fb0a8e6-2377-4935-8e8c-60cf779d4ee5">

<img width="444" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/bc8353a3-6f42-4b7d-866f-c6f051a65a62">

**Expected behavior**

**Environment**
- Walrus version: main 700d971e404e208e929bcbb088ec974f07c871c0
- UI version: dev-7fa18e0

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main e92cde6256262c9671c9d51475f0bee8669f6628
- UI version: dev-7fa18e0

**Test Result:** pass

Issue #1982, opened 2024-01-22
Return more information when the resource upgrade fails through the workflow


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

When the resource upgrade fails through the workflow, only a status code is currently returned, lacking sufficient information for users to pinpoint where the error occurred.

<img width="705" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/78580dfb-3a04-4201-a495-b650e8750939">

**Describe the solution you'd like**

Return more information when the upgrade fails.

**Describe alternatives you've considered**

**Additional context**

Steps to reproduce:
1. New a workflow and add a 'Deploy Resource' step.
2. Run the workflow.
3. Edit the workflow, edit the step and change the resource type.
4. Run the workflow again.

Finenyaco:
**Environment**
- Walrus version: main e92cde6256262c9671c9d51475f0bee8669f6628
- UI version: dev-7fa18e0

**Test Result:** pass

<img width="653" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/cf1f2f32-0523-438d-aa2e-4eebf36c4293">

<img width="553" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f39dac81-f6f5-4df8-a8bf-5c22d2af7065">

PR #2128, opened 2024-02-07
fix: workflow variables and logs

alexcodelf:
**Problem:**

- Can not reference workflow variables with hyphen
- When update create or update resource in step error occurs, workflow step will not show curl error.

**Solution:**

- support hyphen in workflow, as we adjust variables name rules in 0.5.0
- show more info in workflow service step. As the upgrade API of resource will not return any information, we may change it in the future with K8s API.

**Related Issue:** 
#2126  #1982

### DIFF
--- a/pkg/dao/workflow.go
+++ b/pkg/dao/workflow.go
@@ -353,7 +353,7 @@ func parseWorkflowVariables(
 		return nil, err
 	}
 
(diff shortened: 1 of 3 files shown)


gitlawr (approved):
LGTM

Issue #2073, opened 2024-01-30
Remove isDevVersion in cli version output


gitlawr:
**Steps to reproduce**
1. Download Walrus cli
2. Run `walrus version`

**Result**
```
clientVersion:
  version: v0.5.0
  gitCommit: 44305cd5fb89716a1e183344072ec5f805819605
  isDevVersion: false
```

**Expected behavior**
Do not print isDevVersion

Finenyaco:
**Test Result:** Pass

```
root@ip-192-168-1-125:~# walrus version
clientVersion:
  version: main
  gitCommit: 3b93fab7fcdac5c5ba61984ee7431021a7593592
serverVersion:
  version: main
  gitCommit: e92cde6256262c9671c9d51475f0bee8669f6628
```

Issue #2103, opened 2024-02-02
Enhancements of not being able to get the server version


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

When failed to get to the Walrus server version, it's also unable to view the current version of the Walrus CLI client.
```
root@ip-192-168-1-125:/opt/walrus# w version
got not found response from https://xxx/debug/version, failed to decode response body: readObjectStart: expect { or n, but found <, error found in #1 byte of ...|<html>
<he|..., bigger context ...|<html>
<head><title>404 Not Found</title></head>
|...
```

```
 ~/seal/ w version
unexpected status code 503 from https://192.168.50.182/debug/version, <html>
<head><title>503 Service Temporarily Unavailable</title></head>
<body>
<center><h1>503 Service Temporarily Unavailable</h1></center>
<hr><center>nginx</center>
</body>
</html>
```

**Describe the solution you'd like**

1. Even though it's failed to get the server version, the client version should still be displayed.
2. Optimize the logs for when getting the server version fails to make it more understandable.

**Describe alternatives you've considered**

**Additional context**

clientVersion:
  version: main
  gitCommit: 5d939ccf90935cc12b8b7cd7d28e5bd255c22df5

thxCode:
also consider https://github.com/seal-io/walrus/issues/2073

Finenyaco:
**Test Result:** Pass

Did not print server version when failing to retrieve it.
```
 ~/ w version
clientVersion:
  version: main
  gitCommit: 3b93fab7fcdac5c5ba61984ee7431021a7593592
```

PR #2139, opened 2024-02-23
refactor: enhance cli version command

aiwantaozi:
**Problem:**
1. isDev info is useless, version is enough.
2. Now it won't return the client version while it can't get the server version.

**Solution:**

1. Removed isDev info.
2. Still return cli version, while the server isn't reachable.

**Related Issue:**
#2073

### DIFF
--- a/pkg/cli/api/version.go
+++ b/pkg/cli/api/version.go
@@ -14,19 +14,17 @@ type VersionInfo struct {
 
 // Version include the version and commit.
 type Version struct {
-	Version      string `json:"version" yaml:"version"`
-	GitCommit    string `json:"gitCommit" yaml:"gitCommit"`
-	IsDevVersion bool   `json:"isDevVersion" yaml:"isDevVersion"`
+	Version   string `json:"version" yaml:"version"`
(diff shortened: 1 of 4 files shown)


pkg/cli/api/version.go
```diff
@@ -14,45 +14,42 @@ type VersionInfo struct {
...
 		},
 	}
 
 	if sc != nil && sc.Reachable {
 		// Server version.
 		sv, err := sc.ServerVersion()
 		if err != nil {
-			return nil, err
+			// Return client version if server version is not reachable.
+			return info
```
gitlawr, line 36: Is there any message showing the server is not reachable?

aiwantaozi: Prefer just show the client version while the server isn't reachable. While the user run other commands need interact with the server would show the unreachable error.

gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

gitlawr:
https://github.com/seal-io/walrus/issues/2103

Issue #2096, opened 2024-02-01
 Default values for array-type items were not submitted when creating the resource


Finenyaco:
**Describe the bug**

Default values for array-type items were not submitted when creating the resource

**Steps to reproduce**

https://github.com/seal-io/walrus/assets/26084756/2a9eed17-926c-4d20-87b7-9f7ce4883f5f

- The definition of 'Test Number List'.

<img width="593" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/1baeaafb-c269-4b40-a39a-2b0d981fe8f0">

- Check the POST request:

<img width="722" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/2653a4c9-efb0-4d0a-a2ad-575046aa2292">

**Expected behavior**

**Environment**
- Walrus version: main 57e7bb83ce7aacac99c5a1c723bb488ad5ef5a4e
- UI version: dev-a712ac5

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main e92cde6256262c9671c9d51475f0bee8669f6628
- UI version: dev-78d87e4

**Test Result:** pass

Finenyaco:
**Environment**
- Walrus version: main 30ceb5ac9dcaade04505b64a3f6e6ef3bc1c80b2
- UI version: dev-1305100

**Test Result:** pass

<img width="884" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/8dc70792-fb67-4d01-bfcb-0ba9fde7651b">

PR #2141, opened 2024-02-28
refactor: add array item default to computed attributes

aiwantaozi:
**Problem:**
While UI schema configs the default value for an object's field in the array items, and creates a resource, the default value isn't set to computed attributes.

**Solution:**
Set the default value to array items while the item is object type.

**Related Issue:**
#2096


gitlawr (approved):
LGTM

Issue #2045, opened 2024-01-26
UI schema will be reset after refreshing the templates


Finenyaco:
**Describe the bug**
 
UI schema will be reset after refreshing the templates

**Steps to reproduce**
1. Modify a template UI schema.
2. Refresh the template.

**Result**

1. The UI schema will be reset after refresh.

**Expected behavior**

**Environment**
- Walrus version: v0.5.0-rc3

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 30ceb5ac9dcaade04505b64a3f6e6ef3bc1c80b2
- UI version: dev-1305100

**Test Result:** pass

PR #2143, opened 2024-02-28
fix: template ui schema is reset after refreshed

aiwantaozi:
**Problem:**
UI schema will be reset after refreshing the template

**Solution:**
Prevent reset user edited UI schema while refreshing

**Related Issue:**
#2045

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -65,6 +65,7 @@ require (
 	github.com/r3labs/sse v0.0.0-20210224172625-26fe804710bc
 	github.com/redis/go-redis/v9 v9.4.0
 	github.com/sashabaranov/go-openai v1.18.1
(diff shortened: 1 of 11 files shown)


pkg/apis/templateversion/basic_view.go
```diff
@@ -35,5 +38,14 @@ func (r *UpdateRequest) Validate() error {
 		r.TemplateVersionUpdateInput.Schema = tv.Schema
 	}
 
+	// Set user edited tags.
```
gitlawr, line 41: Why do we set tags? Can we reject the update by checking tv status?

gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #1531, opened 2023-11-23
Prettify CLI json output


gitlawr:
It's missing indentation.
<img width="439" alt="image" src="https://github.com/seal-io/walrus/assets/5697937/70208b04-af9c-4f3f-93d7-07418ae0a0a8">

Finenyaco:
**Environment**
- Walrus version: main 387b458b939852736bdcc40f2f83fec184da6b3f
- UI version: dev-4478a77

1. Create a service using `kubernetes-containerservice` template.
2. Execute cli command `walrus resource get-outputs <service-name>`

**Test Result:** 
Need better readability.
<img width="1669" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/6f1aa683-0b63-42e8-9433-8f151daad40f">

Finenyaco:
**Test Result:** Failed

```
clientVersion:
  version: main
  gitCommit: 0a3779654c43dfb1a5d3299c4862502a909f2878
serverVersion:
  version: main
  gitCommit: 0a3779654c43dfb1a5d3299c4862502a909f2878
```

The formatting is incomplete.

<img width="718" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/0ba1ffc3-3867-4d91-8ed5-f6058792ed99">

<img width="625" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/3fb192b3-d77c-47e5-ae72-551793c0f115">

PR #2148, opened 2024-03-01
refactor: use encoding/json while generate cli output

aiwantaozi:
**Problem:**
json-iterator's MarshalIndent only only indent the first level. While generating cli from ci, it will force to use json-iterator, which will cause wrong indent.
https://github.com/json-iterator/go/issues/564
https://github.com/json-iterator/go/issues/645

**Solution:**
Update to use default encoding/json now, will update to json-iterator after bugs get fixed.

**Related Issue:**
#1531

### DIFF
--- a/pkg/cli/formatter/json.go
+++ b/pkg/cli/formatter/json.go
@@ -2,11 +2,10 @@ package formatter
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"net/http"
-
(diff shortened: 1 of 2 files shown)


gitlawr:
Can you leave a comment so that it's noticeable in code

aiwantaozi:
> Can you leave a comment so that it's noticeable in code

Updated.

alexcodelf (approved):
LGTM

gitlawr (approved):
LGTM

Issue #1998, opened 2024-01-23
After resetting the schema, the returned schema is not correct


Finenyaco:
**Describe the bug**

After resetting the schema, the returned schema is not correct; manual page refresh is needed.

**Steps to reproduce**

https://github.com/seal-io/walrus/assets/26084756/c0dc4aac-f02d-40c7-b606-76eccc055528

**Expected behavior**

**Environment**
- Walrus version: main 0b2678c0df661cd20cafbdb716a641717f158e7f
- UI version: v0.5.0-rc2

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 7bb79389ab0436969f4c4214e38918ee6c6e9070
- UI version: dev-1be55ba

**Test Result:** pass

PR #2156, opened 2024-03-04
fix: reset ui schema couldn't get the result immediately

aiwantaozi:
**Problem:**
Now reset UI schema is asynchronous, it need to clone the repo to get the UI schema.
And need to refresh the UI to get the UI schema after reset.

**Solution:**
Store the original UI schema, and set UI schema to the original one when reset.

**Related Issue:**
#1998


pkg/dao/types/schema.go
```diff
@@ -243,6 +243,10 @@ func (s *UISchema) Validate() error {
 type TemplateVersionSchema struct {
 	Schema `json:",inline"`
 
+	// OriginalUISchema retains the UI schema either from the schema file or exposed from the schema.
+	// It replaces the UI schema with the value when the user resets the UI schema.
+	OriginalUISchema UISchema `json:"originalUISchema,omitempty"`
```
gitlawr, line 248: Why don't we use two TemplateVersionSchemas instead of gluing them together

pkg/templates/loader/terraform_loader.go
```diff
@@ -125,61 +137,65 @@ func (l *TerraformLoader) loadSchema(
...
 
-// getVariableSchema generate variable schemas from terraform files or schema.yaml.
-func (l *TerraformLoader) getVariableSchema(
-	rootDir string,
-	mod *tfconfig.Module,
-	mode Mode,
-) (*openapi3.Schema, *openapi3.Info, error) {
-	fromOriginal, err := l.getVariableSchemaFromTerraform(mod)
+// getVariableSchemaAndInfoFromFile get variables schema and openapi info from schema.yaml.
+func (l *TerraformLoader) getSchemaFromFile(rootDir string, originalSchema *openapi3.T) (*openapi3.T, error) {
```
gitlawr, line 153: Inconsistency between funcion name and the comment

gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #2152, opened 2024-03-01
Panic when viewing the undeployed resource details page


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. New a resource and save as draft.
2. Go to the resource details page.

**Result**

1. Returned an error.
<img width="2447" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f984fc54-4282-4ffb-a64c-1a916735372c">

2. Check Walrus logs.

```
2024-03-01T10:08:24.258Z	E	api	panic observing: runtime error: invalid memory address or nil pointer dereference, callstack:
runtime.panicmem
	runtime/panic.go:261 (0x62e9b)
runtime.sigpanic
	runtime/signal_unix.go:861 (0x62e68)
github.com/seal-io/walrus/pkg/apis/resource.Handler.getResourceOutputs
	github.com/seal-io/walrus/pkg/apis/resource/extension.go:243 (0x51a596c)
github.com/seal-io/walrus/pkg/apis/resource.Handler.RouteGetOutputs
	github.com/seal-io/walrus/pkg/apis/resource/extension.go:237 (0x51a4f6f)
reflect.Value.call
	reflect/value.go:596 (0xfaec3)
reflect.Value.Call
	reflect/value.go:380 (0xfa343)
github.com/seal-io/walrus/pkg/apis/runtime.(*Router).Routes.func1
	github.com/seal-io/walrus/pkg/apis/runtime/router_route.go:150 (0x4ff97db)
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x501640f)
github.com/seal-io/walrus/pkg/auths.Account.Filter
	github.com/seal-io/walrus/pkg/auths/account.go:56 (0x50163f4)
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x10ad5ab)
github.com/seal-io/walrus/pkg/apis.(*Server).Setup.If.func5
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_filter.go:54 (0x565e44b)
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x10ad5ab)
github.com/seal-io/walrus/pkg/apis.(*Server).Setup.RequestThrottling.func3
[...]

PR #2157, opened 2024-03-04
fix: server crash when get output for draft resource

aiwantaozi:
**Problem:**
server crash when get output for draft resource 

**Solution:**
Skip decoding output when isn't existed

**Related Issue:**
#2152


alexcodelf (approved):
LGTM

gitlawr (approved):
LGTM

Issue #2024, opened 2024-01-25
Search users with name shows unexpected result.


alexcodelf:
**Describe the bug**

Incorrect result when querying users

**Steps to reproduce**
1.  Create four users `admin`, `test`, `manager`, `ad`
2. Search users with keyworkds `n`

**Result**

<img width="3008" alt="image" src="https://github.com/seal-io/walrus/assets/85595852/5f6385f3-4891-482c-8ca5-68853f2b68eb">

**Expected behavior**

Only shows manager and admin, the name contains key words.

**Environment**
- Walrus version: v0.5.0-rc3

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 7bb79389ab0436969f4c4214e38918ee6c6e9070
- UI version: dev-1be55ba

**Test Result:** pass

PR #2159, opened 2024-03-04
fix: search users with name unexpected result

shanewxy:
**Problem:**

Searching users with keywords `n` returns all records, because `n` is in the value `builtin` of the domain field.

**Solution:**

Remove the domain field from the query fields.

**Related Issue:**
#2024


gitlawr (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #2163, opened 2024-03-04
Resource run stuck in Canceling state


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. New a resource and click `Save and Preview`.
2. Deploy the resource directly.

**Result**

1. The first resource run stuck in Canceling state. 
<img width="2430" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/20f5ee98-b1cf-4959-ba95-f2df1002778d">

**Expected behavior**

**Environment**
- Walrus version: main 30ceb5ac9dcaade04505b64a3f6e6ef3bc1c80b2
- UI version: dev-5cf4cb0

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main aba5070fb2d7f86be3c14bd419ee7be9a8b3d302
- UI version: dev-bb39486

**Test Result:** pass


<img width="2424" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/06a576da-85ca-4650-b59a-3aaad9761719">

PR #2165, opened 2024-03-04
fix: canceling status and dependency failed update

alexcodelf:
**Problem:**

- resource run status stuck when canceling.
- resource schedule failed run status will stuck.

**Solution:**

- fix resource run status stuck in canceling.
- fix resource scheduler failed run status not updated.

**Related Issue:**

#2163


gitlawr (dismissed):
LGTM

shanewxy (approved):
LGTM

Issue #2175, opened 2024-03-05
Incorrect resource status when plan failed


Finenyaco:
**Describe the bug**

Incorrect resource status when plan failed.

**Steps to reproduce**
1. New a resource and reference an output of a resource that doesn't exist.
2. Check the resource status.

**Result**

1. The resource is stuck in the "Deploying" state.
<img width="2540" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/87c52989-da9f-4145-b49e-73eca61df7e4">

**Expected behavior**

**Environment**
- Walrus version: main 7bb79389ab0436969f4c4214e38918ee6c6e9070
- UI version: dev-1be55ba

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main aba5070fb2d7f86be3c14bd419ee7be9a8b3d302
- UI version: dev-8f22705

**Test Result:** Failed 

The resource stuck in `Progressing` state now. And the message of status is incorrect: `Waiting for dependent resources to be ready:`

<img width="2455" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/3bab7eb4-6863-43fc-889e-7da5a6e74404">

<img width="2388" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/7edb43d0-e2e2-43c4-98ba-eda522a7ba15">

<img width="883" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/ed2b838a-6510-43b8-9f76-9949fc20d8c3">

Issue #2172, opened 2024-03-05
 Preview does not take effect during rollback


Finenyaco:
**Describe the bug**

Preview does not take effect during rollback.

**Steps to reproduce**
1. Rollback a resource and preview.
2. Check the run result.

**Result**

1. It applied automatically when the plan is completed

**Expected behavior**

1. Does not apply automatically.

**Environment**
- Walrus version: main 7bb79389ab0436969f4c4214e38918ee6c6e9070
- UI version: dev-1be55ba

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main aba5070fb2d7f86be3c14bd419ee7be9a8b3d302
- UI version: dev-8f22705

**Test Result:** pass

<img width="2430" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/2d06d6a7-8dc7-4bc9-8bb6-d5aef350fa7e">

Issue #2167, opened 2024-03-04
Panic when deploying a resources from a draft state


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. New a resource and save draft.
2. Start the resource. 

**Result**

1. Failed to deploy.
2. Checked the Walrus logs. A panic occurred.

```
2024-03-04T10:46:57.876Z	E	api	panic observing: runtime error: invalid memory address or nil pointer dereference, callstack:
runtime.panicmem
	runtime/panic.go:261 (0x62e9b)
runtime.sigpanic
	runtime/signal_unix.go:861 (0x62e68)
github.com/seal-io/walrus/pkg/apis/resourcerun.Handler.RouteGetTerraformStates
	github.com/seal-io/walrus/pkg/apis/resourcerun/extension.go:59 (0x518b9f4)
reflect.Value.call
	reflect/value.go:596 (0xfaec3)
reflect.Value.Call
	reflect/value.go:380 (0xfa343)
github.com/seal-io/walrus/pkg/apis/runtime.(*Router).Routes.func1
	github.com/seal-io/walrus/pkg/apis/runtime/router_route.go:150 (0x4ffae3b)
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x5017a6f)
github.com/seal-io/walrus/pkg/auths.Account.Filter
	github.com/seal-io/walrus/pkg/auths/account.go:56 (0x5017a54)
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x10acfbb)
github.com/seal-io/walrus/pkg/apis.(*Server).Setup.If.func5
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_filter.go:54 (0x565f33b)
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x10acfbb)
github.com/seal-io/walrus/pkg/apis.(*Server).Setup.RequestThrottling.func3
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_flowcontrol.go:86 (0x565f75b)
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174 (0x10acfbb)
github.com/seal-io/walrus/pkg/apis.(*Server).Setup.I18n.func6
[...]

PR #2179, opened 2024-03-05
fix: preview draft state

alexcodelf:
**Problem:**

- draft resource does not have state record will cause panic when start or upgrade.
- progress task will mark incorrect resource state to deploying.
- rollback preview does not take effect

**Solution:**

- create resource state for draft resource
- remove resource state transitioning as run will handle resource state update.
- fix preview for rollback

**Related Issue:**
#2175 #2172 #2167

### DIFF
--- a/pkg/scheduler/resource/relationship_check_task.go
+++ b/pkg/scheduler/resource/relationship_check_task.go
@@ -124,8 +124,7 @@ func (in *RelationshipCheckTask) applyResources(ctx context.Context) error {
 			continue
 		}
 
-		// Deploy.
-		err = in.deployResource(ctx, res)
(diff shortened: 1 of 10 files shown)


pkg/scheduler/resource/relationship_check_task.go
```diff
@@ -292,14 +292,6 @@ func (in *RelationshipCheckTask) checkDependants(ctx context.Context, res *model
 }
 
 func (in *RelationshipCheckTask) deployResource(ctx context.Context, entity *model.Resource) error {
```
gitlawr, line 294: Do we still need this function?

gitlawr (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #1611, opened 2023-11-25
View logs from the app container by default, not the init container


linyinli:
View logs from the app container by default, not the init container:
<img width="1504" alt="image" src="https://github.com/seal-io/walrus/assets/42955482/28ff4fce-db09-4bca-ad80-7a0bd2e388fe">

hibig:
The data returned in the API  ```projects/{projectID}/environments/{environmentID}/resources/{resourceID}/components```  requires the value of ```resource.name```.



![Image](https://github.com/seal-io/walrus/assets/24664190/a583eebe-f7f1-4f9d-9ff0-6f23472a1431)

Finenyaco:
**Environment**
- Walrus version: main 1a7081b174c9781533116ce301abd5bc46d4043a
- UI version: dev-71a5388

**Test Result:** pass


<img width="485" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/5e011e9b-32f0-424d-bc64-a3961d2f6dab">

PR #2185, opened 2024-03-06
feat: view logs from the run container before the init container

shanewxy:
**Problem:**

The init container is returned before the run container when getting keys, which makes the init container the default one in the logs view.

**Solution:**

Sort the containers by their types in the order run<init<ephemeral to show the run container at first.

**Related Issue:**
#1611


gitlawr (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #2191, opened 2024-03-06
Creating resources using template kubernetes-containertask fails: Panic: parse index key failed


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Create a resource using template `builtin/kubernetes-containertask`.
<img width="2402" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/3d47e9b7-eb64-48aa-a980-4a763cfa425f">

2. Check the deployment result.

**Result**

1. Deploy failed.
<img width="889" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/eb283107-acd3-4787-9498-2c72980ecdef">

2. Check Walrus logs.
```
2024-03-06T08:37:01.482Z	D	api	200 | HTTP/2.0 |        0 B |        0 B |    2.348112ms |   192.168.117.1 | GET     /v1/projects/503617565914854658/environments/503617565914854658/resources/503626206818628866/runs/503626206835406082/terraform-states
[...]

Issue #2171, opened 2024-03-05
Clarify the 'Edited' label for resoruces


Finenyaco:
**Is your enhancement related to a problem? Please describe.**

The "Edited" label may also appear when there are no changes made to the resource configuration:

1. During deployment, the "Edited" label will be displayed first, and then it will disappear automatically after the automatic apply.
<img width="2208" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/6b595b88-a808-411f-9278-2a6c921332ef">

2. During the stop/start process of a resource, the "Edited" label is also displayed.
<img width="2191" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/c0855403-ea6a-4e72-b3b8-fab351611bca">

3. Save the stopped resource as a draft without modifying any configurations.
<img width="2238" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/5f353646-6097-4d56-89bf-ba3ce6b28733">

4. Which pages require displaying the "edited" label？Currently, only the environment resource list page displays it.

**Environment**
- Walrus version: main 7bb79389ab0436969f4c4214e38918ee6c6e9070
- UI version: dev-1be55ba

Finenyaco:
**Environment**
- Walrus version: main 96792e2d3b2c3b1eae368faeb3c2690f91fbffc5
- UI version: dev-040e825

**Test Result:**

- Remove `Edited` label.
- Add `Latest Run Status` on resource list.

<img width="2468" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/bf51938d-36a5-48c0-adfd-dc744f3fe5cf">

Issue #2155, opened 2024-03-04
Issues with stopping/starting resources


Finenyaco:
**Describe the bug**

**Steps to reproduce**

Issue 1:  Failed to start resource in the stopped state.
1. Create a resource.
2. Stop the resource. 
3. Start the resource.

Issue 2: Failed to stop an environment when some resources are in stopped state.
1. Create two resources in an environment.
2. Stop one of the resource.
3. Stop the environment.

**Test Result:**

**Result**

Issue 1: 
<img width="2460" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/968805ae-7d44-4fec-afca-c56a2b77e10f">

```
2024-03-04T05:07:27.783Z	E	api	error requesting POST /v1/projects/502909006822567170/environments/502909598756302082/resources/503304090328034562/start: 1 error occurred:
	* cannot start resource "nginx": in "Stopped" status


github.com/seal-io/walrus/pkg/apis/runtime.erroring
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_error.go:45
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.recovering
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_recovery.go:39
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.observing
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_observation.go:89
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/gin-gonic/gin.(*Engine).handleHTTPRequest
	github.com/gin-gonic/gin@v1.9.1/gin.go:620
github.com/gin-gonic/gin.(*Engine).ServeHTTP
	github.com/gin-gonic/gin@v1.9.1/gin.go:576
github.com/seal-io/walrus/pkg/apis/runtime.(*Router).ServeHTTP
	github.com/seal-io/walrus/pkg/apis/runtime/router.go:194
net/http.serverHandler.ServeHTTP
	net/http/server.go:2938
net/http.initALPNRequest.ServeHTTP
	net/http/server.go:3546
net/http.(*http2serverConn).runHandler
	net/http/h2_bundle.go:6193
[...]

PR #2193, opened 2024-03-06
fix: clearify resource and run status

alexcodelf:
**Problem:**

- The current resources and run status are unclear, and the `edited` of resource meanings are confusing.
- An error occurred while parsing the state.

**Solution:**

- Remove `is_modified` as we will show latest runs in resource list.
- Status of resource `Progressing` will no longer represents *waiting to deploy*, instead of run status pending to handle this.
https://github.com/seal-io/walrus/issues/2175#issuecomment-1978558857

    For the case
    1. preview is **false**
        create/update/start/rollback will mark the status of resource as `Deploying`, stop will marked as `Stopping` and delete will marked as `Deleting`
    2. preview is **true**
        resource status will not be changed, until the preview plan is applied
    3. dependency
        If a resource is dependent on another resource, the status pattern will follow the previous 1 or 2. The difference is that the run status will be marked as `Pending` with message `Waiting for dependent resources to be ready: xxx`. 
        Scheduler will using pending `ResourceRun` to get resource with dependenies to do apply jobs.
    4. resource list will show lasted run status.
    5. error
        update logic of resource and run when dependency is not worked as expected.

- Using reflect to get index key type to parse the right terraform module index.

**Related Issue:**
#2191 #2175 #2155  #2171


gitlawr (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #1461, opened 2023-11-18
Service draft cannot be saved as a draft again after the first save


linyinli:
**Describe the bug**
Service draft cannot be saved as a draft again after the first save.

**Steps to reproduce**
1. Go to 'Projects-Environments-Services/Resources' ,create a service and save as draft;
2. Edit the service draft and save as draft again.

**Result**
It will be deployed automatically.

**Expected behavior**
Service draft can be saved as a draft again.

**Environment**
- Walrus version: v0.4.0-rc2

alexcodelf:
Env: walrus-server: 89cdec08768c3a8ed0dbf1490e04f03ad3ecf630
Result: PASS. Draft resource can be saved as a draft again.
<img width="2922" alt="image" src="https://github.com/seal-io/walrus/assets/85595852/f01c4b47-738d-4d12-b083-a442dcddba42">

Finenyaco:
**Environment**
- Walrus version: main 1f87819b63d844001b3353645ed96d110406783c
- UI version: dev-71a5388

**Test Result:**

This issue can be reproduced in main.

<img width="2444" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/ce9825ac-6646-4d9b-b9fe-32290f24f98a">


```
2024-03-06T11:53:59.563Z	E	api	error requesting PUT /v1/projects/503645663356034050/environments/503645663356034050/resources/503646001031060482/upgrade: 1 error occurred:
	* cannot save resource "draft" as draft: in "" status


github.com/seal-io/walrus/pkg/apis/runtime.erroring
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_error.go:45
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.recovering
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_recovery.go:39
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/seal-io/walrus/pkg/apis/runtime.observing
	github.com/seal-io/walrus/pkg/apis/runtime/middleware_observation.go:89
github.com/gin-gonic/gin.(*Context).Next
	github.com/gin-gonic/gin@v1.9.1/context.go:174
github.com/gin-gonic/gin.(*Engine).handleHTTPRequest
[...]

PR #2195, opened 2024-03-06
fix: remove validation in operation, update message

alexcodelf:
**Problem:**

- validation status failed as request of resource does not pass status.
- modify dependency error message
- query stopping resource condition error

**Solution:**

- remove validation in operation upgrade
- update query conditon in  resource run dependant check

**Related Issue:**

#1461 #2155

### DIFF
--- a/pkg/resources/operation.go
+++ b/pkg/resources/operation.go
@@ -238,11 +238,6 @@ func upgrade(
 	}
 
 	if opts.Draft {
-		// Only stopped resource can save as draft.
-		if !status.ResourceStatusStopped.IsTrue(entity) {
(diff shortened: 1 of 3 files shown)


pkg/resources/operation.go
```diff
@@ -238,11 +238,6 @@ func upgrade(
 	}
 
 	if opts.Draft {
-		// Only stopped resource can save as draft.
-		if !status.ResourceStatusStopped.IsTrue(entity) {
```
gitlawr, line 242: Why remove it?

alexcodelf: The input resource comes from req.Model() which status is empty, can not using this to validation.

gitlawr (dismissed):
LGTM

aiwantaozi (approved):
LGTM

Issue #2196, opened 2024-03-06
After cloning and previewing resource with dependencies, resource status changes to 'DeployFailed' immediately


Finenyaco:
**Describe the bug**

After cloning and previewing resources with dependencies, the resource status immediately changes to "DeployFailed."

**Steps to reproduce**
1. Create resources by importing 'nextcloud-mysql' yaml file.
<img width="1033" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/c6b0318e-da7c-4b0e-ba36-ede004729c6d">

2. Clone the environment in step 1 and preview.
3. Check the resource list after clone.

**Result**

1. The resource status of 'nextcloud' changed to `DeployFailed` immediately.

<img width="2392" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/e4b08a08-ab93-4c59-b4f8-adb2944e40e5">

**Expected behavior**

**Environment**
- Walrus version: main 96792e2d3b2c3b1eae368faeb3c2690f91fbffc5
- UI version: dev-040e825

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 54608149edeebd87ed53da7b8e51f384ae386650
- UI version: dev-b0233e3

**Test Result:** pass
<img width="2476" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/8e6fccbf-780b-46e3-8c5e-65ed22179460">

PR #2197, opened 2024-03-06
fix: dependency stop resource query wrong

alexcodelf:
**Problem:**

- A continuous fix for #2195, as dependant query get wrong resources.  https://github.com/seal-io/walrus/issues/2155#issuecomment-1980916339
- Remove useless resource information for resource run list API
- Remove inactive check

**Solution:**

Correct query of dependant query.

**Related Issue:**
#2196 #2195


shanewxy (approved):
LGTM

Issue #2198, opened 2024-03-07
The dashboard lacks statistics for certain resource run status


Finenyaco:
**Describe the bug**

**Steps to reproduce**
`Planning`, `Planed`, `Canceled` and `Pending` status are not being included in the dashboard statistics.

<img width="2475" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/1a8936f9-ff98-46be-bb49-f40ac80b86de">

**Expected behavior**

**Environment**
- Walrus version: main 96792e2d3b2c3b1eae368faeb3c2690f91fbffc5
- UI version: dev-040e825

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 39304a76ab4b41f764ee0087ebaee793437efca4
- UI version: dev-ae07951

**Test Result:** pass

<img width="2466" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/8c014a13-24c7-40db-966d-60322c1a45cf">

PR #2201, opened 2024-03-07
fix: dashboard lacks statistics for some status

shanewxy:
**Problem:**

Planning, Planned, Canceled, and Pending statuses are not included in the dashboard statistics.

**Solution:**

Ignore plan runs and keep only succeed, running, and failed statuses in Deployments to be consistent with statistics.

**Related Issue:**
#2198

### DIFF
--- a/pkg/apis/dashboard/extension.go
+++ b/pkg/apis/dashboard/extension.go
@@ -238,7 +238,13 @@ func getResourceRunStatusStats(
 			statMap[t].Failed = c.Count
 		case status.ResourceRunSummaryStatusSucceed:
 			statMap[t].Succeeded = c.Count
-		case status.ResourceRunSummaryStatusRunning:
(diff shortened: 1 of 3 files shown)


alexcodelf:
What's the User Experience after we introduce these statistics status.

alexcodelf (dismissed):
LGTM

gitlawr:
We show `Deployments` in the UI. Should we ignore plan runs given that they are not real deployments?

shanewxy:
> We show `Deployments` in the UI. Should we ignore plan runs given that they are not real deployments? Otherwise we may need to change deployments to runs

Updated. Ignore plan runs and keep only succeed, running, and failed statuses in `Deployments` to be consistent with statistics.

gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

PR #2207, opened 2024-03-07
fix: sensitive data shows in component changes

alexcodelf:
**Problem:**

- sensitive data shows for components changed
- show duplicated error when resource created

**Solution:**

- mask sensitive data with marked leaf keys
- return directly when resource created but error occurs.

**Related Issue:**
#1003 https://github.com/seal-io/internal/issues/16


pkg/dao/types/resource_run.go
```diff
@@ -134,3 +124,152 @@ type Change struct {
...
+		c.Type = ResourceComponentChangeTypeUpdate
+	case c.Actions.Delete():
+		c.Type = ResourceComponentChangeTypeDelete
+	case c.Actions.NoOp(), c.Actions.Read():
+		c.Type = ResourceComponentChangeTypeNoChange
+	}
+
+	c.Change.Before = patchLeaf(c.Change.Before, c.Change.BeforeSensitive, "<sensitive>")
+	c.Change.After = patchLeaf(c.Change.After, c.Change.AfterSensitive, "<sensitive change>")
+	c.Change.After = patchLeaf(c.Change.After, c.Change.AfterUnknown, "<after known>")
```
shanewxy, line 145: Why use <after known> instead of <after unknown>?

gitlawr, line 145:
```
<sensitive value>
<sensitive value(changed)>
<known after apply>
```

alexcodelf: Let's clear it with `known after apply`

alexcodelf: updated.

gitlawr (approved):
LGTM

shanewxy (approved):

Issue #2204, opened 2024-03-07
Failed to get google redis instance status


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. New a resource to create a google redis instance.

**Result**

1. Unable to get google redis instance status.
<img width="2384" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/4f80ff25-1431-45ba-8886-ddca2d1f5805">

<img width="817" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/0aca36f2-555b-4411-bc10-737b71f25a30">

2. Check Walrus logs.
```
2024-03-07T11:49:10.759Z	D	api	200 | HTTP/2.0 |     11 KiB |        0 B |   12.932704ms |       127.0.0.1 | PUT     /v1/projects/503740965677225986/environments/503751901871921154/resources/503789528654338050/runs/503789528671115266/terraform-states
2024-03-07T11:49:13.330Z	D	signal.handler	reaped child process 45909
2024-03-07T11:49:13.583Z	E	api.resource-run	error stating resources: error get resource google_redis_instance:projects/project-xxx/locations/asia-east2/instances/default-azure-google-redis-666pahet2w from 503773208953018370: failed to get google resource google_redis_instance projects/project-xxx/locations/asia-east2/instances/default-azure-google-redis-666pahet2w: rpc error: code = InvalidArgument desc = Malformed name: 'projects/project-xxx/locations/asia-east2/instances/projects/project-xxx/locations/asia-east2/instances/default-azure-google-redis-666pahet2w'
error details: name = BadRequest field = projects/project-xxx/locations/asia-east2/instances/projects/project-xxx/locations/asia-east2/instances/default-azure-google-redis-666pahet2w desc = Invalid resource pathname in resource URL path
github.com/seal-io/walrus/pkg/apis/resourcerun.reconcileResourceComponents
	github.com/seal-io/walrus/pkg/apis/resourcerun/extension.go:411
github.com/seal-io/walrus/pkg/apis/resourcerun.manageResourceComponentsAndEndpoints.func2
	github.com/seal-io/walrus/pkg/apis/resourcerun/extension.go:336
github.com/alitto/pond.(*WorkerPool).executeTask
	github.com/alitto/pond@v1.8.3/pond.go:454
github.com/alitto/pond.worker
	github.com/alitto/pond@v1.8.3/worker.go:32
[...]

PR #2208, opened 2024-03-07
fix: failed to get google redis instance status

shanewxy:
**Problem:**

The name of the Google Redis instance is already the full identifier, so no need to prepend the extra path.

**Solution:**

Use the name directly to retrieve the Redis instance status.

**Related Issue:**
#2204


gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #2205, opened 2024-03-07
Incorrect format for google_service_networking_connection instance name


Finenyaco:
**Describe the bug**

In the name of the google_service_networking_connection instance, the "/" is displayed as "%2F".

**Steps to reproduce**
1. New a resource to create a google mysql instance and do not specific the vpc_id.
2. Check the resource components list.

**Result**

1. The "/" is displayed as "%2F".

<img width="2268" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/e075b203-360f-4462-b62d-c941b508b8ed">

<img width="837" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/f436af15-98b2-42cf-91bc-3717ae4f7d14">

**Expected behavior**

**Environment**
- Walrus version: v0.6.0-rc1
- template version: google-mysql@v0.1.1

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main fa6e56aca10e61259243c938130c8e63dd19ba27
- UI version: dev-4218197

**Test Result:** pass

<img width="877" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/59caa688-8205-43cd-bc78-dd9b58667bd7">

PR #2217, opened 2024-03-08
fix: google_service_networking_connection name format

shanewxy:
**Problem:**

Google provider returns URL-encoded representation of google_service_networking_connection name in terraform state, which causes "%2F" to be displayed on walrus UI.

**Solution:**

Decode the instance ID to avoid incorrect format from the provider.

**Related Issue:**
#2205


gitlawr (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #2166, opened 2024-03-04
Resource label change triggered re-deployment


pengjiang80:
**Describe the bug**
Resource label change will trigger re-deployment 

**Steps to reproduce**
1. Create a create.
2. Modify the resource label, for example, change the stoppable from true to false.

**Result**
A redeployment will be triggered. 

**Expected behavior**
Resource label is metadata instead of resource spec which should not trigger re-deployment. 

**Environment**
- Walrus version: v0.5.1

PR #2219, opened 2024-03-10
fix: update metadata should not trigger deployment

alexcodelf:
**Problem:**

- resource metadata update with attributes update should not create deployment

**Solution:**

Only update metadata.

**Related Issue:**
#2166


alexcodelf:
Fix this after refactor framework. Let's handle this in v0.7.0

Issue #2213, opened 2024-03-08
Errors occurred during the initialization of the built-in resource definition


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Refresh the built-in catalog.
2. Check Walrus logs.

**Result**

1. Returned some errors.
```
2024-03-08T06:12:57.453Z	E	builtin.resource-definitions	failed to create matching rule for builtin alibaba-redis: model: template_version not found
github.com/seal-io/walrus/pkg/resourcedefinitions.newResourceDefinition
	github.com/seal-io/walrus/pkg/resourcedefinitions/builtin.go:104
github.com/seal-io/walrus/pkg/resourcedefinitions.SyncBuiltinResourceDefinitions
	github.com/seal-io/walrus/pkg/resourcedefinitions/builtin.go:66
reflect.Value.call
	reflect/value.go:596
reflect.Value.Call
	reflect/value.go:380
github.com/seal-io/walrus/utils/bus.bus.Publish
	github.com/seal-io/walrus/utils@v0.0.0-00010101000000-000000000000/bus/bus.go:92
github.com/seal-io/walrus/utils/bus.Publish
	github.com/seal-io/walrus/utils@v0.0.0-00010101000000-000000000000/bus/bus.go:143
github.com/seal-io/walrus/pkg/bus/builtin.Notify
	github.com/seal-io/walrus/pkg/bus/builtin/bus.go:16
github.com/seal-io/walrus/pkg/catalog.UpdateStatusWithSyncErr
	github.com/seal-io/walrus/pkg/catalog/catalog.go:257
github.com/seal-io/walrus/pkg/catalog.catalogSyncer.Do.func1
	github.com/seal-io/walrus/pkg/catalog/catalog.go:230
github.com/alitto/pond.(*WorkerPool).executeTask
	github.com/alitto/pond@v1.8.3/pond.go:454
github.com/alitto/pond.worker
	github.com/alitto/pond@v1.8.3/worker.go:32
2024-03-08T06:12:57.453Z	E	builtin.resource-definitions	failed to create matching rule for builtin aws-redis: model: template_version not found
github.com/seal-io/walrus/pkg/resourcedefinitions.newResourceDefinition
	github.com/seal-io/walrus/pkg/resourcedefinitions/builtin.go:104
github.com/seal-io/walrus/pkg/resourcedefinitions.SyncBuiltinResourceDefinitions
[...]

PR #2220, opened 2024-03-11
fix: errors occurred while creating the built-in resource definitions

aiwantaozi:
**Problem:**
As the template version name doesn't consistently follow the format fmt.Sprintf("%s/%s-%s", catalog.BuiltinCatalog().Name, connectorType, resourceType), such as in cases like https://github.com/walrus-catalog/terraform-alicloud-rds-mysql, the generated name might encounter issues when attempting to query the template version.

**Solution:**
Use the templateID to query the template version.

**Related Issue:**
#2213

### DIFF
--- a/pkg/resourcedefinitions/builtin.go
+++ b/pkg/resourcedefinitions/builtin.go
@@ -2,20 +2,18 @@ package resourcedefinitions
 
 import (
 	"context"
-	"fmt"
+	"sort"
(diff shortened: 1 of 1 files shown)


shanewxy (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #2216, opened 2024-03-08
The resource cannot be edited or deployed while it is in the Deploying state


linyinli:
**Describe the bug**
The resource cannot be edited or deployed while it is in the deploying state.

**Steps to reproduce**
1. Create a resource, click Save and Preview;
2. Click Apply after planned.

**Result**
When run fails, the resource is stuck in a deploying state, and cannot edit or deploy it.
<img width="1811" alt="image" src="https://github.com/seal-io/walrus/assets/42955482/d07ef87d-ff4e-4899-9389-6d5ea0d49912">

**Expected behavior**
When run fails,  resource goes into DeployFailed state.

**Environment**
- Walrus version: v0.6.0-rc1

Finenyaco:
This seems to be related to the issue that sometimes resources get stuck in a Deploying state https://github.com/seal-io/walrus/issues/814#issuecomment-1982279057.

alexcodelf:
Can not find a way to reproduce it, add log print to trace it.

Finenyaco:
**Environment**
- Walrus version: main 61070fa9f82dd3630b4aadb8626245dcef93e722
- UI version: dev-f26d810

**Test Result:** Reproduce in main.

<img width="2428" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/6d0ee01b-51ae-4442-bf90-e0fc59027d50">

```
2024-03-13T02:38:04.760Z	D	api	200 | HTTP/2.0 |     10 KiB |        0 B |   12.806525ms |       127.0.0.1 | PUT     /v1/projects/504341382106710018/environments/504465041026711554/resources/504602230469427202/runs/504602230502981634/terraform-states
2024-03-13T02:38:05.424Z	D	task	executed	{"createBy": "ResourceComponentLabelApplyCronExpr", "duration": "5.409238732s"}
2024-03-13T02:38:07.047Z	D	signal.handler	reaped child process 1917
2024-03-13T02:38:08.074Z	I	k8sctrl.deployer.tf	succeed	{"resource-run": "504602230502981634"}
2024-03-13T02:38:08.082Z	D	resource-run.syncer	resource azure-redis "create" run 504602230502981634 status "Succeeded"
2024-03-13T02:38:08.082Z	D	resource-run.syncer	set resource azure-redis status to "Preparing"
[...]

PR #2224, opened 2024-03-11
fix: upgrade resource and run created with error

alexcodelf:
**Problem:**

- When resource upgrade succeed but run job failed, it should not return error in Upgrade API, users should check logs in runs. https://github.com/seal-io/walrus/issues/1003#issuecomment-1987511659
- Add some debug information in resource run syncer.

**Solution:**

- Skip upgrade error when run created for resource upgrade.
- Print debug log for tracing status issues. #2216

**Related Issue:**
#1003

### DIFF
--- a/pkg/apis/resource/extension.go
+++ b/pkg/apis/resource/extension.go
@@ -14,6 +14,7 @@ import (
 	pkgcomponent "github.com/seal-io/walrus/pkg/resourcecomponents"
 	pkgresource "github.com/seal-io/walrus/pkg/resources"
 	tfparser "github.com/seal-io/walrus/pkg/terraform/parser"
+	"github.com/seal-io/walrus/utils/errorx"
 	"github.com/seal-io/walrus/utils/log"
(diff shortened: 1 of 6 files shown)


gitlawr (dismissed):
LGTM

pkg/apis/resource/extension.go
```diff
@@ -26,18 +27,19 @@ func (h Handler) RouteUpgrade(req RouteUpgradeRequest) error {
...
 	})
-	if err != nil {
-		return err
+	// NB(alex): If resource run created successfully, users can check the resource run status and logs.
+	if run != nil {
+		return nil
 	}
 
-	return nil
+	return errorx.Wrap(err, "failed to upgrade resource")
```
shanewxy, line 42: Will it still return this error message when err == nil and run == nil?

shanewxy (approved):
LGTM

gitlawr (approved):

Issue #2206, opened 2024-03-07
Failed to get deployment logs: the server could not find the requested resource


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. New a resource to deploy a google mysql instance.
2. View the deployment logs.

**Result**

1. Failed to get deployment logs.
<img width="878" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/d4e54149-5a8b-47a1-a769-6c9f8f0a2b0d">

<img width="888" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/ffdba5b0-6408-44fd-b414-26cef07fcdb2">

```
2024-03-07T12:20:01.769Z	D	api	200 | HTTP/2.0 |        0 B |        0 B |  1.415880115s |   192.168.216.1 | US      /v1/projects/503740965677225986/environments/503751901871921154/resources/503776165887006722/components?watch=true
2024-03-07T12:20:02.026Z	E	k8sctrl.deployer.tf	failed to get job pod logs	{"error": "the server could not find the requested resource (get pods tf-job-plan-503792589103749122-dlkm5)", "resource-run": "503792589103749122"}
github.com/seal-io/walrus/utils/log.logrSinker.Error
	github.com/seal-io/walrus/utils@v0.0.0-00010101000000-000000000000/log/logr.go:67
github.com/go-logr/logr.Logger.Error
	github.com/go-logr/logr@v1.4.1/logr.go:301
github.com/seal-io/walrus/pkg/resourceruns/job.Reconciler.syncRunStatus
	github.com/seal-io/walrus/pkg/resourceruns/job/reconcile.go:86
github.com/seal-io/walrus/pkg/resourceruns/job.Reconciler.Reconcile
	github.com/seal-io/walrus/pkg/resourceruns/job/reconcile.go:47
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Reconcile
	sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:119
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler
	sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:316
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	sigs.k8s.io/controller-runtime@v0.16.3/pkg/internal/controller/controller.go:266
[...]

PR #2225, opened 2024-03-11
fix: should not set apply status base on plan job

aiwantaozi:
**Problem:**
Apply will generate two jobs, while the plan job success and it updates the run status. Should update the run status until the apply job succeeds.

**Solution:**
Only update the job-related status

**Related Issue:**
 #2206

### DIFF
--- a/pkg/resourceruns/job/reconcile.go
+++ b/pkg/resourceruns/job/reconcile.go
@@ -2,6 +2,7 @@ package job
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -12,7 +13,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	ctrl "sigs.k8s.io/controller-runtime"
-	client "sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	runbus "github.com/seal-io/walrus/pkg/bus/resourcerun"
 	"github.com/seal-io/walrus/pkg/dao/model"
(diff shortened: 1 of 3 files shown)


gitlawr (dismissed):
LGTM

pkg/resourceruns/status/status.go
```diff
@@ -62,11 +62,16 @@ func IsStatusSucceeded(run *model.ResourceRun) bool {
 
 // SetStatusFalse sets the status of the resource run to false.
 func SetStatusFalse(run *model.ResourceRun, errMsg string) {
+	jobType := types.RunTaskTypeApply
+	if run.Preview {
```
alexcodelf, line 66: can not use preview to identify the jobType, as every run has plan tasks.

pkg/resourceruns/job/reconcile.go
alexcodelf, line 1: Shall we check the job type with labels to determine the job is plan or not, combine with the current run status to do the run status decision?

alexcodelf (approved):
LGTM

pkg/resourceruns/job/reconcile.go
```diff
@@ -175,3 +181,66 @@ func (r Reconciler) cleanPlanFiles(run *model.ResourceRun) {
...
+func taskTypeFromK8sJob(job *batchv1.Job) types.RunJobType {
+	switch {
+	case strings.Contains(job.Name, string(types.RunTaskTypeApply)):
+		return types.RunTaskTypeApply
+	case strings.Contains(job.Name, string(types.RunTaskTypePlan)):
+		return types.RunTaskTypePlan
+	case strings.Contains(job.Name, string(types.RunTaskTypeDestroy)):
+		return types.RunTaskTypeDestroy
+	default:
+		return "unknown"
```
gitlawr, line 244: Do we have labels to differentiate the task type?

aiwantaozi: Now we don't have, but we can add.

gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

PR #2227, opened 2024-03-11
feat: introduce new warning status summary

alexcodelf:
**Problem:**

As the status and color is not clear in walrus. We need introduce a new color to represent a `break point` status.
We make following status color desision, after this PR merged.
| Stage        | Color  |
| ------------ | ------ |
| Transitioning| Blue   |
| Inactive     | Gray   |
| Warning      | Yellow |
| Succeed      | Green  |
| Failed       | Red    |

@hibig UI need to be modified synchronously. Be careful with the color changed. Warning represents pause status.

**Related Issue:**

### DIFF
--- a/pkg/dao/types/status/converter.go
+++ b/pkg/dao/types/status/converter.go
@@ -13,19 +13,26 @@ type (
 // status not in the normalStatus and errorStatus will set transitioning to true.
 //   - `normalStatus` specifies the normal status list, won't change the error and transition,
 //     while summary status is in the normal status list.
+//   - `warningStatus` specifies the warn status list, set the warn to true,
+//     while the summary status is in the warn status list.
 //   - `inactiveStatus` specifies the inactive status list, set the transitioning to false,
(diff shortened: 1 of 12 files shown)


pkg/dao/types/status/converter.go
```diff
@@ -58,24 +67,34 @@ func (w *converter[T]) Convert(sm, msg string) *Status {
...
 		st.Error = false
 		st.Inactive = true
+		st.Warning = false
 		st.Transitioning = false
 	case isNormal:
 		st.Error = false
 		st.Inactive = false
+		st.Warning = false
+		st.Transitioning = false
+	case isWarn:
```
gitlawr, line 89: isWarning

aiwantaozi (approved):
LGTM

pkg/dao/types/status/converter.go
alexcodelf: updatede

gitlawr (approved):
LGTM

Issue #2222, opened 2024-03-11
Failed to load some templates from terraform-aws-modules


aiwantaozi:
**Describe the bug**
Failed

**Steps to reproduce**
1. create a catalog with address https://github.com/terraform-aws-modules
2. check the created templates
3. found some templates failed to load
<img width="1395" alt="image" src="https://github.com/seal-io/walrus/assets/5936271/9696b064-a8c1-419e-9d73-507cb9523741">

4. check the template error
<img width="731" alt="image" src="https://github.com/seal-io/walrus/assets/5936271/e40948ef-3d65-4c9e-8b40-abe4c1615a4d">

**Expected behavior**
Load templates without error.

**Environment**
- Walrus version:
61fe2d23f46237867cf199c44a3275ee71b7805d

**Additional context**

Finenyaco:
**Environment**
- Walrus version: main 34e8294e82578691027f04164b18f1ec65137ce9
- UI version: dev-b878771

**Test Result:**
Some tags only have a README.md file will be imported as well.
<img width="2468" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/018df0c5-5a20-45c3-9272-558a0e7d6fb7">
<img width="2558" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/a6bc1dfc-d549-4b58-a0e0-529fcf92203c">

Finenyaco:
**Environment**
- Walrus version: main 0a0f275bc92d1f7f379e5d97337b3362096fe508
- UI version: dev-87c5bd5

**Test Result:** pass

PR #2228, opened 2024-03-11
fix: terraform module doesn't always include providers, variables and…

aiwantaozi:
… outputs

**Problem:**
Failed to load some templates from terraform-aws-modules.

**Solution:**
Remove must exist providers, variables or outputs validation, as some module may not include them.

**Related Issue:**
#2222


shanewxy (approved):
LGTM

PR #2229, opened 2024-03-12
fix: skip load tags without terraform configuration files

aiwantaozi:
**Problem:**
Load tags without terraform configuration files

**Solution:**
Skip load tags without terraform configuration files

**Related Issue:**
#2222

### DIFF
--- a/pkg/templates/loader/terraform_loader.go
+++ b/pkg/templates/loader/terraform_loader.go
@@ -66,6 +66,10 @@ func (l *TerraformLoader) Load(
 
 // loadMod load the terraform module.
 func (l *TerraformLoader) loadMod(rootDir string) (*tfconfig.Module, error) {
+	if !tfconfig.IsModuleDir(rootDir) {
+		return nil, fmt.Errorf("no terraform configuration files found")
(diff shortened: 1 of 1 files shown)


pkg/templates/loader/terraform_loader.go
```diff
@@ -66,6 +66,10 @@ func (l *TerraformLoader) Load(
 
 // loadMod load the terraform module.
 func (l *TerraformLoader) loadMod(rootDir string) (*tfconfig.Module, error) {
+	if !tfconfig.IsModuleDir(rootDir) {
+		return nil, fmt.Errorf("no terraform configuration files found")
```
gitlawr, line 70: How do we handle the error at the end, should it be an error?

aiwantaozi: For catalog, walrus server will log these errors and skip this version, continually importing other versions, no error in the template status; for imported template with URL, will show this error in the template status.

gitlawr (approved):
LGTM

alexcodelf (approved):
LGTM

Issue #2202, opened 2024-03-07
Deploying resource failed: Failed to load "plan.out" as a plan file


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. New a resource and deploy.
2. Check the deployment result.

**Result**

1. Failed to deploy. Haven't found a method to reproduce it yet.

<img width="902" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/be51d1f5-f703-4626-a220-55af9855fb4e">

```
024-03-07T03:53:26.948Z	D	deployer.tf	k8s job tf-job-plan-503742632728516610 created
2024-03-07T03:53:26.950Z	D	api	201 | HTTP/1.1 |      168 B |      879 B |  159.326663ms |   192.168.216.1 | POST    /v1/projects/503742627393361922/environments/503742627410139138/resources
2024-03-07T03:53:27.009Z	D	api	201 | HTTP/1.1 |      210 B |      941 B |   54.034042ms |   192.168.216.1 | POST    /v1/projects/503742627393361922/environments/503742627410139138/resources
2024-03-07T03:53:27.027Z	D	api	201 | HTTP/1.1 |      243 B |     1007 B |   11.998102ms |   192.168.216.1 | POST    /v1/projects/503742627393361922/environments/503742627410139138/resources
2024-03-07T03:53:27.037Z	D	api	200 | HTTP/1.1 |        0 B |    1.5 KiB |     3.39513ms |   192.168.216.1 | GET     /v1/projects/503742627393361922/environments/503742627410139138/graph
2024-03-07T03:53:27.084Z	D	deployer.tf	k8s job tf-job-plan-503742633131169794 created
2024-03-07T03:53:27.086Z	D	api	201 | HTTP/1.1 |      168 B |      857 B |   36.837474ms |   192.168.216.1 | POST    /v1/projects/503742627393361922/environments/503742627443693570/resources
2024-03-07T03:53:27.471Z	D	api	200 | HTTP/2.0 |        0 B |        0 B |    2.402671ms |       127.0.0.1 | GET     /v1/projects/503742627393361922/environments/503742627410139138/resources/503742632711739394/runs/503742632728516610/terraform-states
2024-03-07T03:53:27.515Z	D	api	200 | HTTP/2.0 |        0 B |        0 B |    2.656546ms |       127.0.0.1 | GET     /v1/projects/503742627393361922/environments/503742627443693570/resources/503742633114392578/runs/503742633131169794/terraform-states
[...]

PR #2230, opened 2024-03-12
fix: show s3 bucket already owned error

alexcodelf:
- provide region query for s3 config
- skip s3 bucket already created error

**Problem:**

- when do multi deployment runs, jobs will check and create bucket for the s3 service. But once one of them create the bucket successfully, the other try to create the same bucket will return already create error, we should ignore the error.
- region of s3 is not configurable, using the default `ap-north-1` as the region
- when plan set failed, the job still runs as success

**Solution:**

- ignore the `bucket already exit err`
- support config s3 region
- add curl params `--fail-with-body` to show error and exit when http response code is greater than 400, https://curl.se/docs/manpage.html
image

**Related Issue:**
#2202


gitlawr (approved):
LGTM

Issue #2150, opened 2024-03-01
Can't sync Azure terraform modules due to massive number of repos


pengjiang80:
Version: v0.5.1
Step to reproduce:
1. Go to operations - catalog and add https://github.com/azure. This org contains terraform azure modules and other repos, the total repo number is more than 2K.
2. A standalone docker run installation will fail to sync catalog with the following error:
error syncing "azure-catalog:azure-sdk-for-java" of catalog "502862623004426243": error downloading '
[https://github.com/Azure/azure-sdk-for-java':](https://github.com/Azure/azure-sdk-for-java%27:)
/usr/bin/git exited with 128: Cloning into '/tmp/seal-template-Rk3bOmmnjX'... fatal: write error: No space left on device fatal: fetch-pack
3. The screenshots for catalog and docker space usage are as below:
![image](https://github.com/seal-io/walrus/assets/16434640/4df29f3a-3943-43bd-ba51-d7de62696d47)
![image](https://github.com/seal-io/walrus/assets/16434640/2d83bacc-6f23-45f8-8600-e40d419bbe0b)

PR #2233, opened 2024-03-12
fix: sync massive number repos catalog failed

alexcodelf:
**Problem:**

- Can't sync terraform modules due to massive number of repos

**Solution:**

- provider catalog regexp pattern to filter repos with name

**Related Issue:**
#2150


pkg/catalog/catalog.go
```diff
@@ -64,15 +65,24 @@ func getRepos(ctx context.Context, c *model.Catalog, ua string, skipTLSVerify bo
 		return nil, err
 	}
 
-	list := make([]*vcs.Repository, len(repos))
+	var nameReg *regexp.Regexp
+	if c.FilterPattern != "" {
+		nameReg = regexp.MustCompile(c.FilterPattern)
```
gitlawr, line 70: Avoid using MustCompile, it causes panic

alexcodelf: updated

pkg/apis/catalog/basic_view.go
```diff
@@ -40,6 +41,10 @@ func (r *CreateRequest) Validate() error {
 		return fmt.Errorf("unsupported catalog type %q", r.Type)
 	}
 
+	if _, err = regexp.Compile(r.FilterPattern); err != nil {
+		return err
```
gitlawr, line 45: How does the error look like? Should we make it more friendly

alexcodelf: image
If regexp is not vaild, this is how it looks like , shall we warp this error?

gitlawr: fmt.Errorf("invalid filter pattern: %w", err)

alexcodelf: Then the error will be:
```json
{
  "message": "invalid filter pattern: error parsing regexp: missing closing ]: `[email\\]@example.com`",
  "status": 400,
  "statusText": "Bad Request"
}
```

gitlawr (approved):
LGTM

PR #2234, opened 2024-03-12
fix: measure s3 ignore error failed

alexcodelf:
**Problem:**

- measuring ignore error condition failed, as the minio-go error is not match s3 sdk error

**Solution:**

- correct err condition, using code to identify the true error.

**Related Issue:**
#2022


shanewxy (approved):
LGTM

gitlawr (approved):
LGTM

PR #2235, opened 2024-03-13
fix: resource status stucking in deploying

alexcodelf:
**Problem:**

- resource sometimes will stuck in deploying. `Component status check` scheduler will select older resource that is in `deploying` status and overwrite status to db after the task finished.

**Solution:**

- query resource again and skip deploying status resource update in scheduler. As the scheduler may take a long time to complete. A better way is using CAS.

**Related Issue:**
#2216


gitlawr (approved):
LGTM

shanewxy (approved):
LGTM

Issue #2236, opened 2024-03-13
Index key of components is not int or float64 will be ignored


alexcodelf:
**Describe the bug**

Parsing resource components will ignore the index key that is not int or float64.

**Steps to reproduce**
1. Using azure redis module to create resources.
2. Watching logs.

**Result**

![image](https://github.com/seal-io/walrus/assets/85595852/e9b49691-a069-484e-87c4-99c82aa1ca19)

**Expected behavior**

- should parse the components without warning.

**Environment**
- Walrus version:  3f1df62c62428bb601867c035c5a23c6a2a49128

**Additional context**

alexcodelf:
The issue is just warning message, as it will not affect the component change.

Issue #2214, opened 2024-03-08
Unable to synchronize the azure redis instance status while it is being created


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. Add a template with source https://github.com/walrus-catalog-sandbox/terraform-azure-redis.
2. Using the template to create a  resource.
3. Go to the resource details page and check the component list.

**Result**

1. The run status was succeed. But it didn't show the  component while the redis instance was being created.
<img width="877" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/88da592c-bf2f-45e1-95c9-c73c64897623">

![image](https://github.com/seal-io/walrus/assets/26084756/8a55dc1a-8bf0-4063-8457-11808584bd45)

2. Component was synchronized successfully only when it was created successfully.
![image](https://github.com/seal-io/walrus/assets/26084756/0cc1fb1c-4829-4ef5-9bfc-4fff6c225f5f)

**Expected behavior**

**Environment**
- Walrus version:  v0.6.0-rc1

**Additional context**

shanewxy:
This seems to be a bug of the Azurerm provider. The Redis instance information inside the state file is not updated when deploying Redis directly through Terraform as well.

PR #2237, opened 2024-03-13
fix: parsing resource component will be skipped

alexcodelf:
Problem: 

- fix parse index key failed.

Solution: 

- parse index key as int when it is number.

#2236

### DIFF
--- a/pkg/terraform/parser/parser.go
+++ b/pkg/terraform/parser/parser.go
@@ -526,10 +526,10 @@ func ParseIndexKey(rs resourceState, is instanceObjectState) (string, error) {
 	}
 
 	if is.IndexKey != nil {
-		switch reflect.TypeOf(is.IndexKey).Kind() {
-		case reflect.String:
+		switch {
+		case IsString(is.IndexKey):
 			return strs.Join(".", rs.Module, rs.Type, rs.Name) + fmt.Sprintf("[\"%v\"]", is.IndexKey), nil
-		case reflect.Int, reflect.Float64:
+		case IsNumber(is.IndexKey):
 			return strs.Join(".", rs.Module, rs.Type, rs.Name) + fmt.Sprintf("[%v]", is.IndexKey), nil
 		default:
 			logger.Warnf("unsupported index key: %v", is.IndexKey)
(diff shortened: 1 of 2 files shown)


gitlawr:
Is it related to https://github.com/seal-io/walrus/issues/2214?

gitlawr:
What's the consequence when index parsing fails?

pkg/terraform/parser/parser.go
```diff
@@ -538,3 +538,14 @@ func ParseIndexKey(rs resourceState, is instanceObjectState) (string, error) {
 
 	return strs.Join(".", rs.Module, rs.Type, rs.Name), nil
 }
+
+// IsNumber returns true if the index key is a number.
+// https://github.com/hashicorp/terraform/blob/main/internal/states/statefile/version4.go#L114C1-L122C32.
+// As terraform indexKey treated as int key, assume all number types are int when parsing index key.
+func IsNumber(indexKey any) bool {
+	return reflect.TypeOf(indexKey).Kind() >= reflect.Int && reflect.TypeOf(indexKey).Kind() <= reflect.Float64
```
gitlawr, line 546: What's the difference comparing to previous implementation? What cases we were missing?

pkg/terraform/parser/parser.go
```diff
@@ -538,3 +538,14 @@ func ParseIndexKey(rs resourceState, is instanceObjectState) (string, error) {
 
 	return strs.Join(".", rs.Module, rs.Type, rs.Name), nil
 }
+
+// IsNumber returns true if the index key is a number.
+// https://github.com/hashicorp/terraform/blob/main/internal/states/statefile/version4.go#L114C1-L122C32.
```
gitlawr, line 543: Use a permanent link in case the main branch changes.

alexcodelf:
> What's the consequence when index parsing fails?

If fails, the component will not be saved and shows in components list.

gitlawr:
Is it the cause of https://github.com/seal-io/walrus/issues/2214?

alexcodelf:
Not sure if [https://github.com/seal-io/walrus/issues/2214](https://github.com/seal-io/walrus/issues/2214) is caused by this issue as the component creation may be affect by this issue

pkg/terraform/parser/parser.go
alexcodelf: It will treat all number type as the case. 
```
package main

import (
	"fmt"
)

func main() {
	var a int = 1
	var b int8 = 2
	var c int16 = 3
	var d int32 = 4
	var e int64 = 5
	var f uint = 6
	var g uint8 = 7
	var h uint16 = 8
	var i uint32 = 9
	var j uint64 = 10
	var k float32 = 11
	var l float64 = 12

	fmt.Printf("a:%v, b:%v, c:%v, d:%v, e:%v, f:%v, g:%v, h:%v, i:%v, j:%v, k:%v, l:%v\n", a, b, c, d, e, f, g, h, i, j, k, l)
}
```

The result will be `a:1, b:2, c:3, d:4, e:5, f:6, g:7, h:8, i:9, j:10, k:11, l:12`, just in case there are other number type. Event though the `encoding/json` will only parse to `float64`

Issue #2249, opened 2024-03-14
Deployment logs are not displayed on the workflow step


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. New a workflow and add a 'Deploy Resource' step.
2. Execute the workflow and check the deployment logs.

**Result**

1. It did not displays the deployment logs.
<img width="1133" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/6b848bc7-691d-412d-bb4b-6d680d42060a">

**Expected behavior**

**Environment**
- Walrus version: main c6bba6b527201671bcf74167e17f4fdd4fcfa40d
- UI version: dev-8e2e304

**Additional context**

Finenyaco:
And  deploying an aws mysql resource via workflow, the step shows failure soon, but the resource is actually still running.

<img width="949" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/0e900edf-798e-4c98-a524-0b4716e4f0f9">

<img width="1718" alt="image" src="https://github.com/seal-io/walrus/assets/26084756/abb4d947-3f76-4e7d-8163-78a93204596e">

Issue #2192, opened 2024-03-06
The Running logs did not display during the deployment progress when deploy directly


Finenyaco:
**Describe the bug**

**Steps to reproduce**
1. New a resource.
2. Save and deploy.
3. Watch the deployment logs.

https://github.com/seal-io/walrus/assets/26084756/fed127e2-b037-49e0-a425-6082dcaa958b

**Result**

It didn't display the correct running logs during the deployment progress.   The running logs only display correctly after closing and reopening the window.

**Expected behavior**

**Environment**
- Walrus version: main 96dd76526ad969a2776aecc30fbe79c812b81c2f
- UI version: dev-874ecc7

**Additional context**

hibig:
When the status transitions from running to succeeded, the ```record```  field has no value.

<img width="974" alt="Snipaste_2024-03-14_16-44-01" src="https://github.com/seal-io/walrus/assets/24664190/aa3be4da-ec72-4f78-9ac7-76501b4d26db">

Finenyaco:
**Environment**
- Walrus version: main d6f2761cdd38ed8bf6027cf7eeaec7165d7e61bc
- UI version: dev-073f7c5

**Test Result:** pass

PR #2252, opened 2024-03-14
fix: workflow shows no logs

alexcodelf:
**Problem:**

- Workflow step shows no logs and return error.
- Resource run list stream doest not return record after succeeded. https://github.com/seal-io/walrus/issues/2192#issuecomment-1996902795

**Solution:**

- As the resource run has two stage, the workflow step needs update to fetch two step logs.
- Return plan and apply records.

**Related Issue:**
#2249

### DIFF
--- a/pkg/workflow/step/service_step.go
+++ b/pkg/workflow/step/service_step.go
@@ -34,66 +34,92 @@ if [ "{{workflow.parameters.tlsVerify}}" == "true" ]; then
 	tlsVerify=""
 fi
 
+common_get() {
+    curl -sS --fail-with-body "$1" -X GET -H "Authorization: Bearer $token" $tlsVerify --compressed
+}
+
+wait_run_status() {
+    local resourceName="$1"
+    local runID="$2"
+    local breakSummaryStatus="$3"
+    local timeout=30
+    local factor=1
+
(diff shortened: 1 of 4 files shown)


pkg/workflow/step/service_step.go
```diff
@@ -31,69 +72,53 @@ commonPath="$serverURL/v1/projects/$projectID/environments/$environmentID"
...
-	timeout=$((timeout - factor))
-done
+# Watch run plan logs until the plan finished.
+common_get "$commonPath/resources/$resourceName/runs/$runID/log?jobType=plan&watch=true" --compressed
+
+# Check run status, wait until run status is running.
+wait_run_status "$resourceName" "$runID" "Running"
+
+# Watch run logs until the run finished.
+common_get "$commonPath/resources/$resourceName/runs/$runID/log?jobType=apply&watch=true" --compressed
```
gitlawr, line 118: Is the compressed arg valid? Looks like common_get is not handling it.

alexcodelf: updated.

gitlawr (approved):
LGTM

aiwantaozi (approved):
LGTM

Issue #2254, opened 2024-03-15
Update Alibaba cloud provider name


gitlawr:
**Describe the bug**
Change `Alibaba` to `Alibaba Cloud`

**Steps to reproduce**
1. Connector type:
<img width="659" alt="image" src="https://github.com/seal-io/walrus/assets/5697937/e4f7d955-95c1-4b9b-9e65-4145ac605d4c">
2. Built-in definition rule: 
<img width="384" alt="image" src="https://github.com/seal-io/walrus/assets/5697937/a930e462-f07b-4425-861b-4179570f33a1">

**Environment**
- Walrus version: v0.6.0-rc2

PR #2269, opened 2024-04-11
Connector CRD and API

shanewxy:
**Problem:**

**Solution:**

**Related Issue:**


pkg/apis/walruscore/v1/connector.go
```diff
@@ -20,14 +20,69 @@ type Connector struct {
 
 var _ runtime.Object = (*Connector)(nil)
 
+const (
+	ConnectorTypeDocker     string = "Docker"
+	ConnectorTypeKubernetes string = "Kubernetes"
+	ConnectorTypeAlibaba    string = "Alibaba"
+	ConnectorTypeAWS        string = "AWS"
+	ConnectorTypeAzure      string = "Azure"
+	ConnectorTypeGoogle     string = "Google"
```
gitlawr: Should we move these out of the coreapis package?

Also update Alibaba to address https://github.com/seal-io/walrus/issues/2254

pkg/clients/applyconfiguration/walrus/v1/connectorbinding.go
```diff
@@ -0,0 +1,208 @@
+// SPDX-FileCopyrightText: 2024 Seal, Inc
+// SPDX-License-Identifier: Apache-2.0
+
+// Code generated by "walrus", DO NOT EDIT.
```
gitlawr, line 4: move these changes to the generate commit

pkg/operator/types/types.go
```diff
@@ -0,0 +1,138 @@
...
+	walruscore "github.com/seal-io/walrus/pkg/apis/walruscore/v1"
+	"github.com/seal-io/walrus/pkg/clients/clientset"
+)
+
+// Type indicates the type of Operator,
+// e.g. Kubernetes, AWS, etc.
+type Type = string
+
+// Operator holds the actions that an operator must satisfy.
+type Operator interface {
```
gitlawr: The name can cause confusion in the context of Kubernetes. Let's change it to ResourceHandler.

And adapt with existing packaging pattern. Move the interface to pkg resourcehandler and implementations to resourcehandlers.

pkg/extensionapis/walrus/connector.go
```diff
@@ -19,22 +20,76 @@ import (
...
+			extensionapi.JSONPathTableColumnDefinition{
+				TableColumnDefinition: meta.TableColumnDefinition{
+					Name: "Type",
+					Type: "string",
+				},
+				JSONPath: ".spec.type",
+			},
+			extensionapi.JSONPathTableColumnDefinition{
+				TableColumnDefinition: meta.TableColumnDefinition{
+					Name: "Applicable-Environment-Type",
```
gitlawr: Let's use the `Applicable Environment Type` format as a convention.

pkg/apis/walruscore/v1/connector_binding.go
```diff
@@ -0,0 +1,55 @@
...
+
+var _ runtime.Object = (*ConnectorBinding)(nil)
+
+// ConnectorBindingSpec defines the desired state of ConnectorBinding.
+type ConnectorBindingSpec struct {
+	// ConnectorRef is the reference to the connector.
+	ConnectorRef ConnectorReference `json:"connector_ref"`
+
+	// ConnectorType is the type of the connector.
+	ConnectorType string `json:"connector_type"`
```
gitlawr: Why do we need the type

pkg/controllers/walruscore/connector.go
```diff
@@ -3,29 +3,57 @@ package walruscore
...
 
 	walruscore "github.com/seal-io/walrus/pkg/apis/walruscore/v1"
+	"github.com/seal-io/walrus/pkg/clients/clientset"
 	"github.com/seal-io/walrus/pkg/controller"
 )
 
 // ConnectorReconciler reconciles a v1.Connector object.
-type ConnectorReconciler struct{}
+type ConnectorReconciler struct {
+	client clientset.Clientset
```
gitlawr: Let's use `ctrlcli.Client` in controller which implements caching

pkg/resourcehandler/consts.go
```diff
@@ -0,0 +1,17 @@
+package resourcehandler
+
+const (
+	ConnectorTypeDocker     string = "Docker"
+	ConnectorTypeKubernetes string = "Kubernetes"
+	ConnectorTypeAlibaba    string = "Alibaba Cloud"
```
gitlawr: rename the constant name and remove the space

pkg/apis/walruscore/v1/connector_binding.go
shanewxy: Creating connector_binding needs validation for type uniqueness under the same environment. Without the type will need extra queries for each connector.

gitlawr: It's determined by the ref, should we put it in status

pkg/apis/walruscore/v1/connector_binding.go
```diff
@@ -0,0 +1,55 @@
...
+	Spec   ConnectorBindingSpec   `json:"spec,omitempty"`
+	Status ConnectorBindingStatus `json:"status,omitempty"`
+}
+
+var _ runtime.Object = (*ConnectorBinding)(nil)
+
+// ConnectorBindingSpec defines the desired state of ConnectorBinding.
+type ConnectorBindingSpec struct {
+	// ConnectorRef is the reference to the connector.
+	ConnectorRef ConnectorReference `json:"connector_ref"`
```
aiwantaozi: connector_ref -> connectorRef

pkg/apis/walruscore/v1/connector.go
```diff
@@ -20,14 +20,51 @@ type Connector struct {
...
-	// TODO: your spec here
+	// ApplicableEnvironmentType is the environment type that the connector is applicable to.
+	//
+	// +k8s:validation:enum=["Development","Staging","Production"]
+	ApplicableEnvironmentType EnvironmentType `json:"applicable_environment_type,omitempty"`
+
+	// Category is the category of the connector.
+	//
+	// +k8s:validation:enum=["Docker","Kubernetes","Custom","VersionControl","CloudProvider"]
+	Category string `json:"category"`
```
aiwantaozi: When the category is type string, it won't generate openapi enum validation, need to create another type like EnvironmentType

pkg/apis/walruscore/v1/connector.go
```diff
@@ -20,14 +20,51 @@ type Connector struct {
...
+
+	// Type is the type of the connector.
+	Type string `json:"type"`
+
+	// ConfigData is the configuration data of the connector.
+	ConfigData map[string]Property `json:"config_data"`
```
aiwantaozi: config_data -> configData

pkg/apis/walruscore/v1/connector.go
```diff
@@ -20,14 +20,51 @@ type Connector struct {
...
+
+	// Connector config version.
+	// +optional
+	ConfigVersion string `json:"config_version,omitempty"`
```
aiwantaozi: config_version - > configVersion

pkg/apis/walruscore/v1/connector.go
```diff
@@ -20,14 +20,51 @@ type Connector struct {
...
+
+	// Description is the description of the connector.
+	// +optional
+	Description string `json:"description,omitempty"`
+
+	// SecretName is the name of the secret that contains the credentials for the connector.
+	// +optional
+	SecretName string `json:"secret_name,omitempty"`
```
aiwantaozi: secret_name -> secretName

pkg/apis/walruscore/v1/connector.go
```diff
@@ -20,14 +20,51 @@ type Connector struct {
...
+	Visible bool `json:"visible"`
+}
+
 // ConnectorSpec defines the desired state of Connector.
 type ConnectorSpec struct {
-	// TODO: your spec here
+	// ApplicableEnvironmentType is the environment type that the connector is applicable to.
+	//
+	// +k8s:validation:enum=["Development","Staging","Production"]
+	ApplicableEnvironmentType EnvironmentType `json:"applicable_environment_type,omitempty"`
```
aiwantaozi: applicable_environment_type -> applicableEnvironmentType

pkg/apis/walruscore/v1/connector.go
```diff
@@ -20,14 +20,51 @@ type Connector struct {
...
+
+	// ConfigData is the configuration data of the connector.
+	ConfigData map[string]Property `json:"config_data"`
+
+	// Connector config version.
+	// +optional
+	ConfigVersion string `json:"config_version,omitempty"`
+
+	// Description is the description of the connector.
+	// +optional
```
aiwantaozi: optional could be removed, field that contains `omitempty` is optional

pkg/apis/walrus/v1/environment.go
```diff
@@ -25,38 +26,12 @@ type Environment struct {
...
-	}
-}
-
 // EnvironmentSpec defines the desired state of Environment.
 type EnvironmentSpec struct {
 	// Type is the type of the environment.
 	//
 	// +k8s:validation:enum=["development","staging","production"]
-	Type EnvironmentType `json:"type"`
+	Type walruscore.EnvironmentType `json:"type"`
```
aiwantaozi, line 34: `+k8s:validation:enum=["development","staging","production"]`, should use the same value as walruscore.EnvironmentType

pkg/webhooks/walruscore/connector.go
```diff
@@ -2,42 +2,124 @@ package walruscore
...
+	return r.validateConnector(ctx, obj)
+}
+
+func (r *ConnectorWebhook) validateConnector(ctx context.Context, obj runtime.Object) (ctrladmission.Warnings, error) {
+	conn := obj.(*walruscore.Connector)
+	if stringx.StringWidth(conn.Name) > 30 {
+		return nil, field.TooLongMaxLength(field.NewPath("name"), stringx.StringWidth(conn.Name), 30)
+	}
+
+	err := resourcehandlers.IsConnected(ctx, conn, r.client)
```
aiwantaozi, line 58: Is this reachability check also necessary when updating?

pkg/webhooks/walruscore/connector.go
```diff
@@ -2,42 +2,124 @@ package walruscore
...
+	"github.com/seal-io/walrus/pkg/kubemeta"
+	"github.com/seal-io/walrus/pkg/resourcehandlers"
 	"github.com/seal-io/walrus/pkg/webhook"
 )
 
 // ConnectorWebhook hooks a v1.Connector object.
 //
 // nolint: lll
+// +k8s:webhook-gen:mutating:group="walruscore.seal.io",version="v1",resource="connectors",scope="Namespaced"
+// +k8s:webhook-gen:mutating:operations=["CREATE","UPDATE"],failurePolicy="Fail",sideEffects="None",matchPolicy="Equivalent",timeoutSeconds=10
```
aiwantaozi: sideEffects should be NoneOnDryRun since it creates a secret here.

pkg/webhooks/walruscore/connector_binding.go
```diff
@@ -0,0 +1,147 @@
...
+	}
+
+	return nil, nil
+}
+
+func (r *ConnectorBindingWebhook) ValidateUpdate(
+	ctx context.Context,
+	oldObj, newObj runtime.Object,
+) (ctrladmission.Warnings, error) {
+	return nil, nil
```
aiwantaozi, line 110: Is validation needed when updating?

pkg/extensionapis/walrus/connector.go
```diff
@@ -19,22 +20,76 @@ import (
...
+					Type: "string",
+				},
+				JSONPath: ".spec.type",
+			},
+			extensionapi.JSONPathTableColumnDefinition{
+				TableColumnDefinition: meta.TableColumnDefinition{
+					Name: "Applicable Environment Type",
+					Type: "string",
+				},
+				JSONPath: ".spec.applicable_environment_type",
```
aiwantaozi: Should update when applicable_environment_type update to applicableEnvironmentType

pkg/extensionapis/walrus/connector.go
```diff
@@ -19,22 +20,76 @@ import (
...
+			},
+			extensionapi.JSONPathTableColumnDefinition{
+				TableColumnDefinition: meta.TableColumnDefinition{
+					Name: "Project",
+					Type: "string",
+				},
+				JSONPath: ".spec.project",
```
aiwantaozi: .status.project

pkg/extensionapis/walrus/connector.go
```diff
@@ -19,22 +20,76 @@ import (
 type ConnectorHandler struct {
 	extensionapi.ObjectInfo
 	extensionapi.CurdOperations
+
+	Client ctrlcli.Client
```
aiwantaozi: Where is this Client used?

pkg/extensionapis/walrus/connector_binding.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+// ConnectorBindingHandler handles v1.ConnectorBinding objects.
+//
+// ConnectorBindingHandler proxies the v1.ConnectorBinding objects to the walrus core.
+type ConnectorBindingHandler struct {
+	extensionapi.ObjectInfo
+	extensionapi.CurdOperations
+
+	Client ctrlcli.Client
```
aiwantaozi: Where is this used?

pkg/extensionapis/walrus/connector_binding.go
```diff
@@ -0,0 +1,119 @@
...
+	// Create table convertor to pretty the kubectl's output.
+	var tc rest.TableConvertor
+	{
+		tc, err = extensionapi.NewJSONPathTableConvertor(
+			extensionapi.JSONPathTableColumnDefinition{
+				TableColumnDefinition: meta.TableColumnDefinition{
+					Name: "Connector",
+					Type: "string",
+				},
+				JSONPath: ".spec.connector_ref.name",
```
aiwantaozi: Maybe need extra info like namespace? Since we have the global connector.

pkg/resourcehandler/types.go
```diff
@@ -0,0 +1,138 @@
...
+type ResourceComponentOperationKeys struct {
+	// Labels stores label of layer,
+	// its length means each key contains levels with the same value as level.
+	Labels []string `json:"labels,omitempty"`
+	// Keys stores key in tree.
+	Keys []ResourceComponentOperationKey `json:"keys,omitempty"`
+}
+
+// ResourceComponentOperationKey holds hierarchy query keys.
+type ResourceComponentOperationKey struct {
```
aiwantaozi: The definition here needs to consider how the resource component is used.

aiwantaozi (changes requested):

pkg/apis/walruscore/v1/connector_binding.go
```diff
@@ -0,0 +1,55 @@
...
+// ConnectorBindingSpec defines the desired state of ConnectorBinding.
+type ConnectorBindingSpec struct {
+	// ConnectorRef is the reference to the connector.
+	ConnectorRef ConnectorReference `json:"connectorRef"`
+}
+
+// ConnectorBindingStatus defines the observed state of ConnectorBinding.
+type ConnectorBindingStatus struct {
+	// ConnectorType is the type of the connector.
+	ConnectorType string `json:"connectorType"`
```
thxCode: s/ConnectorType/Type/

it's better to record the `Catalog/(represented) SecretName` here too.

pkg/apis/walruscore/v1/connector.go
```diff
@@ -20,14 +20,50 @@ type Connector struct {
...
+
+	// Type is the type of the connector.
+	Type string `json:"type"`
+
+	// ConfigData is the configuration data of the connector.
+	ConfigData map[string]Property `json:"configData"`
+
+	// Connector config version.
+	// +optional
+	ConfigVersion string `json:"configVersion,omitempty"`
```
thxCode: Can we merge these fields?

```
type ConnectorConfigEntry struct {
  Value string `json:"value"`
  Visible bool `json:"visible"`
}

type ConnectorConfig struct {
  Version string `json:"version"`
  Data map[string]ConnectorConfigEntry `json:"data"`
}
```

pkg/apis/walruscore/v1/connector.go
```diff
@@ -20,14 +20,50 @@ type Connector struct {
...
+
+	// Description is the description of the connector.
+	Description string `json:"description,omitempty"`
+
+	// SecretName is the name of the secret that contains the credentials for the connector.
+	// +optional
+	SecretName string `json:"secretName,omitempty"`
```
thxCode: does it represent where we store the `ConnectorConfig` (I pointed out above)?

if so, I think this should be a part of `ConnectorStatus`.

pkg/apis/walruscore/v1/connector_binding.go
```diff
@@ -0,0 +1,55 @@
...
+	Spec   ConnectorBindingSpec   `json:"spec,omitempty"`
+	Status ConnectorBindingStatus `json:"status,omitempty"`
+}
+
+var _ runtime.Object = (*ConnectorBinding)(nil)
+
+// ConnectorBindingSpec defines the desired state of ConnectorBinding.
+type ConnectorBindingSpec struct {
+	// ConnectorRef is the reference to the connector.
+	ConnectorRef ConnectorReference `json:"connectorRef"`
```
thxCode: s/ConnectorRef/Connector/

pkg/controllers/walruscore/connector.go
```diff
@@ -3,29 +3,142 @@ package walruscore
...
+			return ctrl.Result{}, err
+		}
+	}
+
+	defer func() {
+		if err != nil {
+			apistatus.ConnectorConditionConnected.False(obj, apistatus.ConnectorConditionReasonDisconnected, err.Error())
+			err = r.updateStatus(ctx, obj)
+			if err != nil {
+				ctrllog.FromContext(ctx).Error(err, "failed to update connector status", "connector", obj)
```
thxCode: make `logger := ctrllog.FromContext(ctx)` at top, then use logger here.

pkg/controllers/walruscore/connector.go
```diff
@@ -3,29 +3,142 @@ package walruscore
...
+		}
+	}()
+
+	rh, err := resourcehandlers.Get(ctx, resourcehandler.CreateOptions{
+		Connector: *obj,
+		Client:    r.client,
+	})
+	if err != nil {
+		return ctrl.Result{}, err
+	}
```
thxCode: logger.Error(err, "fetch resource handler")
return ctrl.Result{}, ctrlcli.IgnoreNotFound(err)

pkg/apis/walruscore/v1/connector.go
```diff
@@ -20,14 +20,50 @@ type Connector struct {
...
 }
 
 // ConnectorStatus defines the observed state of Connector.
 type ConnectorStatus struct {
-	// TODO: your status here
+	// StatusDescriptor defines the status of the Connector.
+	StatusDescriptor `json:",inline"`
+
+	// Project is the project that the connector belongs to.
+	Project string `json:"project"`
```
thxCode: json:"project,omitempty" for global connector

pkg/resourcehandlers/alibaba/resourceexec/ecs_instance.go
```diff
@@ -0,0 +1,280 @@
...
+type ecsInstance struct {
+	cred          *optypes.Credential
+	ecsCli        *ecs.Client
+	assistCli     *assistclient.Client
+	realConnected bool
+}
+
+func getEcsInstance(ctx context.Context) (optypes.ExecutableResource, error) {
+	var (
+		res = &ecsInstance{}
```
thxCode: res ecsInstance

pkg/resourcehandlers/alibaba/resourceexec/ecs_instance.go
```diff
@@ -0,0 +1,280 @@
...
+		if !r.realConnected {
+			return
+		}
+
+		if err := r.assistCli.SendCloseMessage(); err != nil {
+			klog.Warningf("error send close message: %v", err)
+		}
+	}()
+
+	ctxWithCancel, cancel := context.WithCancel(ctx)
```
thxCode: s/ctxWithCancel/ctx

pkg/resourcehandlers/aws/resourceexec/ec2_instance.go
```diff
@@ -0,0 +1,143 @@
...
+		default:
+			_, err := io.Copy(c, opts.In)
+			if err != nil {
+				return fmt.Errorf("error write to remote: %w", err)
+			}
+
+			return nil
+		}
+	})
+
```
thxCode: remove this line

pkg/resourcehandlers/k8s/resourcehandler.go
```diff
@@ -0,0 +1,151 @@
...
+			// NB(shanewxy): Keep the real error from request.
+			lastErr = IsConnected(context.TODO(), op.CoreCli.RESTClient())
+
+			return lastErr == nil, ctx.Err()
+		},
+	)
+
+	if lastErr != nil {
+		err = lastErr // Use last error to overwrite context error while existed.
+	}
```
thxCode: try waitx.PollUntilContextCancel

pkg/systemkuberes/connector.go
```diff
@@ -0,0 +1,161 @@
...
+
+func installDefaultKubernetesConnector(
+	ctx context.Context,
+	cli clientset.Interface,
+	project string,
+	envType walruscore.EnvironmentType,
+) (*walrus.Connector, error) {
+	connCli := cli.WalrusV1().Connectors(project)
+
+	config, err := readKubeConfig()
```
thxCode, line 29: try kubeconfig.ConvertRestConfigToApiConfig

pkg/systemkuberes/connector.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	conn, err := kubeclientset.Create(
+		ctx,
+		connCli,
+		c,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("install default kubernetes connector: %w", err)
+	}
```
thxCode: under docker run mode, the embedded k3s will change its generated config,
we need to update the local connector config at bootstrap, please use `kubeclientset.Update(,,, kubeclientset.WithCreateIfNotExisted[*walrus.Connector]())`.

staging/github.com/seal-io/utils/hash/sum.go
```diff
@@ -0,0 +1,33 @@
...
+
+func SumSHA256(bs []byte) string {
+	sum := sha256.Sum256(bs)
+	return hex.EncodeToString(sum[:])
+}
+
+func SumSHA224(bs []byte) string {
+	sum := sha256.Sum224(bs)
+	return hex.EncodeToString(sum[:])
+}
```
thxCode: try stringx package

staging/github.com/seal-io/utils/stringx/string.go
```diff
@@ -89,3 +89,12 @@ func ReplaceFunc(s string, rp func(rune) rune) string {
...
 }
+
+// LastContent retrieves the last characters of a string.
+func LastContent(content string, length int) string {
+	if len(content) < length {
+		return content
+	}
+
+	return content[len(content)-length:]
+}
```
thxCode: LastContent causes one memory assigned.

for Golang, getting the last content of a string is very easy to achieve.

let's remove this function, and use string[last_content_start:].

pkg/controllers/walruscore/connector.go
```diff
@@ -3,29 +3,145 @@ package walruscore
...
+		}
+	}
+
+	defer func() {
+		if err != nil {
+			apistatus.ConnectorConditionConnected.False(obj, apistatus.ConnectorConditionReasonDisconnected, err.Error())
+			err = r.updateStatus(ctx, obj)
+			if err != nil {
+				logger.Error(err, "failed to update connector status", "connector", obj)
+			}
```
thxCode: let's keep the original error here.

pkg/controllers/walruscore/connector_binding.go
```diff
@@ -0,0 +1,90 @@
...
+
+				labels := ns.GetLabels()
+				delete(labels, walruscore.ProviderLabelPrefix+strings.ToLower(cb.Status.Type))
+				ns.SetLabels(labels)
+
+				err = r.client.Update(ctx, ns)
+				if err != nil {
+					ctrllog.FromContext(ctx).Error(err, "failed to update namespace", "namespace", cb.Namespace)
+				}
+			},
```
thxCode: Avoid doing heavy processing during an event transiting.

Use finalizer to remove/update the related resources.

thxCode: see walrus/subject.authz or project_subject.authz

pkg/apis/walruscore/v1/connector_binding.go
```diff
@@ -0,0 +1,58 @@
...
+// ConnectorBinding is the schema for the connectorbindings API.
+//
+// +genclient
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+// +k8s:crd-gen:resource:scope="Namespaced",subResources=["status"]
+type ConnectorBinding struct {
+	meta.TypeMeta   `json:",inline"`
+	meta.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   ConnectorBindingSpec   `json:"spec,omitempty"`
```
aiwantaozi: Spec should be required for ConnectorBinding and Connector?

pkg/webhooks/walruscore/connector.go
aiwantaozi: NoneOnDryRun is for mutating webhook.

pkg/webhooks/walruscore/connector_binding.go
```diff
@@ -0,0 +1,151 @@
...
+	walruscore "github.com/seal-io/walrus/pkg/apis/walruscore/v1"
+	"github.com/seal-io/walrus/pkg/systemmeta"
+	"github.com/seal-io/walrus/pkg/webhook"
+)
+
+// ConnectorBindingWebhook hooks a v1.ConnectorBinding object.
+//
+// nolint: lll
+// +k8s:webhook-gen:mutating:group="walruscore.seal.io",version="v1",resource="connectorbindings",scope="Namespaced"
+// +k8s:webhook-gen:mutating:operations=["CREATE","UPDATE"],failurePolicy="Fail",sideEffects="None",matchPolicy="Equivalent",timeoutSeconds=10
```
aiwantaozi: sideEffects is NoneOnDryRun since it will update namespace.

pkg/webhooks/walruscore/connector_binding.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	err = r.client.Get(ctx, ctrlcli.ObjectKeyFromObject(ns), ns)
+	if err != nil {
+		return nil, err
+	}
+
+	t := systemmeta.DescribeResourceNote(ns, "type")
+	if walruscore.EnvironmentType(t) != connector.Spec.ApplicableEnvironmentType {
+		return nil, field.Invalid(
+			field.NewPath("spec", "connectorRef", "name"),
```
aiwantaozi: wrong field name connectorRef

pkg/webhooks/walruscore/connector_binding.go
```diff
@@ -0,0 +1,151 @@
...
+
+	err = r.client.List(ctx, connList, ctrlcli.InNamespace(cb.Namespace))
+	if err != nil {
+		return nil, err
+	}
+
+	for _, conn := range connList.Items {
+		if conn.Status.Type == connector.Spec.Type {
+			return nil, field.Invalid(
+				field.NewPath("status", "connectorType"),
```
aiwantaozi: wrong field name: connectorType

pkg/controllers/walruscore/connector.go
```diff
@@ -3,29 +3,145 @@ package walruscore
...
 var _ ctrlreconcile.Reconciler = (*ConnectorReconciler)(nil)
 
 func (r *ConnectorReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
-	_ = ctrllog.FromContext(ctx)
+	logger := ctrllog.FromContext(ctx)
+
+	obj := new(walruscore.Connector)
+	err := r.client.Get(ctx, req.NamespacedName, obj)
+	if err != nil {
+		return ctrl.Result{}, err
```
aiwantaozi: IgnoreNotFound

pkg/controllers/walruscore/connector_binding.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+var _ ctrlreconcile.Reconciler = (*ConnectorBindingReconciler)(nil)
+
+func (r *ConnectorBindingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	obj := new(walruscore.ConnectorBinding)
+
+	err := r.client.Get(ctx, req.NamespacedName, obj)
+	if err != nil {
+		return ctrl.Result{}, err
```
aiwantaozi: IgnoreNotFound

pkg/controllers/walruscore/connector_binding.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
```
aiwantaozi, line 40: Skip deletion

pkg/resourcehandler/types.go
```diff
@@ -0,0 +1,118 @@
...
+	Log(context.Context, string, LogOptions) error
+
+	// Exec executes commands to the given key.
+	Exec(context.Context, string, ExecOptions) error
+
+	// Label apply labels to the resource.
+	//
+	// The given model.ResourceComponent item must specify the following fields:
+	// ID, DeployerType, Type and Name.
+	Label(context.Context, *walrus.ResourceComponent, map[string]string) error
```
aiwantaozi: Label function is for cost, we no need to migrate the label function to 0.7

pkg/resourcehandlers/alibaba/resourcehandler.go
```diff
@@ -0,0 +1,115 @@
...
+}
+
+// Exec implements resourcehandlers.ResourceHandler.
+func (op Operator) Exec(ctx context.Context, key string, opts resourcehandler.ExecOptions) error {
+	newCtx := context.WithValue(ctx, types.CredentialKey, op.cred)
+	return resourceexec.Exec(newCtx, key, opts)
+}
+
+// Label implements resourcehandlers.ResourceHandler.
+func (op Operator) Label(ctx context.Context, resource *walrus.ResourceComponent, labels map[string]string) error {
```
aiwantaozi: Remove.

pkg/resourcehandlers/aws/resourcehandler.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+// Exec implements resourcehandlers.ResourceHandler.
+func (op Operator) Exec(ctx context.Context, key string, opts resourcehandler.ExecOptions) error {
+	newCtx := context.WithValue(ctx, types.CredentialKey, op.cred)
+	return resourceexec.Exec(newCtx, key, opts)
+}
+
+// Label implements resourcehandlers.ResourceHandler.
+func (op Operator) Label(ctx context.Context, resource *walrus.ResourceComponent, labels map[string]string) error {
```
aiwantaozi: Ditto.

pkg/resourcehandlers/docker/resourcehandler.go
```diff
@@ -0,0 +1,99 @@
...
+	return nil
+}
+
+// Exec implements resourcehandlers.ResourceHandler.
+func (op Operator) Exec(ctx context.Context, key string, opts resourcehandler.ExecOptions) error {
+	return nil
+}
+
+// Label implements resourcehandlers.ResourceHandler.
+func (op Operator) Label(ctx context.Context, resource *walrus.ResourceComponent, labels map[string]string) error {
```
aiwantaozi: Ditto.

pkg/resourcehandlers/google/resourcehandler.go
```diff
@@ -0,0 +1,101 @@
...
+	return nil
+}
+
+// Exec implements resourcehandlers.ResourceHandler.
+func (op Operator) Exec(ctx context.Context, key string, opts resourcehandler.ExecOptions) error {
+	return nil
+}
+
+// Label implements resourcehandlers.ResourceHandler.
+func (op Operator) Label(ctx context.Context, resource *walrus.ResourceComponent, labels map[string]string) error {
```
aiwantaozi: Ditto

pkg/resourcehandlers/k8s/resourcehandler.go
```diff
@@ -0,0 +1,142 @@
...
+	return nil
+}
+
+// Exec implements resourcehandlers.ResourceHandler.
+func (op Operator) Exec(ctx context.Context, key string, opts resourcehandler.ExecOptions) error {
+	return nil
+}
+
+// Label implements resourcehandlers.ResourceHandler.
+func (op Operator) Label(ctx context.Context, resource *walrus.ResourceComponent, labels map[string]string) error {
```
aiwantaozi: Ditto

pkg/resourcehandler/types.go
```diff
@@ -0,0 +1,118 @@
...
+	// The given walrus.ResourceComponent item must specify the following fields:
+	// ID, DeployerType, Type and Name.
+	GetStatus(context.Context, *walrus.ResourceComponent) ([]meta.Condition, error)
+
+	// GetComponents gets components of the given resource,
+	// returns list must not be `nil` unless unexpected input or raising error,
+	// it can be used to clean stale items safety if got an empty list.
+	//
+	// The given walrus.ResourceComponent item must specify the following fields:
+	// ID, DeployerType, Type, Name, ProjectID, EnvironmentID, ResourceID and ConnectorID.
```
aiwantaozi: Update the comment.

pkg/resourcehandlers/alibaba/resourcehandler.go
```diff
@@ -0,0 +1,115 @@
...
+	return nil
+}
+
+func (op Operator) GetKeys(ctx context.Context, component *walrus.ResourceComponent) (*resourcehandler.ResourceComponentOperationKeys, error) {
+	return nil, nil
+}
+
+func (op Operator) GetStatus(ctx context.Context, component *walrus.ResourceComponent) ([]meta.Condition, error) {
+	return nil, nil
```
aiwantaozi, line 111: Add TODO comment for all GetStatus functions, since they need to be migrated later.

pkg/controllers/walruscore/connector_binding.go
```diff
@@ -0,0 +1,110 @@
...
+		ns := &core.Namespace{
+			ObjectMeta: meta.ObjectMeta{
+				Name: obj.Namespace,
+			},
+		}
+		err = r.client.Get(ctx, ctrlcli.ObjectKeyFromObject(ns), ns)
+		if err != nil {
+			logger.Error(err, "failed to get namespace", "namespace", obj.Namespace)
+		}
```
thxCode: return ctrl.Result{}, ctrlcli.IgnoreNotFound(err)

pkg/controllers/walruscore/connector_binding.go
```diff
@@ -0,0 +1,110 @@
...
+
+		labels := ns.GetLabels()
+		delete(labels, walruscore.ProviderLabelPrefix+strings.ToLower(obj.Status.Type))
+		ns.SetLabels(labels)
+
+		err = r.client.Update(ctx, ns)
+		if err != nil {
+			logger.Error(err, "failed to update namespace", "namespace", obj.Namespace)
+		}
```
thxCode: return ctrl.Result{}, ctrlcli.IgnoreNotFound(err)

pkg/webhooks/walruscore/connector_binding.go
```diff
@@ -0,0 +1,154 @@
...
+
+	return nil, nil
+}
+
+func (r *ConnectorBindingWebhook) ValidateUpdate(
+	ctx context.Context,
+	oldObj, newObj runtime.Object,
+) (ctrladmission.Warnings, error) {
+	return r.validate(ctx, newObj)
+}
```
thxCode: ConnectorBindingSpec should be immutable, use reflect.DeepEqual to compare new/old specs.

```
func (r *ConnectorBindingWebhook) ValidateUpdate(
  ctx context.Context,
  oldObj, newObj runtime.Object,
) (ctrladmission.Warnings, error) {
  var oldCb, newCb = oldObj.(*walruscore.ConnectorBinding), newObj.(*walruscore.ConnectorBinding)
  if !reflect.DeepEqual(oldCb.Spec, newCb.Spec) {
    // return error 
  }

  return nil
}
```

pkg/webhooks/walruscore/connector_binding.go
```diff
@@ -0,0 +1,154 @@
...
+
+	labels[walruscore.ProviderLabelPrefix+strings.ToLower(conn.Spec.Type)] = "true"
+	ns.SetLabels(labels)
+
+	err = r.client.Update(ctx, ns)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
thxCode: we should avoid modifying other resources during admission if we can.

the mutating webhook can be used to update the ConnectorBinding status.

```
func (r *ConnectorBindingWebhook) Default(ctx context.Context, obj runtime.Object) error {
  cb := obj.(*walruscore.ConnectorBinding)
  if cb.DeletionTimestamp != nil {
    return nil
  }

  // Get Connector via spec.

  // Update connector binding's status here.
  cb.Status.Type = ...
  cb.Status.Category = ...

  return nil
}
```

pkg/webhooks/walruscore/connector_binding.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+var _ ctrlwebhook.CustomValidator = (*ConnectorBindingWebhook)(nil)
+
+func (r *ConnectorBindingWebhook) ValidateCreate(
+	ctx context.Context,
+	obj runtime.Object,
+) (ctrladmission.Warnings, error) {
+	return r.validate(ctx, obj)
+}
```
thxCode: Operate Environment rather than Namespace.

```
func (r *ConnectorBindingWebhook) ValidateCreate(
  ctx context.Context,
  obj runtime.Object,
) (ctrladmission.Warnings, error) {
  cb := obj.(*walruscore.ConnectorBinding)

  // Validate connector applicable environment type.
  {
    conn := &walruscore.Connector{
      ObjectMeta: meta.ObjectMeta{
        Name:      cb.Spec.Connector.Name,
        Namespace: cb.Spec.Connector.Namespace,
      },
    }
    err := r.client.Get(ctx, ctrlcli.ObjectKeyFromObject(conn), conn)
    if err != nil {
      return nil, err
    }
    envList := new(walruscore.EnvironmentList)
    err = r.client.List(ctx, envList, ctrlcli.MatchingFields{"metadata.name": cb.Namespace})
    if err != nil {
      return nil, err
    }
    if len(envList.Items) != 1 {
      return nil, field.Forbidden(
        field.NewPath("metadata", "namespace"), "connector must be created to a valid environment")
    }

    if envList.Items[0].Spec.Type != conn.Spec.ApplicableEnvironmentType {
      // return error
    }
  }
  
  // Validate duplicated binding.
  {
    cbList := new(walruscore.ConnectorBindingList)
    err = r.client.List(ctx, cbList, ctrlcli.InNamespace(cb.Namespace))
    if err != nil {
      return nil, err
    }

    for i := range cbList.Items {
      if cbList.Items[i].Status.Category == cb.Status.Category &&
        cbList.Items[i].Status.Type == cb.Status.Type {
        return nil, field.Invalid(
          field.NewPath("spec", "connector"), cb.Spec.Connector, "connectors for the same purpose cannot be bound repeatedly"),
        )
      }
    }
  }

  return nil
}
```

pkg/controllers/walruscore/connector_binding.go
```diff
@@ -0,0 +1,112 @@
...
+)
+
+// ConnectorBindingReconciler reconciles a v1.ConnectorBinding object.
+type ConnectorBindingReconciler struct {
+	client ctrlcli.Client
+}
+
+var _ ctrlreconcile.Reconciler = (*ConnectorBindingReconciler)(nil)
+
+func (r *ConnectorBindingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
```
thxCode, line 31: Make reconciling for the same resources.

```
func (r *ConnectorBindingReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
  logger := ctrllog.FromContext(ctx)

  cb := new(walruscore.ConnectorBinding)
  err := r.client.Get(ctx, req.NamespacedName, cb)
  if err != nil {
    // logging
    return ctrl.Result{}, ctrlcli.IgnoreNotFound(err)
  }

  if cb.DeletionTimestamp != nil {
    if systemmeta.Unlock(obj) {
      return ctrl.Result{}, nil
    }

    // Unlabel environment.
    envList := new(walruscore.EnvironmentList)
    err = r.client.List(ctx, envList, ctrlcli.MatchingFields{"metadata.name": cb.Namespace})
    if err != nil {
      return ctrl.Result{}, err
    }
    if len(envList.Items) == 1 {
        // reset environment's label

    }

    // Unlock.
    _, err = kubeclientset.UpdateWithCtrlClient(ctx, r.Client, cb)
    if err != nil {
        logger.Error(err, "unlock connector binding")
        return ctrl.Result{}, ctrlcli.IgnoreNotFound(err)
    }

    return ctrl.Result{}, nil
  }

  // Lock if not.
  if !systemmeta.Lock(obj) {
    obj, err = kubeclientset.UpdateWithCtrlClient(ctx, r.client, obj)
    if err != nil {
        logger.Error(err, "lock connector binding")
        return ctrl.Result{}, err
    }
  }
  
  // Label environment.
  envList := new(walruscore.EnvironmentList)
  err = r.client.List(ctx, envList, ctrlcli.MatchingFields{"metadata.name": cb.Namespace})
  if err != nil {
    return ctrl.Result{}, err
  }
  if len(envList.Items) == 1 {
    // set environment's label

  }

  return ctrl.Result{}, nil
}
```

pkg/controllers/walruscore/connector_binding.go
```diff
@@ -0,0 +1,112 @@
...
+
+	err = r.client.Status().Update(ctx, obj)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{}, nil
+}
+
+func (r *ConnectorBindingReconciler) SetupController(_ context.Context, opts controller.SetupOptions) error {
```
thxCode: Add the following content.

```
	// Configure field indexer.
	fi := opts.Manager.GetFieldIndexer()
	err = fi.IndexField(ctx, &walrus.Environment{}, "metadata.name",
		func(obj ctrlcli.Object) []string {
			if obj == nil {
				return nil
			}
			return []string{obj.GetName()}
		})
	if err != nil {
		return fmt.Errorf("index environment 'metadata.name': %w", err)
	}
```

pkg/webhooks/walruscore/connector.go
```diff
@@ -2,42 +2,125 @@ package walruscore
...
+	r.client = opts.Manager.GetClient()
+
 	return &walruscore.Connector{}, nil
 }
 
 var _ ctrlwebhook.CustomValidator = (*ConnectorWebhook)(nil)
 
 func (r *ConnectorWebhook) ValidateCreate(ctx context.Context, obj runtime.Object) (ctrladmission.Warnings, error) {
-	// TODO: your logic here
+	return r.validateConnector(ctx, obj)
```
thxCode: during creation, we should verify the length of name and connectivity, but for updating, we should verify the connectivity if the connector's spec has changed.

pkg/webhooks/walruscore/connector.go
```diff
@@ -2,42 +2,125 @@ package walruscore
...
+	err := r.client.List(ctx, list, &ctrlcli.ListOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	for _, cb := range list.Items {
+		if cb.Spec.Connector.Name == con.Name && cb.Spec.Connector.Namespace == con.Namespace {
+			return nil, fmt.Errorf("connector is in use")
+		}
+	}
```
thxCode: leverage the mutating webhook of ConnectorBinding, record a label to ConnectorBinding, like walrus.seal.io/connector = "${connector.namespace}/${connector.name}"

then select the related ConnectorBindings with the label selector.
```
cbList := new(walruscore.ConnectorBindingList)
err := r.client.List(ctx, cbList, ctrlcli.MatchingLabelsSelector{"walrus.seal.io/connector": "..."}
if err != nil {
  return nil, err
}
if len(cbList.Items) > 0 {
  return nil, field.Forbidden(,)
}
```

pkg/webhooks/walruscore/connector.go
```diff
@@ -2,42 +2,125 @@ package walruscore
...
 
 	return nil, nil
 }
+
+func (r *ConnectorWebhook) Default(ctx context.Context, obj runtime.Object) error {
+	conn := obj.(*walruscore.Connector)
```
thxCode, line 103: ignore if deleting.

pkg/webhooks/walruscore/connector.go
```diff
@@ -2,42 +2,125 @@ package walruscore
...
+	}
+
+	conn.Spec.SecretName = name
+
+	hideSensitiveData(conn)
+	kubemeta.SanitizeLastAppliedAnnotation(conn)
+
+	_, err := kubeclientset.UpdateWithCtrlClient(ctx, r.client, eSec, kubeclientset.WithCreateIfNotExisted[*core.Secret]())
+
+	return err
```
thxCode: the secret name may exceed the limitation, use stringx.SumByFNV64a() to create a hash suffix.

pkg/webhooks/walruscore/connector.go
```diff
@@ -2,42 +2,125 @@ package walruscore
...
+				Value:   "",
+				Visible: false,
+			}
+		}
+	}
+
+	connector.Spec.ConnectorConfig.Data = configData
+
+	kubemeta.SanitizeLastAppliedAnnotation(connector)
+}
```
thxCode: inline this function into Default as no other callers.

pkg/systemkuberes/environment.go
```diff
@@ -7,18 +7,41 @@ import (
...
+
+	// DefaultConnectorName is the name of the default connector.
+	DefaultConnectorName = "local"
+
+	// LocalEnvironmentKubernetes indicates creating local environment with management kubernetes.
+	localEnvironmentModeKubernetes = "kubernetes"
+	// LocalEnvironmentDocker indicates creating local environment with docker.
+	localEnvironmentModeDocker = "docker"
+	// LocalEnvironmentModeDisabled disables local environment creation.
+	localEnvironmentModeDisabled = "disabled"
```
thxCode: i think using string literals is enough, don't need constants.

thxCode (changes requested):

pkg/webhooks/walruscore/connector_binding.go
thxCode: @shanewxy , after some searching, the status of an object cannot be modified or default during the mutating webhook. so we change this back to reconcile block.

pkg/controllers/walruscore/connector.go
```diff
@@ -3,29 +3,142 @@ package walruscore
...
+					ObjectMeta: meta.ObjectMeta{
+						Namespace: obj.Namespace,
+						Name:      obj.Spec.SecretName,
+					},
+				}
+				err := r.client.Delete(ctx, sec)
+				if err != nil {
+					ctrllog.FromContext(ctx).Error(err, "failed to delete connector", "connector", obj)
+				}
+			}},
```
thxCode: can we use a finalizer to release the related secret?

pkg/controllers/walruscore/connector_binding.go
```diff
@@ -0,0 +1,149 @@
...
+
+			// Unlock.
+			_, err = kubeclientset.UpdateWithCtrlClient(ctx, r.client, cb)
+			if err != nil {
+				logger.Error(err, "failed to unlock ConnectorBinding", "namespace", cb.Namespace, "name", cb.Name)
+				return ctrl.Result{}, ctrlcli.IgnoreNotFound(err)
+			}
+
+			return ctrl.Result{}, nil
+		}
```
thxCode: we need to unlock the ConnectorBinding even if the len(envList.Items) == 0 .

pkg/controllers/walruscore/connector_binding.go
```diff
@@ -0,0 +1,149 @@
...
+		return ctrl.Result{}, err
+	}
+
+	cb.Status.Type = conn.Spec.Type
+	cb.Status.Category = conn.Spec.Category
+
+	err = r.client.Status().Update(ctx, cb)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
```
thxCode:
```
if cb.Status.Type != conn.Spec.Type || cb.Status.Category != conn.Spec.Category {
  // do status update here.
}
```

pkg/controllers/walruscore/connector_binding.go
```diff
@@ -0,0 +1,149 @@
...
+		}
+
+		labels[walruscore.ProviderLabelPrefix+strings.ToLower(cb.Status.Type)] = "true"
+		env.SetLabels(labels)
+
+		err = r.client.Update(ctx, &env)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+	}
```
thxCode: let's do some defending programming here.

```
if len(envList.Items) != 1 {
  // NB: we should never reach here.
  logger.Error(nil, "cannot fetch corresponding environment")
  return ctrl.Result{RequeueAfter: time.Second}, nil
}
```

pkg/controllers/walruscore/connector_binding.go
```diff
@@ -0,0 +1,149 @@
...
+				return nil
+			}
+			return []string{obj.GetName()}
+		})
+	if err != nil {
+		return fmt.Errorf("index environment 'metadata.name': %w", err)
+	}
+
+	return ctrl.NewControllerManagedBy(opts.Manager).
+		For(&walruscore.ConnectorBinding{}).
```
thxCode:
```
For (
  &walruscore.ConnectorBinding{}, 
  ctrlbuilder.WithPredicates(ctrlpredicate.GenerationChangedPredicate{})
  )
```

pkg/extensionapis/walrus/connector_binding.go
```diff
@@ -0,0 +1,133 @@
...
+	return (*walrus.ConnectorBinding)(uo)
+}
+
+func (h *ConnectorBindingHandler) CastObjectListTo(dol *walrus.ConnectorBindingList) *walruscore.ConnectorBindingList {
+	return (*walruscore.ConnectorBindingList)(dol)
+}
+
+func (h *ConnectorBindingHandler) CastObjectListFrom(uol *walruscore.ConnectorBindingList) *walrus.ConnectorBindingList {
+	return (*walrus.ConnectorBindingList)(uol)
+}
```
thxCode: remove these lines

pkg/controllers/walruscore/connector.go
```diff
@@ -3,29 +3,142 @@ package walruscore
...
+		Client:    r.client,
+	})
+	if err != nil {
+		logger.Error(err, "fetch resource handler")
+		return ctrl.Result{}, ctrlcli.IgnoreNotFound(err)
+	}
 
+	err = rh.IsConnected(ctx)
+	if err != nil {
+		return ctrl.Result{}, err
```
gitlawr: No need to reconcile when the check fail

pkg/controllers/walruscore/connector.go
thxCode:
```
For (
  &walruscore.Connector{}, 
  ctrlbuilder.WithPredicates(ctrlpredicate.GenerationChangedPredicate{})
)
```

pkg/controllers/walruscore/connector.go
```diff
@@ -3,29 +3,142 @@ package walruscore
...
+	// Initialize status.
+	{
+		apistatus.ConnectorConditionConnected.Unknown(obj, apistatus.ConnectorConditionReasonConnecting, "Connecting")
+		if apistatus.ConnectorConditionRefresh.IsUnknown(obj) {
+			apistatus.ConnectorConditionRefresh.False(obj, "", "")
+		}
+		err = r.client.Status().Update(ctx, obj)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
```
thxCode, line 68: use predication to prevent a new enqueue event after calling `connectors/status` api.

pkg/controllers/walruscore/connector_binding.go
```diff
@@ -0,0 +1,149 @@
...
+	cb.Status.Category = conn.Spec.Category
+
+	err = r.client.Status().Update(ctx, cb)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
+
+	// Label environment.
+	envList := &walrus.EnvironmentList{}
+	if err := r.client.List(ctx, envList, ctrlcli.MatchingFields{"metadata.name": cb.Namespace}); err != nil {
```
gitlawr, line 110: Why not Get()

pkg/webhooks/walruscore/connector.go
```diff
@@ -2,42 +2,157 @@ package walruscore
...
+	})
+
+	cbList := new(walruscore.ConnectorBindingList)
+	err := r.client.List(ctx, cbList, ctrlcli.MatchingLabelsSelector{Selector: labelSelector})
+	if err != nil {
+		return nil, err
+	}
+
+	if len(cbList.Items) > 0 {
+		return nil, field.Forbidden(field.NewPath("metadata", "name"), "connector is still in use by connector bindings")
```
gitlawr: tell which env is using it

pkg/controllers/walruscore/connector.go
```diff
@@ -3,29 +3,142 @@ package walruscore
...
+	}
```
thxCode: we should give a judgment to evaluate when to initialize the status's conditions, however, we cannot do that in reconciling only. we must leverage admission webhooks.

1. add webhook-gen marker with `subResources=["status"]`
2. validating webhook `ValidateUpdate` to check
  - if the old conditions are completed, like ConnectorConditionConnected == True/False, then allows to change to Unknown. 
  - if the old conditions are processing, like ConnectorConditionConnected == Unknown, then allows to change to True/False.
3. Predicate with `if the generation is changed OR some conditions are Unknown`
  - increasing generation means Spec changing.
  - `ConnectorConditionConnected == Unknown` or the conditions are nil/empty means the User make changing, like we `kubectl edit xxx --subresource=status` to modify the conditions.
4. Skip only if the object is deleted without locking.
  - if the object is deleted and still locked, we should remove the related secret.

Now the reconciling object must be Spec changing or Status initializing.
1. if it's Spec changing, we can do conditions initialization, like give a ConnectorConditionConnected: Unknown to the condition slice, and do `client.Status().Update()`.
2. Keep going with connectivity verification, do `client.Status().Update()` to update the completed status.

pkg/controllers/walruscore/connector_binding.go
thxCode: there is no environment namespace here.

pkg/apistatus/connector_status.go
```diff
@@ -0,0 +1,37 @@
+package apistatus
+
+import v1 "github.com/seal-io/walrus/pkg/apis/walruscore/v1"
+
+const (
+	ConnectorConditionConnected v1.ConditionType = "Connected"
+	ConnectorConditionRefresh   v1.ConditionType = "Refresh"
```
gitlawr: Remove the condition. It can be done by resetting the `Connected` condition.

pkg/apis/walruscore/v1/connector.go
```diff
@@ -14,20 +14,54 @@ type Connector struct {
...
+	// Category is the category of the connector.
+	//
+	// +k8s:validation:enum=["Docker","Kubernetes","Custom","CloudProvider"]
+	Category ConnectorCategory `json:"category"`
+
+	// Type is the type of the connector.
+	Type string `json:"type"`
+
+	// ConnectorConfig is the configuration of the connector.
+	ConnectorConfig ConnectorConfig `json:"connectorConfig"`
```
gitlawr: s/ConnectorConfig/Config/ given that it's in connectorSpec

pkg/controllers/walruscore/connector.go
gitlawr: If the check fail, set the condition to false. 
defer is not neccessary, do something like https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/kube-aggregator/pkg/controllers/status/available_controller.go#L361-L374

shanewxy: updated

gitlawr (approved):
LGTM

thxCode (approved):
LGTM

aiwantaozi (approved):

